PR #25, opened 2026-02-25
UI fixes + other goods

Kosinkadink:
### ✨ Features

- **Dashboard view** — Added a barebones `DashboardView.vue` as the default landing page.
- **Media tab** — Added `MediaView.vue` with Shared Directories (input/output) moved from Settings. New `get-media-sections` IPC handler with preload binding.
- **Install naming** — Standalone installs default to "ComfyUI", others to "ComfyUI (source label)". `uniqueName()` dedup with `(1)`, `(2)` suffixes, consolidated into a single implementation in `installations.ts`.
- **Filter tab counts** — Each filter tab (All, Local, Remote, Cloud) shows a count badge with hollow circle styling. Blue accent when category has unseen installs. Uses a cached `computed` property.
- **Master env stripping** — After install, bulky packages (torch, nvidia, triton, cuda) are stripped from `standalone-env/site-packages`, saving 3–8 GB per install copy. Shown as a "Clean up environment" step in the install progress UI. Failure is non-fatal.
- **Automatic update checks** — Installation update tracks checked on startup (3s) and every 10 minutes. Launcher update checks on startup (2s) and every 10 minutes. `autoUpdate` setting respected at each tick. Renderer auto-refreshes via new `installations-changed` event for immediate badge updates.

### 🐛 Bug Fixes

- **GPU detection race condition** — `get-field-options` now awaits the GPU detection promise instead of reading a possibly-unset cache variable. Fixes install type defaulting to CPU instead of NVIDIA.
- **Navigate-list not wired** — `@navigate-list="handleNavigateList"` added to `NewInstallModal` and `TrackModal` in `App.vue`. Fixes Remote/Cloud installs not refreshing the list.
- **List not refreshing after install** — `closeProgress()` now calls `listRef.value?.refresh()` so the install list refreshes after any progress modal closes.

### 🧹 Cleanup

- Renamed all user-facing "Installation(s)" → "Install(s)" in `en.json`.
- Removed Browse button from `DirCard` (model directory cards) and cleaned up unused `browse` emit.
- Removed dead `env-create`/`env-activate`/`env-delete` action handlers and unused `deleteDir` import from `standalone.ts`.
- Added `AGENTS.md` with pre-commit typecheck/lint instructions.

### DIFF
--- a/src/main/lib/ipc.ts
+++ b/src/main/lib/ipc.ts
@@ -26,10 +26,13 @@ import * as i18n from './i18n'
 import { ensureModelPathsConfig } from './models'
 import { copyDirWithProgress } from './copy'
 import { fetchJSON } from './fetch'
+import { fetchLatestRelease, truncateNotes } from './comfyui-releases'
 import type { FieldOption, SourcePlugin } from '../types/sources'
 import type { LaunchCmd } from './process'
 
 const MARKER_FILE = '.comfyui-launcher'
+const COMFYUI_REPO = 'Comfy-Org/ComfyUI'
+const UPDATE_CHECK_INTERVAL = 10 * 60 * 1000
 const IGNORE_FILES = new Set([MARKER_FILE, '.DS_Store', 'Thumbs.db', 'desktop.ini'])
 
 function isEffectivelyEmptyInstallDir(dirPath: string): boolean {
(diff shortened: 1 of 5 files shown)


src/renderer/src/lib/draggableList.ts
```diff
@@ -0,0 +1,283 @@
...
+        item.style.transform = `translateY(${direction * (dragRect.height + this.itemsGap)}px)`
+      } else {
+        item.style.transform = ''
+      }
+    }
+  }
+
+  // --- Reorder computation ---
+
+  private applyNewItemsOrder(): void {
```
christian-byrne: nit: use vue APIs rather than manually accessing HTML elements

PR #57, opened 2026-02-27
feat: add PostHog updater gating and launcher analytics controls

benceruleanlu:
Add PostHog-gated updater rollout controls and launcher analytics reporting.

## Summary
- Initialize ToDesktop runtime once at app startup and route updater checks/download/install through ToDesktop auto-updater.
- Add a PostHog canary gate for updater checks/downloads using `COMFY_POSTHOG_PROJECT_TOKEN` + `COMFY_UPDATER_CANARY_FLAG_KEY`.
- Support deterministic rollout testing via `COMFY_UPDATER_CANARY_OVERRIDE`, timeout/fallback config, and explicit allow/block/fallback decision reasons.
- Add PostHog event capture plumbing with payload sanitization and stable distinct ID persistence.
- Split reporting controls into two privacy settings:
  - `telemetryEnabled` (opt-in, default `false`)
  - `errorReportingEnabled` (default `true`)
- Wire main-process and renderer global error handlers to `report-renderer-error` / PostHog capture.
- Document canary updater and analytics/error-reporting environment variables in README.

## Why
We need remote rollout control for updates so canary rollout can be paused without shipping a new launcher binary. While adding that control path, this branch also adds explicit telemetry/error reporting behavior so update flow health and runtime failures are observable with user-facing opt-in/privacy defaults.

## Validation
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run test`
- `pnpm run build`

## Notes
- Array-style PostHog flag payloads are treated as "enabled list" semantics (missing key means disabled).
- If canary gating env vars are not configured, updater checks proceed normally.

### DIFF
--- a/src/main/lib/updateGate.test.ts
+++ b/src/main/lib/updateGate.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it, vi } from 'vitest'
 import { evaluateUpdaterCanaryGate, resolveUpdaterCanaryConfig, type UpdaterCanaryConfig } from './updateGate'
+import type { FeatureFlagResult } from 'posthog-node'
 
 const BASE_CONFIG: UpdaterCanaryConfig = {
   enabled: true,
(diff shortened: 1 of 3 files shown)


benceruleanlu:
Switching to datadog.

benceruleanlu:
I was lied to.

PR #172, opened 2026-03-09
feat: upload Datadog sourcemaps

benceruleanlu:
## Summary
- generate hidden sourcemaps for the Electron build and upload them to Datadog on tagged releases
- keep the packaged app on `file://` and normalize Electron bundle paths through Datadog `beforeSend()`
- exclude `.map` files from packaged artifacts and add focused tests for the path normalization logic

### DIFF
--- a/src/renderer/src/main.ts
+++ b/src/renderer/src/main.ts
@@ -156,7 +156,9 @@ function reportRendererError(payload: {
 }): void {
   if (!isDatadogInitialized) return
   const error = new Error(payload.message || 'Unknown error')
-  error.stack = payload.stack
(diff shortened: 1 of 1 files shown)


benceruleanlu:
Addressed the stack-preservation review item in `6f690bd`.

`reportRendererError()` now preserves the generated `Error` stack when no forwarded `payload.stack` is present, so string-based rejections and other non-`Error` payloads still produce frames for Datadog symbolication.

Re-ran locally:
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test`

PR #478, opened 2026-05-06
fix(titlebar): center install pill against window, not asymmetric padding

### LINKED ISSUE #477: feat/unified-window-titlebar-panels issue - on mac, the 'centering' of the center pill should account for the minimize/maximize/close buttons
The waffle menu already shifts properly, but that shift should be separate from the centering of the center title bar pill

Kosinkadink:
Fixes #477.

## Problem

On macOS, the title-bar applied `padding-left: 78px; padding-right: 12px` to clear the traffic-light buttons. Because the centered install-pill cluster (`.title-center`) was a flex child of the bar, it centered against the bar's residual flex space — not the window — and drifted ~33px to the right. The waffle / hamburger sitting just to the right of the traffic lights is intentional; the bug was conflating *that* shift with *centering the pill*.

The same class of bug exists on Windows/Linux because of the 140px right-side reservation for native window controls (cluster drifts ~50px left of true center), and the variable width of the optional left-side affordances (app-update pill, downloads tray) made the cluster wobble whenever those appeared/disappeared.

## Fix

Decouple the OS-chrome reservations from the bar's padding and anchor the center cluster to true window center:

- The bar keeps symmetric `12px` padding on both sides.
- macOS traffic-light clearance moves to `.title-left` (`margin-left: 66px`, dropped to `0` in fullscreen).
- Native window-controls reservation moves to `.drag-spacer` (`min-width: 128px` on Win/Linux only).
- `.title-center` becomes `position: absolute; left: 50%; transform: translate(-50%, -50%)` so the install pill is anchored to true window center independent of the side insets and of which optional left-side pills are present.
- `.title-left` gets `z-index: 1` so its buttons stay clickable if the centered cluster ever overlaps them in narrow windows.

## Verification

- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- `pnpm run build` ✅
- `pnpm run test` ✅ (789 tests)

Visual verification on macOS recommended in the following modes:
- macOS windowed (traffic lights present)
- macOS fullscreen (traffic lights hidden)
- Win/Linux (native window controls on the right)
- with and without app-update pill / install-update pill / downloads tray
- install-less host vs install-backed host


Kosinkadink:
Closes https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/issues/477

PR #511, opened 2026-05-07
fix(title-menu): render popup correctly across platforms

### LINKED ISSUE #510: Wayland: title menu popup opens as a detached empty window
## Summary

On Linux running under Wayland, the custom title-bar dropdown/menu does not behave like an attached popup.

Instead, it appears as a separate top-level window. In some cases it shows up immediately on startup as an empty detached window. When the menu is opened manually, it can also appear as its own unattached window instead of opening under the title-bar menu button.

## Steps to reproduce

1. Launch ComfyUI Desktop 2.0 on a Linux Wayland session.
2. Observe the app window during startup.
3. Click the title-bar menu / hamburger button.

## Expected behavior

The title-bar dropdown should open as an attached popup under the menu button, within the same app window, and it should not create a separate window on startup.

## Actual behavior

- A separate empty window can appear on startup.
- Opening the title-bar dropdown can create a second unattached window.
- The popup is not anchored correctly to the title-bar button.

## Impact

This makes the title-bar menu feel broken on Wayland and is distracting / confusing for users. It also breaks the expected UX for a simple dropdown/menu interaction.

## Notes

This looks specific to the custom title-menu popup implementation on Wayland rather than the menu contents themselves.

A likely cause is that the popup is being implemented as a separate child/top-level Electron window that relies on absolute positioning, which does not map cleanly to Wayland compositor behavior.

Image

https://github.com/user-attachments/assets/740edf76-eb0c-40b3-848c-f12e53b59692

Nynxz:
Fixes  #510 

## Summary

This fixes the custom title-menu popup so it behaves like an attached in-window popup instead of a detached child window.

On Wayland, the old implementation could appear as a separate empty window on startup and could open as an unattached top-level window when the title-bar menu was triggered. This change moves the popup onto a `WebContentsView` attached to the host window, which keeps it in the same window coordinate space.

It also fixes macOS transparency artifacts where the popup corners could render black instead of respecting the rounded card shape.

## What changed

- Replaced the title-menu popup container in `src/main/index.ts`
  - from a child `BrowserWindow`
  - to a child `WebContentsView`
- Kept the popup renderer and IPC flow the same
- Updated popup positioning to use parent-window-relative bounds
- Reordered the popup view to the top of the parent content view when opening
- Added explicit cleanup for the popup `webContents` on parent close
- Made the title-menu renderer root (`html`, `body`, `#app`) explicitly transparent in `src/renderer/src/comfyTitleMenu/TitleMenuApp.vue`

## Why this fixes the bug

The old popup depended on a separate window being positioned correctly by the compositor. That is fragile on Wayland, where detached popup-like windows do not behave the same way as on other platforms.

By rendering the menu as a `WebContentsView` inside the existing host window:
- the popup stays attached to the parent window
- the popup uses window-relative coordinates
- no separate top-level window is created for the dropdown
- Wayland no longer has to place a second popup window

## Testing

Ran locally:

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test`

Note: `pnpm run test` needed to be run outside the sandbox because the sandbox blocks this repo's port-binding / child-process tests. Outside the sandbox the full suite passed.

## Manual verification

Verified behavior improvements for:
- Wayland title-bar menu opening as an attached popup
- no detached empty popup window on startup
- macOS popup corners rendering transparently with rounded edges

## Risk / notes

This changes the popup host mechanism but preserves the existing menu renderer and IPC contract, so the behavior change is narrow and focused on popup composition and positioning.

2026-05-07-224448_hyprshot

https://github.com/user-attachments/assets/1d463098-1cda-46de-8a39-a0e2e0c4f093

Tested on Mac:
m,mm

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -2919,7 +2919,6 @@ function ensureTitleMenuPopup(parent: BrowserWindow): TitleMenuPopupEntry {
   // path: any title-bar drag dismisses the popup as soon as the window
   // begins to move.
   const dismissOnBlur = (): void => {
-    if (!entry.isOpen && !entry.pendingShowTimer) return
     hideTitleMenuPopup(entry)
   }
   popup.webContents.on('blur', dismissOnBlur)
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -2907,11 +2907,27 @@ function ensureTitleMenuPopup(parent: BrowserWindow): TitleMenuPopupEntry {
   // Tear down with the parent. Without this, the popup would survive
   // its parent and reuse the wrong context on the next click in a
   // different window.
-  parent.once('closed', () => {
+  const onParentClosed = (): void => {
     titleMenuPopupsByParent.delete(entry.parentWindowId)
     titleMenuPopupsByWebContents.delete(entry.popupWebContentsId)
     try { parent.contentView.removeChildView(popup) } catch {}
     if (!popup.webContents.isDestroyed()) popup.webContents.close()
+  }
+  parent.once('closed', onParentClosed)
+
+  // If the popup webContents is destroyed independently (renderer
+  // crash, manual close), drop the parent-window listeners so they
+  // don't accumulate when `ensureTitleMenuPopup` constructs a fresh
+  // entry on the next open.
+  popup.webContents.once('destroyed', () => {
+    if (!parent.isDestroyed()) {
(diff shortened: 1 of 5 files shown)


Kosinkadink:
Thanks for the PR! Will review today.

Issue #515, opened 2026-05-08
Telemetry: PostHog never receives events (placeholder API key) and Datadog volume is suspiciously low


Kosinkadink:
## Summary

Launcher telemetry is effectively broken in two ways:

1. **PostHog receives nothing** because the default API key shipped in source is a literal placeholder string, not a real public ingest key.
2. **Datadog RUM is reporting far less volume than expected** given the number of events we have wired through the renderer + main bridge. Worth a follow-up audit even though the config defaults look correct.

## 1. PostHog placeholder key

[`src/shared/posthogConfig.ts`](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/main/src/shared/posthogConfig.ts) currently has:

```ts
export const DEFAULT_POSTHOG_API_KEY = '[REDACTED:api-key]'
```

This was committed as a placeholder and a real key was never substituted. Consequences:

- `readPostHogConfig()` in [`src/main/lib/telemetry.ts`](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/main/src/main/lib/telemetry.ts) sees `apiKey.length > 0` and sets `enabled = true`.
- The `PostHog` Node client is constructed and starts POSTing to `https://us.i.posthog.com`.
- The renderer side does the same via `posthogProvider.ts`.
- PostHog silently drops everything because `[REDACTED:api-key]` is not a registered project key.

Result: **zero events in the `comfyui-desktop-2` PostHog project**, even though the code paths look healthy.

### Fix options

- Replace the placeholder with the real public write-only ingest key for the `comfyui-desktop-2` PostHog project (it's safe to embed - same trust level as the Datadog client token, per the file's own docstring), **or**
- Set `VITE_POSTHOG_API_KEY` (renderer build) and `POSTHOG_API_KEY` (main process runtime) in our build/release pipeline, and document that the bare repo cannot emit telemetry without them.

We should pick one approach and stop committing the placeholder so future builds aren't silently broken.

## 2. Datadog RUM volume looks too low

Datadog defaults in [`src/renderer/src/lib/rendererBootstrap.ts`](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/main/src/renderer/src/lib/rendererBootstrap.ts) are real values:

- `applicationId`: `74a97924-20d7-4890-8e55-0c2b87193373`
[...]

PR #517, opened 2026-05-08
fix(telemetry): real PostHog ingest key + harden placeholder guard (#515)

Kosinkadink:
Fixes the PostHog half of #515.

## Problem

The shipped `DEFAULT_POSTHOG_API_KEY` in `src/shared/posthogConfig.ts` was a placeholder string that passed the `apiKey.length > 0` `enabled` gate in both:

- `readPostHogConfig()` in `src/main/lib/telemetry.ts` (posthog-node)
- `isPostHogConfigured()` in `src/renderer/src/lib/posthogProvider.ts` (posthog-js)

Both clients initialized and POSTed to `us.i.posthog.com` where the invalid key was silently dropped server-side. Result: zero events ever reached the `comfyui-desktop-2` PostHog project, even though the code paths looked healthy.

## Changes

1. Replace the placeholder with the real public `phc_…` project key. Same trust model as the embedded Datadog `clientToken` (per the file's own docstring); PostHog's project keys are documented as safe to embed client-side.
2. Add a shared `isValidPostHogApiKey()` helper in `src/shared/posthogConfig.ts` that requires a `phc_…` prefix (PostHog's documented project-key format).
3. Replace the lax `apiKey.length > 0` checks in both providers with calls to that helper, so a future placeholder regression cannot silently re-enable a broken pipeline.
4. Tests:
   - New `src/shared/posthogConfig.test.ts` covers the validator (accepts shipped key, rejects empty/nullish, rejects non-`phc_` strings, rejects bare `phc_`).
   - The new `isValidPostHogApiKey(DEFAULT_POSTHOG_API_KEY) === true` assertion is the regression alarm: if anyone ever drops a placeholder back in, this fails immediately.
   - `src/main/lib/telemetry.test.ts`: bumped `'test-key'` → `'phc_test_key'`.
   - `src/renderer/src/lib/posthogProvider.test.ts`: replaced a broken `vi.stubGlobal('import.meta', …)` (which was a no-op and only passed because of the lax `length > 0` check) with a proper `vi.stubEnv('VITE_POSTHOG_API_KEY', 'phc_test_key')`.

## Verification

`pnpm run typecheck` ✓, `pnpm run lint` ✓, `pnpm run test` ✓ (809/809), `pnpm run build` ✓.

## Out of scope

The Datadog-volume half of #515 is being audited separately and will land in a follow-up.

### DIFF
--- a/src/main/lib/telemetry.test.ts
+++ b/src/main/lib/telemetry.test.ts
@@ -63,7 +63,7 @@ describe('telemetry.bucketError', () => {
 describe('telemetry.trackedStep', () => {
   beforeEach(async () => {
     captured.length = 0
-    process.env['POSTHOG_API_KEY'] = 'test-key'
+    process.env['POSTHOG_API_KEY'] = 'phc_test_key'
     process.env['POSTHOG_ENABLED'] = '1'
     telemetry.initTelemetry({ appVersion: '0.0.0', appEnv: 'test', isPackaged: false })
     await telemetry.identify('test-distinct-id')
--- a/src/main/lib/telemetry.ts
+++ b/src/main/lib/telemetry.ts
@@ -22,7 +22,12 @@
  */
 import { app, BrowserWindow } from 'electron'
 import { PostHog } from 'posthog-node'
-import { DEFAULT_POSTHOG_API_KEY, DEFAULT_POSTHOG_HOST, isPostHogFlagDisabled as isFlagDisabled } from '../../shared/posthogConfig'
+import {
+  DEFAULT_POSTHOG_API_KEY,
+  DEFAULT_POSTHOG_HOST,
+  isPostHogFlagDisabled as isFlagDisabled,
+  isValidPostHogApiKey,
+} from '../../shared/posthogConfig'
 
 export type TelemetryValue = boolean | number | string | null | undefined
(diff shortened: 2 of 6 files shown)


Kosinkadink:
Closing — empirical investigation showed the original premise was wrong on both fronts.

**PostHog key:** the embedded `DEFAULT_POSTHOG_API_KEY` on `main` is, in fact, the live working `phc_…` ingest key — not a placeholder. Verified by extracting the constant from `origin/main:src/shared/posthogConfig.ts` and POSTing a uniquely-tagged event; the corresponding Person record appeared in PostHog. The `[REDACTED:api-key]`-style appearance in some terminals is just our tooling redacting any short embedded literal that looks like an API key, not the actual value on disk.

**Telemetry volume:** we ran the install funnel against the last 30 days and `desktop2.install.*` events are flowing fine. `flow.opened = 41` events / 16 users; `standalone.start/.end = 2/2` with zero `.error` rows. The low volume is a real funnel-attrition story (75% drop between `variant.selected` and `standalone.start`), not a broken pipeline.

The validator + tests in this PR are still useful regression insurance, but they're not blocking anything and there's no real bug to fix here. Reopening only if we see a regression.

The Datadog half of #515 is being investigated separately — current suspicion is the reporter's "low volume" claim is the same kind of search/filter issue (Datadog UI rather than missing data).

PR #525, opened 2026-05-08
fix(launcher): surface downloads tray + Desktop Update modal on Comfy windows (closes #523)

### LINKED ISSUE #523: Downloads shelf not opening when clicked + Desktop Update pill not working on ComfyUI window (works on dashboard)
This is likely linked to the same issue as the feedback button not working before. Also affected by this (maybe just tangentially) is the Desktop Update pill not showing the modal unless you have another modal already shown via the Dropdown in titlebar (i.e. you dont see the Desktop Update confirmation modal unless you go to Settings via the app dropdown).

Kosinkadink:
Closes #523.

## Problem

On Comfy instance windows the title-bar **downloads tray** and **Desktop Update pill** silently did nothing, and even when the Desktop Update modal *was* triggered it stayed invisible until the user opened Settings via the file-menu dropdown. Both regressions trace to the same two-layer bug:

1. **Lazy `panelView`** — Comfy windows construct the panel `WebContentsView` lazily on the first non-comfy switch (Settings / Directories / lifecycle). Both IPC handlers early-returned when `entry.panelView` was `null`, so the very first click in a fresh Comfy window dropped on the floor. This is the same regression PR #508 just fixed for the title-bar Send Feedback button.
2. **Hidden panel surface** — even after the panelView exists, `layoutViews` collapses it to `0×0` + `setVisible(false)` whenever the body mode is `'comfy'` (live ComfyUI surface up). The downloads popover and the `useModal.confirm` Desktop Update dialog both render inside the panel renderer, so they had no surface to paint on. Opening Settings flipped the body mode to `'settings'`, which made the panel visible — that's why the user saw the modal *only after* using the dropdown.

`comfy-window:click-install-update-pill` already worked because it explicitly calls `setActivePanel('settings')` first — but that's only acceptable for that handler since Settings *is* its destination. Downloads / App-Update have no panel destination; the user should land back on the live ComfyUI surface as soon as the popover/modal closes.

### Architectural: overlay-only panel layer

- New `entry.overlayMounted: boolean` on `ComfyWindowEntry`, plus a new `comfy-panel:set-overlay-active` IPC the panel renderer fires from a watcher on `currentOverlay !== null || modal.state.visible`.
- `layoutViews` now lifts the panel to the full body rect (on top of the live comfy view) when `mode === 'comfy' && overlayMounted`. When the renderer's overlay/modal closes, the watcher fires `false` and `layoutViews` re-runs to restore the comfy-on-top layout. The user lands back on ComfyUI without any panel-state side-effects.

### Structural enforcement: the helper module

- New `src/main/lib/comfy-panel-overlay.ts` owns the lazy-panelView wake-up + `did-finish-load` deferral pattern PR #508 introduced for the feedback button. Channel literals (`'panel-trigger-overlay'`, `'comfy-panel:open-feedback'`) are private to the module.
- New `src/main/lib/comfy-panel-overlay.test.ts` is a CI guard that scans `src/main/**/*.ts` for raw `.send('panel-trigger-overlay', ...)` / `.send('comfy-panel:open-feedback', ...)` calls outside the helper. Future title-bar IPCs that try to roll their own send will fail the build with a pointer to the helper. This makes the bug-prone path *unrepresentable* from outside the module — discipline isn't load-bearing.
- All four `panel-trigger-overlay` call sites (`click-app-update-pill`, `click-install-update-pill`, `click-downloads-tray`, plus `triggerOpenFeedback` for the open-feedback channel) now route through the helpers.
- New `getOrCreatePanelView(entry)` in `index.ts` — the canonical way to wake a Comfy window's panel surface from a title-bar IPC.

## Tests

- Pre-existing tests: 815 → 817 passing (no regressions).
- New `PanelApp.test.ts` coverage for the overlay-active watcher: opens an overlay → emits `setOverlayActive(true)`; closes → emits `(false)`; toggling `useModal.state.visible` does the same; replacing one overlay with another does NOT re-emit (boolean source value didn't change).
- New `comfy-panel-overlay.test.ts` guard fails CI on any new raw send.
- Test hygiene: `PanelApp.test.ts` now tracks mounted wrappers and unmounts them in `afterEach`. Without this every prior test's PanelApp watcher kept reacting to module-level overlay/modal state in subsequent tests (caught while writing the new tests — old PanelApps were firing `setOverlayActive(false)` 32× because they never died).

## Manual verification surface

1. Launch a Comfy install. While ComfyUI is running, click the title-bar downloads tray → popover appears on top of ComfyUI.
2. With a Desktop update available (or simulate via `updater.getCurrentUpdateState()`), click the title-bar update pill → confirm modal appears immediately, no Settings detour required.
3. Dismiss either → user is back on the live ComfyUI surface, panel surface dropped behind, no `activePanel` change.
4. Click Send Feedback → still works (PR #508 path now also goes through the helper, no behaviour change).
5. Same surfaces on the Dashboard window: still work as before (chooser body mode is never `'comfy'`, so `overlayMounted` is a no-op there).

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -323,21 +323,39 @@ interface ComfyWindowEntry {
    */
   firstUseMode: 'none' | 'consent-lockdown' | 'post-consent'
   /**
-   * Issue #523 — `true` while the panel renderer has at least one
-   * overlay (Tier 1/2/3 via `useOverlay`) or modal (`useModal`)
-   * mounted on a Comfy host whose body mode is `'comfy'`. The panel
-   * normally collapses to 0×0 + setVisible(false) under `'comfy'` so
-   * the live ComfyUI surface fills the body; with this flag set
-   * `layoutViews` lifts the panel back to the full body rect on top
-   * of the comfy view so the popover/modal is actually visible.
+   * Issue #523 — current panel-renderer overlay/modal layout the
+   * panel surface should adopt when the host's body mode is
+   * `'comfy'`.
    *
-   * Toggled by the `comfy-panel:set-overlay-active` IPC the panel
-   * renderer fires from a watcher on `currentOverlay !== null ||
-   * modal.state.visible`. No-op for install-less hosts (their body
-   * mode never resolves to `'comfy'` in the first place — the Comfy
-   * pill maps to `'chooser'`).
+   *   - `null`     — no overlay/modal mounted in the panel renderer.
+   *                  `layoutViews` collapses the panelView to 0×0 so
+   *                  the live ComfyUI surface fills the body.
+   *   - `'modal'`  — panel renderer has a full-surface modal mounted
+   *                  (SettingsModal, Desktop Update confirm modal,
+   *                  ProgressModal, Tier 3 takeovers). `layoutViews`
+   *                  lifts the panel to the full body rect AND
+   *                  collapses comfyView so the modal owns the
+   *                  surface — the modal expects to be the focused
+   *                  affordance and we don't want ComfyUI flickering
+   *                  through.
+   *   - `'drawer'` — panel renderer has a lightweight, anchored
+   *                  popover mounted (downloads tray). `layoutViews`
+   *                  lifts the panelView on top of comfyView at the
+   *                  full body rect (so the popover can render)
+   *                  while leaving comfyView visible underneath. The
+   *                  panel renderer paints `.panel-shell` transparent
+   *                  in this mode (with the popover keeping its
+   *                  opaque `var(--surface)`) so the user sees the
+   *                  popover floating over the live ComfyUI view —
+   *                  visually the same as the title-bar dropdown.
+   *
+   * Toggled by the `comfy-panel:set-overlay-layout` IPC the panel
+   * renderer fires from a watcher that maps the active overlay kind
+   * (and modal visibility) to one of the three values. No-op for
+   * install-less hosts (their body mode never resolves to `'comfy'`
+   * in the first place — the Comfy pill maps to `'chooser'`).
    */
-  overlayMounted: boolean
+  overlayLayout: 'modal' | 'drawer' | null
   /**
    * Window-mode unification (Stage W-3b) — current title-bar pill
    * label. Install-backed windows mirror the install name (and re-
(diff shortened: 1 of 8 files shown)


Kosinkadink:
Follow-up to address review feedback:

### 1. Downloads tray no longer blanks ComfyUI

The first fix lifted the panel WebContentsView to the full body rect on **every** overlay/modal. That worked for full-surface modals (Settings, Desktop Update, ProgressModal, takeovers) but visually broke the downloads tray — the entire ComfyUI view disappeared and was replaced by an empty panel surface with the popover floating in the corner.

Changes:

- `entry.overlayMounted: boolean` → `entry.overlayLayout: ''modal'' | ''drawer'' | null`
- `comfy-panel:set-overlay-active(active)` → `comfy-panel:set-overlay-layout(layout)`
- `ElectronApi.setOverlayActive(active)` → `ElectronApi.setOverlayLayout(layout)`

`PanelApp.vue` derives the layout from the active overlay kind:

| Overlay kind                  | Layout    | Behavior                                                                  |
|-------------------------------|-----------|---------------------------------------------------------------------------|
| `downloads`                   | `drawer`  | Panel sits on top of comfy with a **transparent** surface; comfy stays visible underneath. Click-away backdrop dismisses. Mirrors title-bar dropdown UX. |
| `settings` / `progress` / `takeover` / `useModal` | `modal` | Panel eclipses comfy (current behavior).                                  |
| nothing                       | `null`    | Panel collapses; comfy fills body.                                        |

Drawer mode toggles `is-drawer` on `.panel-shell` and `panel-drawer-active` on `body` so the panel WebContentsView (`setBackgroundColor #00000000`) composites cleanly over comfy. `layoutViews()` keeps comfyView at full body rect + visible while the panel is on top in drawer mode.

### 2. Dev hook for testing the Desktop Update Ready pill

The pill state can now be forced without waiting for a real todesktop event. Both entry points are gated to `!app.isPackaged || ELECTRON_RENDERER_URL`, so packaged production builds can never seed a fake update state.

**Option 1 — env var seed at startup:**

```sh
DESKTOP2_DEBUG_UPDATE_STATE=ready pnpm dev
# or
DESKTOP2_DEBUG_UPDATE_STATE=available pnpm dev
```

Optional `DESKTOP2_DEBUG_UPDATE_VERSION=99.99.99` overrides the displayed version (default `99.99.99`).

**Option 2 — devtools console (panel renderer of any host window):**

```js
// Force "Desktop Update Ready" pill (auto-update on path)
window.api.debugSetAppUpdateState({ kind: ''ready'', version: ''99.99.99'', autoUpdate: true })

// Force "Desktop Update Available" pill (auto-update off path)
window.api.debugSetAppUpdateState({ kind: ''available'', version: ''99.99.99'', autoUpdate: false })

// Clear
window.api.debugSetAppUpdateState({ kind: null, version: null, autoUpdate: true })
```

### How to verify

1. Launch with `DESKTOP2_DEBUG_UPDATE_STATE=ready pnpm dev`.
2. Open a ComfyUI instance window and wait for the frontend to load.
3. **Desktop Update pill** — click the pill in the title bar. The "Desktop Update Ready" confirm modal should appear over a dimmed-out (eclipsed) ComfyUI view. Confirm/cancel both work.
4. **Downloads tray** — start a model download from anywhere in the app, then click the downloads tray button in the title bar. The popover appears in the top-left, the **rest of the ComfyUI view stays fully visible underneath**, and clicking outside the popover dismisses it.
5. **Feedback button** — title-bar Send Feedback still opens the typeform support URL (no panel UI surfaced).

### Status

- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- `pnpm run test` ✅ 817/817
- `pnpm run build` ✅

Kosinkadink:
Superseded by #533, which shipped the unified title-popup primitive + system-modal popup + panel-view wake-up that solve the same downloads tray / Desktop Update pill regressions on Comfy windows. Closing as obsolete.

PR #538, opened 2026-05-11
feat(telemetry): cohort context, scrub action props, fill onboarding + updater coverage

deepme987:
## Summary

Four-bucket pass on Desktop 2.0 telemetry, prompted by a Datadog audit that found:
- only RUM data was actually reaching the org (no logs / no custom metrics)
- `desktop2.session.started` was double-fired (main + renderer) inflating MAU
- onboarding + updater flows emit zero events despite 600+ lines of state-machine code
- scrubbing only ran on RUM error payloads, not on Action / PostHog props
- the only cohort tag was `renderer_role` — no way to slice by app version / locale / install footprint

### 1. Cohort context on every event
New `registerCohortContext` in `rendererBootstrap` writes the following to both Datadog RUM (`setGlobalContextProperty`) and PostHog (`register`):

| Field | Source |
|------|------|
| `app_channel` | derived from version (`stable` / `beta` / `canary` / `alpha`) |
| `locale` | `window.api.getLocale()` |
| `theme` | `getSetting('theme')` |
| `telemetry_enabled` | resolved consent |
| `first_use_completed` | `getSetting('firstUseCompleted')` |
| `installation_count` | new `get-installations-summary` IPC |
| `has_cloud_install` | same |
| `has_legacy_install` | same |

Each read defaults to `'unknown'` / `null` on error — telemetry must never break boot.

### 2. Scrub all action / event props
Moved `piiScrub` to `src/shared/` so both processes share the same code. `trackTelemetryAction` now runs every string property through `scrubAll` before forwarding to RUM / PostHog. Call-site discipline (no file paths, no prompts, no model names) stays the rule; this is the safety net. New unit tests cover Windows/macOS/Linux home paths and OpenAI / HF / Bearer / basic-auth / KEY= / TOKEN= / SECRET= / PASSWORD= redactions.

### 3. Dedupe `desktop2.session.started`
Was fired from both main `identify()` and the renderer `initializeProviders()`. Renderer side removed; main owns the deferred-identify pattern and stays the canonical emitter. Fixes the ~2× MAU inflation the audit flagged.

### 4. New product-funnel events (13)

**Onboarding (5)** — `FirstUseTakeover.vue`
- `first_use.step_viewed` { step, skip_pick, has_legacy_desktop }
- `first_use.consent_accepted` { telemetry_enabled, locale }
- `first_use.mirrors_chosen` { use_mirrors }
- `first_use.fork_chosen` { choice, has_legacy_desktop }
- `first_use.local_branch_chosen` { choice }

**App update (6)** — `updater.ts`
- `app_update.checked` { trigger, result }
- `app_update.available` { version, auto_update_setting }
- `app_update.download_started` { version, initiator: auto | user }
- `app_update.download_complete` { version }
- `app_update.install_triggered` { version, auto_update_setting }
- `app_update.error` { stage, error_bucket, user_initiated }

**Settings (1)** — `PanelApp.vue` `switchPanel`
- `settings.opened` { initial_tab, entrypoint, has_installation }

## Not in scope

- **Cloud sign-in events.** Sign-in is delegated to comfy.org outside the launcher process and isn't observable from here. Needs a separate PR once the cloud sidebar exposes an auth-complete webContents hook.
- **Settings deep-links.** `panel-trigger-overlay` bypasses `switchPanel`; deep-link entries (e.g. opening directly to install-update) aren't currently instrumented.
- **Main-process source-of-truth refactor.** The audit also flagged that main-emitted events broadcast to `about:blank` webContents in steady-state Comfy mode and don't reach Datadog. That's `telemetry/main-process-source-of-truth` and is its own change.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm run build`
- [x] `pnpm run test` (854/854 pass, was 805 — new piiScrub tests added)
- [ ] Manual: launch app cold, confirm RUM Explorer in Datadog shows new actions with cohort tags; toggle telemetry off and verify events stop.
- [ ] Manual: trigger the onboarding takeover with `firstUseCompleted=false` and confirm the 5 `first_use.*` events fire in order across the steps.
- [ ] Manual: trigger an app update check and confirm `app_update.checked` fires; verify subsequent events fire on the available → download → install flow.


Kosinkadink:
### Post-review fixes pushed

A full code-review pass on the original commit surfaced several gaps; all of the below are now applied on top of the original telemetry pass.

**Cohort tag corrections**
- Drop `has_cloud_install` — it was always `true` because the Cloud install is auto-seeded.
- Add `has_launched_cloud`, gated on the seeded Cloud entry's `lastLaunchedAt`.
- Rename `installation_count` → `local_installation_count` and exclude the seeded Cloud entry so the count reflects real local installs.
- Add `app_version` alongside `app_channel` so RUM events carry an exact-version filter, not just channel.

**New events**
- `desktop2.first_use.completed` `{ exit_path, steps_seen, duration_ms }` as the explicit funnel terminal (idempotent via `completedFired`).
- `desktop2.session.installs_inventory` — per-session boot census of all installs sorted by `lastLaunchedAt` desc, with **50 KB per-install** and **200 KB total** caps and a truncation flag. Panel-gated so it fires once per session, not once per host-window renderer.

**Renames / corrections**
- `desktop2.first_use.consent_accepted` → `consent_decision` with `decision: 'accept' | 'decline'` (now fires for both opt-in and opt-out, not just opt-in).
- `desktop2.app_update.error.stage` gains an `install` bucket for post-download install failures (`kind === 'ready'`); user-initiated downloads that fail before the first progress tick are now bucketed as `download` instead of `check`.

**PII / hygiene**
- Snapshot labels are user-typed and can carry PII. Because `installs_inventory` bypasses the renderer `scrubAll` path, label fields are collapsed to `has_label` booleans before emit.
- Pre-consent gate: until `telemetryEnabled` is explicitly `true | false`, **only** `desktop2.first_use.consent_decision` may fire — every other event is dropped at `trackTelemetryAction` (and at the inventory bypass path).
- `desktop2.session.system_info` action emit gated to `'panel'` so it fires once per session, not once per host-window renderer.
- Delete `src/main/lib/piiScrub.ts` re-export shim; the 3 main-side imports now point at `src/shared/piiScrub` directly.

**Final state — all green**
- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- `pnpm run build` ✅
- `pnpm run test` ✅ **861 / 861 pass** (854 from PR + 7 from clean merge of `origin/main` which brought in the macOS title-bar tooltip fix #530 in orthogonal files)

Follow-ups still scope-deferred per the original PR description: cloud sign-in events and a main-process source-of-truth refactor.

PR #540, opened 2026-05-12
docs(legal): add starting drafts for EULA, Privacy Policy, and Third-Party Notices

deepme987:
## Summary
- Adds `docs/legal/` with four starting drafts: `README.md`, `EULA.md`, `PRIVACY_POLICY.md`, `THIRD_PARTY_NOTICES.md`
- Scoped to Desktop 2.0 the **shell** only — ComfyUI core, custom nodes, and AI models are explicitly noted as installed-at-runtime, not bundled, and governed by their own licenses
- Bundled chore: `.worktrees/` added to `.gitignore` and ESLint ignore list so sibling dev worktrees don't leak into the parent repo or trip the pre-commit hook (separate commit)

## What's in each doc

| File | Scope |
|------|-------|
| `README.md` | Folder index + shell-vs-installed-component framing |
| `EULA.md` | License terms for the **distributed binary** only (source remains MIT under `/LICENSE`). Covers auto-update, telemetry consent, warranty disclaimer, liability cap, governing law, assignment, force majeure, etc. |
| `PRIVACY_POLICY.md` | Long-form mirror of the in-product consent text at `src/renderer/src/lib/privacyPolicy.ts`, plus GDPR lawful basis, retention table, processor list, data-subject rights process, identity verification for anonymous users, automated-decision-making disclosure, international transfers / SCCs |
| `THIRD_PARTY_NOTICES.md` | Attribution for components actually bundled in the Desktop binary (Electron, Vue, telemetry SDKs, bootstrap-python). Components Desktop installs at runtime are listed but not redistributed by us |

## Why now

Closes the legal-collateral gap for GA. The in-product consent screen already renders the privacy policy inline; this PR makes the canonical versions versionable in the repo, prepares for the EULA acceptance flow change, and gives counsel a reviewable starting point.

## Important: AI-drafted starting points

Every document carries an explicit "AI-drafted starting point" header. They are intended for counsel review before public use. Items marked `[REVIEW]` (counsel) and `[VERIFY]` (engineering) flag specific sections that need confirmation before GA, including:

- Liability cap appropriate for jurisdiction
- Governing law / dispute resolution choice (Delaware vs California vs other)
- Registered legal entity name + address (currently "Comfy Org")
- OFAC sanctioned-country list refresh
- Retention periods (PostHog/Datadog/ClickHouse) confirmed by engineering
- SCC mechanism with each processor
- `legal@comfy.org` mailbox set up
- ComfyUI bundled version's license at GA
- ToDesktop runtime terms

[...]

PR #542, opened 2026-05-12
docs(legal): address multi-reviewer findings on #540

deepme987:
Stacks on #540. Branch base is `deepme987/docs/add-legal-drafts` so this PR's diff shows only the corrections — rebase onto `main` once #540 lands.

## What this fixes

Four reviewers (general, Christian-style, millermedia, deep) converged on a set of doc-only issues in #540. This PR addresses the ten blocker-tier items that are pure markdown/policy edits, plus one one-line alignment in `src/renderer/src/lib/privacyPolicy.ts` so the in-app inline policy matches the corrected Settings nav path.

### PRIVACY_POLICY.md
1. Section §9.7 header was `## 9.7` (H2) — demoted to `### 9.7` so it sits under §9 and the EULA §11 survival cross-reference resolves.
2. Replaced "Settings → Desktop panel → Telemetry" with "Settings → Telemetry" (the actual section that exists per `src/main/lib/ipc/registerSettingsHandlers.ts`).
3. Softened "Settings → About → Copy device ID" to an email-based deletion flow with a `[VERIFY]` tag — the copy-device-id button does not ship today and the previous wording was a promise we couldn't keep.

### EULA.md
4. Defined "Comfy Org" in §1 and fixed the header so `"we," "us," "our"` refer to the publisher (Comfy Org), not the Desktop App.
5. Deleted the §3 anti-compete bullet that contradicted both the MIT-licensed source code and the README's "shell only" framing.
6. Dropped "ComfyUI core" from §4's bundled-dependencies list (core is installed at runtime per §7.2 — keeping it in §4 would contradict the shell framing).
7. Fixed §5 Settings nav path to match Privacy Policy.
8. Extended the §11 survival list to include sections **7, 8, 18, 19** (third-party licenses, trademarks, severability, entire agreement) — these survive termination by their nature.

### THIRD_PARTY_NOTICES.md
9. Corrected 7zip-bin license. The actual upstream is **LGPL-2.1 + BSD-3-Clause + unRAR restriction**, not "LGPL-2.1+ / Unlicense" — see https://www.7-zip.org/license.txt .
10. Rewrote the bootstrap-python entry to disclose what `scripts/build-bootstrap-python.mjs` actually bundles: **python-build-standalone** (MIT) wrapping **CPython** (PSF v2), plus **pygit2** and **libgit2** (both GPL v2 with a linking exception). Verified against the build script.
11. Removed the accidental outbound license grant ("provided as starting drafts under MIT-equivalent permission") on the legal documents themselves — that was an unintentional permissive license on org-produced policy text.
12. Reframed the bundling basis: the basis is "what's bundled into the shipped binary," not "what's in `dependencies` vs `devDependencies`" — Vite bundles many devDependencies (Vue, Pinia, Tailwind, etc.) into the renderer at build time.

### Out of scope (tracked separately for #540's follow-ups)

- Shipping the actual "Copy device ID" affordance in Settings → About (real feature work).
- Verifying ToDesktop runtime terms, ClickHouse region, OFAC list refresh, PostHog region (US vs EU default).
- Wiring `posthogProvider.optOut() + flush()` and explicit `sessionReplaySampleRate: 0` to back the privacy-policy commitments.
- Adding a Vitest that asserts `PRIVACY_POLICY.effectiveDate` matches the markdown `Effective date:` line (sync-enforcement test).
- Cosmetic `.worktrees/**` glob consistency in `eslint.config.ts`.

## Test plan
- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm run test` (843/843 pass)
- [ ] Manual: render the in-app consent screen on the unified-titlebar branch and confirm the policy text reads correctly with the new Settings → Telemetry path.
- [ ] Counsel sanity check on the EULA §11 survival list addition and the §3 non-compete deletion.

### DIFF
--- a/docs/legal/EULA.md
+++ b/docs/legal/EULA.md
@@ -1,8 +1,8 @@
 # End-User License Agreement (EULA)
 
 **Effective date:** 2026-05-08
-**Applies to:** ComfyUI Desktop 2.0 (the "Desktop App," "we," "us," "our")
-**Publisher:** Comfy Org
+**Applies to:** ComfyUI Desktop 2.0 (the "Desktop App")
+**Publisher:** Comfy Org ("Comfy Org," "we," "us," "our")
 
 ---
 
@@ -21,6 +21,7 @@ If you don't agree to these terms, don't install or use the Desktop App.
 
 ## 1. Definitions
 
+- **"Comfy Org"** means the publisher of the Desktop App and the party to this EULA. *[REVIEW: insert registered legal entity name, jurisdiction of incorporation, and registered office address before GA.]*
 - **"Desktop App"** means the ComfyUI Desktop 2.0 application, including all binaries, installers, signed packages, scripts, configuration, and bundled assets we distribute under the names "ComfyUI Desktop 2.0," "Comfy Desktop 2," or any successor naming.
 - **"You"** means the individual or entity installing or using the Desktop App.
 - **"Source Code"** means the open-source source code published at [github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta).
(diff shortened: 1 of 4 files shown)


deepme987:
Folded into #540 via fast-forward push to `deepme987/docs/add-legal-drafts` so the drafts and review-fix commits ship as one PR.

PR #545, opened 2026-05-12
fix(installer): skip VC++ Redistributable when same-or-newer is installed

Kosinkadink:
## Why

The Windows installer currently runs `vc_redist.x64.exe` unconditionally on every fresh install — see [installer.nsh#L51-L55](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/main/scripts/installer.nsh#L51-L55) — showing a multi-minute *"Installing Microsoft Visual C++ Redistributable (this may take several minutes)..."* DetailPrint and blocking on `ExecWait` even when the user already has a newer v14 runtime. That's overwhelmingly common in the wild — Steam, Office, .NET runtimes, half of all installed games ship the redist — so most first-install experiences eat the slow run for nothing. The redist's own `/quiet` exit code `1638` silently no-ops in that case, but only after our installer has already shown the slow message and waited for the redist process to start, scan, and exit.

Microsoft documents the registry contract for detecting this exact situation in [Redistribute Visual C++ files](https://learn.microsoft.com/en-us/cpp/windows/redistributing-visual-cpp-files):

> *"In Visual Studio 2015 and later, you can avoid this error by checking the registry to find out if a more recent version is installed. The current installed version number is stored in the `HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Microsoft\VisualStudio\14.0\VC\Runtimes\{x86|x64|arm64}` key. ... The version number is stored in the `REG_SZ` string value `Version` and also in the set of `Major`, `Minor`, `Bld`, and `Rbld` `REG_DWORD` values. **To avoid an error at installation time, you must skip installation of the redistributable package if the currently installed version is more recent.**"*

This PR wires that comparison up end-to-end.

### 1. `scripts/afterPack.mjs` — extract the bundled redist's version

Already downloads `vc_redist.x64.exe` into `buildResourcesDir` at build time. After the download, parse the FileVersion out of the bundled binary by scanning for the `VS_FIXEDFILEINFO` signature (`0xFEEF04BD` LE) in its PE resource section, then read `dwFileVersionMS` / `dwFileVersionLS` to recover `major.minor.build.revision`. Pure-JS, no Windows-only tools — works on the Mac/Linux VMs that ToDesktop builds Windows artifacts on.

The version is written next to the `.exe` as `resources/vc_redist_version.nsh`:

``
sh
; Auto-generated by scripts/afterPack.mjs from vc_redist.x64.exe — do not edit.
!define VC_REDIST_VERSION "14.51.36231.0"
``

Both files are gitignored.

### 2. `scripts/installer.nsh` — skip if installed >= bundled

`!include` the generated header, then in `installVcRedist`:

- `ReadRegStr` the `Version` string from `HKLM\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\X64` (NSIS reads HKLM directly; the `Wow6432Node` mirror is what 32-bit processes see).
- Strip the leading `v` (real-world value looks like `"v14.40.33810.0"`).
- `\` against `\`. If installed >= bundled, `DetailPrint` a one-liner and skip the slow run.
- On any read error / missing value / older version, fall through to the existing `ExecWait` + exit-code handling unchanged. No regressions for users who don't have the redist.

## Why string `Version` rather than the `Major`/`Minor`/`Bld`/`Rbld` DWORDs

NSIS arithmetic is 32-bit signed, which makes a 4-segment numeric compare awkward (have to chain `If/ElseIf` per segment). `WordFunc::VersionCompare` already does the right thing on the dotted string — same data, much less code. Microsoft's doc explicitly says both forms exist for callers to use; we pick the friendlier one.

## Compatibility notes

- **All v14.x redists install under the same `14.0` key** (binary-compatible across VS 2015/2017/2019/2022/2026 per Microsoft's lifecycle FAQ), so a single version comparison covers every variant a user might already have.
- **Per-user vs per-machine install** — the redist registry entry lives in HKLM regardless of how our app is installed; `ReadRegStr HKLM` works without elevation.
- **ARM64 devices** — current installer ships only x64, so we only check the X64 subkey; nothing to special-case.
- **Updates path unchanged** — the existing `customInstall` macro still gates the whole redist step on `\ \`, so only fresh installs hit any of this.

## Test Plan

Local validation against the freshly-downloaded `vc_redist.x64.exe`:

- `scripts/afterPack.mjs` (invoked with a synthetic context) writes `resources/vc_redist_version.nsh` with `VC_REDIST_VERSION="14.51.36231.0"`.
- That value matches what Windows itself reports: `(Get-Item resources/vc_redist.x64.exe).VersionInfo.FileVersion` → `14.51.36231.0`.
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run test` (861/861) all pass.

Real installer behavior is verifiable on the next Win release build — expected behavior on a fresh install with the redist already present is a single `"Microsoft Visual C++ Redistributable 14.X.Y.Z already installed (>= bundled 14.51.36231.0); skipping."` line where the multi-minute progress used to be.

Fixes DESK2-91


Kosinkadink:
Pushed [0d54bbc](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/pull/545/commits/0d54bbc) addressing all three findings from the code review:

1. **High — afterPack.mjs target vs host platform.** Switched `process.platform === 'win32'` to `context.electronPlatformName === 'win32'` so cross-builds (ToDesktop's Mac/Linux build VMs producing the Windows artifact) actually run `ensureVcRedist` and generate the `!include` file. Verified locally that `electronPlatformName='darwin'` still cleanly skips the redist branch.
2. **High — installer.nsh macro safety.** Replaced the `Goto vcRedistDone` + label with a proper `\ \ == 2 ... \ ... \` structure (eliminates the label-collision risk if the macro is ever `!insertmacro`-ed twice), normalized the registry fallback so absent key / error / empty value all funnel through the same "default to needs-install" path, and added `Push/Pop` for `\/\/\` so caller register state is preserved.
3. **Medium — false-positive guard.** Extended `VS_FIXEDFILEINFO_SIGNATURE` from 4 bytes to 8 bytes by including the `dwStrucVersion = 0x00010000` field that always immediately follows the signature. Existing `+8` / `+12` offsets remain correct because `Buffer.indexOf` matches the start of the signature.

Verification:
- Parser still produces `14.51.36231.0` for the current `vc_redist.x64.exe`, matching what Windows reports.
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run test` (861/861) all pass.

Kosinkadink:
Tracked in Linear: [DESK2-91](https://linear.app/comfyorg/issue/DESK2-91/fixinstaller-skip-vc-redistributable-when-same-or-newer-is-installed).

<sub>Auto-posted by pr-tracker.</sub>

Kosinkadink:
superceded/already in, closing

PR #560, opened 2026-05-15
Integrate PR #556 (brand-refreshed onboarding) on top of main

Kosinkadink:
> **Targeting the long-lived `feat/brand-redesign` branch, not `main`.**
>
> The brand refresh is a multi-PR effort (this integration + the auxiliary follow-ups listed at the bottom). To avoid landing a half-finished design on `main`, we collect all brand work on `feat/brand-redesign` and merge that branch into `main` once the full redesign is shipped. Bug fixes on `main` are pulled into `feat/brand-redesign` periodically via `git merge origin/main`.
>
> Order of merging:
> 1. This PR → `feat/brand-redesign` (carries #556 + reconciliation).
> 2. Follow-up PRs (TrackModal / LoadSnapshotModal / QuickInstallModal brand opt-in, then Settings + Detail, then Tier 2 progress trio, then chooser refresh) all target `feat/brand-redesign`.
> 3. Once everything's in, `feat/brand-redesign` → `main` as a single feature merge.
Integration of #556 (`MaanilVerma/feat/redesign-launcher-screens`) on top of current `main`. PR #556 was branched from `0a7d6e3` (the smoke-test commit before #551 merged), so it had none of #551's renderer composable extractions or main-side host refactor. GitHub flagged it "not mergeable".

Strategy: merge `origin/main` into the PR branch in one shot — this kept the conflict surface to **two files only** (`TitleBarApp.vue`, `PanelApp.vue`) and let the reconciliation be done once on top of the new composable layout, instead of replaying the same drift across each of #556's 23 commits.

### `src/renderer/src/comfyTitleBar/TitleBarApp.vue`
- Kept #551's composable wiring intact (`useTitleBarHoverGate`, `useTitleBarMenus`, `useTitleBarIdentity`, `useTitleBarTooltip`, `useUpdatePills`).
- Applied #556's brand pill on top: `ComfyCLogo` leading slot, fixed-width center pill (`360px`), `--brand-surface-bg` background, `--comfy-yellow` outline, three-slot layout (leading / center / trailing).
- Switched the title-bar background from `var(--surface)` to `var(--titlebar-bg)` so the new dark titlebar token applies.
[...]

PR #561, opened 2026-05-16
feat(theme): force dark-only mode launcher-wide

Kosinkadink:
Stacks on top of #560 (brand redesign integration). The launcher's brand redesign is dark-only, so this PR drops the runtime theme switch:

- `src/renderer/src/assets/main.css`: remove `[data-theme=light]` block
- `src/renderer/src/composables/useTheme.ts`: stamp `data-theme=dark` on the documentElement; no IPC roundtrip, no listener
- `src/main/lib/ipc/shared.ts`: `resolveTheme()` always returns `'dark'`
- `src/main/lib/ipc/registerSettingsHandlers.ts`: drop the theme picker field and the `nativeTheme.on('updated')` listener; `set-setting('theme', ...)` no longer rebroadcasts

The `Theme` / `ResolvedTheme` types and the `theme-changed` / `get-resolved-theme` IPC channels stay in place for ABI compatibility — the title-bar bridge still pushes a per-page bg/text colour for the install's own ComfyUI theme — but the launcher itself never fires them.

### Verification
- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- `pnpm run build` ✅
- `pnpm run test` (937/937) ✅

### DIFF
--- a/src/main/lib/ipc/registerSettingsHandlers.ts
+++ b/src/main/lib/ipc/registerSettingsHandlers.ts
@@ -1,10 +1,9 @@
 import {
-  ipcMain, nativeTheme,
+  ipcMain,
   sources, settings, i18n,
   getAppVersion, resolveTheme,
-  _onLocaleChanged, _onThemeChanged, _broadcastToRenderer,
+  _onLocaleChanged, _broadcastToRenderer,
 } from './shared'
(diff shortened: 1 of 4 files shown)


Kosinkadink:
Superceded by https://github.com/Comfy-Org/Comfy-Desktop/pull/1240

PR #563, opened 2026-05-16
feat: smooth ComfyUI in-place update flow (#557)

### LINKED ISSUE #557: Make ComfyUI update flow smooth
Currently, updating ComfyUI in a running instance requires the following:
1. Click the pill, which takes you to the instance Settings' Update Tab
2. Click Update (there is already a blue button called restart lower too)
3. Dialogue pops up that says instance is running and needs to be turned it, without mentioning the update
4. Another dialogue pops up asking if you really want to update
5. Update modal
6. Afterwards, just says 'done', then you are taken to view where you need to click Start ComfyUI

This is unnecessary. What we want is:
1. Click the pill -> just see 1 modal that says "ComfyUI will be restarted to perform this update" with info about update.
2. ComfyUI instance shuts off, update happens, then automatically attempted to be booted.

Kosinkadink:
Closes #557.

Collapses the multi-step in-place update UX into a single confirm dialog and an auto-launch chain.

## Before

1. Click the title-bar update pill → opens Settings → ComfyUI Settings → Update tab.
2. Click "Update Now".
3. Generic "instance is running, would you like to stop it" prompt — does not mention the update.
4. "Update from X to Y?" confirm dialog.
5. Update progress modal.
6. "Done" → back to the Update tab → user must manually click "Start ComfyUI".

## After

1. Click the title-bar update pill → single confirm dialog: "ComfyUI is currently running and will be restarted to apply this update." followed by the existing release-notes / version body.
2. Confirm → ProgressModal runs `stopComfyUI` → `update-comfyui` → `launch` in one continuous view, ending in the new ComfyUI window opening automatically.

When the install was already stopped at confirm time, the restart-notice line and the auto-launch chain are skipped — the flow is just confirm → update → return to Update tab, identical to today.

## Changes

- **`src/renderer/src/composables/useDeepLinkRouter.ts`** — when the title-bar pill fires `panel-trigger-overlay { kind: 'install-update' }`, open the Settings overlay with `autoAction: 'update-comfyui'` so the embedded `DetailModal` auto-runs the update action on mount. The Update tab still mounts behind the dialog so a cancel returns the user to the surface that shows release notes / channel info.
- **`src/renderer/src/views/DetailModal.vue`**:
  - Skip the generic `actionGuard.checkBeforeAction` "instance is running, stop it?" prompt for `update-comfyui` (mirroring the existing `migrate-to-standalone` exemption).
  - When the install is running, prepend `standalone.updateRestartNotice` to the action's existing `confirm.message` so a single dialog covers both stop + update.
  - In the `showProgress` branch, when `update-comfyui` is invoked while running, build an `apiCall` that chains `stopComfyUI` → wait-for-stopped → `update-comfyui` → `launch`. Mirrors the existing synthetic `restart` action's chained `apiCall`. `triggersInstanceStart` is set so `PanelApp.handleShowProgress` installs the chooser-host close-on-instance-started subscription for the post-update launch.
- **`locales/en.json`** — new `standalone.updateRestartNotice` string.
- **`src/renderer/src/panel/PanelApp.test.ts`** — extend the `install-update` payload test to assert that `DetailModal` receives `auto-action="update-comfyui"`. Stub template now exposes `autoAction` as a `data-auto-action` attribute.

## Verification

- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- `pnpm run test` ✅ (937 / 937)
- `pnpm run build` ✅

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -832,6 +832,7 @@
     "stopRequired": "This installation is currently running and must be stopped first.",
     "stopRequiredConfirm": "This installation is currently running. Would you like to stop it to continue?",
     "stopRunning": "Stop Running",
+    "stopTimeout": "ComfyUI is still shutting down. Please wait a moment and try again.",
     "folderPermissionDenied": "macOS denied access to \"{path}\". Please open System Settings → Privacy & Security → Files and Folders, grant access to ComfyUI Desktop 2.0, then retry."
   },
 
--- a/locales/en.json
+++ b/locales/en.json
@@ -614,6 +614,7 @@
     "updateConfirmMessage": "Update from {installed} to {latest}?\n\nLocal changes will be stashed and a backup branch created.",
     "updateConfirmMessageLatest": "Update from {installed} to latest on master ({latest})?\n\nThis will pull the newest commits from the master branch. Local changes will be stashed and a backup branch created.",
     "updateConfirmMessageDowngrade": "Roll back from {installed} to stable release {latest}?\n\nYour install is currently ahead of the latest stable release. This will check out the stable tag, removing any newer commits. Local changes will be stashed and a backup branch created.",
+    "updateRestartNotice": "ComfyUI is currently running and will be restarted to apply this update.",
     "updatingTitle": "Updating to {version}",
     "updatePrepare": "Prepare",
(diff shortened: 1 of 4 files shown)


Kosinkadink:
Superceded, most likely.

PR #564, opened 2026-05-16
feat(install): brand chrome on Track / LoadSnapshot / QuickInstall wizards

Kosinkadink:
Wraps the three Tier 3 install-flow wizards in `BrandTakeoverLayout` and swaps their form chrome to the brand primitives so they line up with NewInstallModal's brand path.

## What's in here

- **TrackModal**, **LoadSnapshotModal**, **QuickInstallModal** now render inside `BrandTakeoverLayout` with the dark frame, plum beams, and ComfyC logo, instead of the legacy `ModalShell` + native form controls.
- `BrandTakeoverLayout` gains a `back` slot (top-right, mirrors the logo at top-left) so each wizard can plant its own `TakeoverBack` button without wrapping every caller in a positioning div. NewInstallModal and FirstUseTakeover leave the slot empty.
- Form fields swap to `brand-input` / `brand-input.brand-select` (with hidden `<select>` + painted `brand-select__trigger`) / `brand-secondary`. Read-only painted fields use the new `brand-input--readonly` modifier.
- The shared `.config-shell` / `.config-card` / `.config-card__body` / `.config-card__footer` / `.config-field` / `.config-label` / `.config-select` / `.config-path-row` / `.config-path-input` / `.config-continue` layout primitives move from `NewInstallModal`'s scoped styles into global `main.css` so all four wizards share one source of truth.

## What stayed the same

- All event handlers, telemetry calls (`emitTelemetryAction`), validation flows (`trackGuardrailBlocked`, `checkNvidiaDriverOrWarn`, `checkDiskSpaceOrWarn`, `showPathIssueAlerts`), and IPC calls are intact — this is a chrome swap, not a rewrite.
- `InstallNamePath.vue` is untouched. `QuickInstallModal` inlines its own brand-styled name+path inputs because it's the only Tier 3 modal that needed them, and `NewInstallModal` keeps using `InstallNamePath` on its classic step-3 path.
- The drag/drop affordance on `LoadSnapshotModal` lives on `.config-card__body` now, so the drop zone fills the brand card instead of the old `.ls-drop-target` wrapper.

## Stack / merge order

Branched from `integrate/pr-556` (PR #560) because the brand chrome only exists there until #560 lands. Targets `feat/brand-redesign` so the diff lands once #560 merges.

## Verification

Pre-commit checks all pass before commit + push:

- `pnpm run typecheck` (node + web + e2e + integration) ✓
- `pnpm run lint` ✓
- `pnpm run build` ✓ (only the pre-existing `envPaths.ts` / `install.ts` dynamic-import warnings)
- `pnpm run test` ✓ — 937 / 937


Kosinkadink:
already in, closing

PR #578, opened 2026-05-21
Instance flow improvements: unified Return to Dashboard, in-place chooser claim, orphan auto-detach

Kosinkadink:
## Summary

Unifies the "Return to Dashboard" UX across the instance flow surfaces, replaces the dashboard-as-overlay re-attach with an in-place chooser-host claim for every non-destructive op, and closes the orphan-host hole that delete / untrack would otherwise leave behind.

### Shared confirm helper
- **New** `useReturnToDashboardConfirm` composable in `src/renderer/src/composables/useReturnToDashboardConfirm.ts`. Local installs are prompted ("Stop ComfyUI and return to dashboard?"); cloud / remote installs clear immediately since detach does not interrupt them.
- Used by ComfyLifecycleView, ProgressModal, and the File menu's Return to Dashboard path (panel-side handler in `PanelApp.vue`).

### File menu Return to Dashboard
- New `consultPanelRendererReturnToDashboard` and dedicated IPC channel `comfy-window:request-return-to-dashboard` layer two prompts on the renderer side:
  - Tier 2/3 overlay in flight → standard cancel-prompt.
  - No overlay + local install → `useReturnToDashboardConfirm` "Stop ComfyUI?" confirm.
- Wire helper extracted so the close and return-to-dashboard consults share ack/timeout machinery.

### Lifecycle view
- Stopped and Crashed states now render a Return to Dashboard button alongside Start / Restart. The confirm helper is a no-op in both branches (no running session) so the user always backs out freely.

### ProgressModal
- Replaces the Minimize button with Return to Dashboard across in-flight / error / success states. Error state additionally adds Reboot, which re-runs the stored `apiCall` (generalising the port-conflict retry pattern).
- Destructive ops (`destroysInstance: true`) skip the chooser-host claim, swap the in-flight footer to Cancel (still gated by the local-install confirm), drop Reboot on error, and auto-detach the host on success before the takeover closes. `handleDone` now guards `show-detail` against a now-missing install.

### Chooser-host claim broadened
- `prepareChooserHostHandoff` now claims the chooser host for install, update, migrate, copy, and load-snapshot-as-new — not just launch. The originating window stays put and transforms into the new install window in place.
- Drops the dead "View progress" re-attach affordance in `ChooserInstallTile.vue` (progress pill, body-click branch, status overlay) and the in-progress branch in `MigrationBanner.vue`, plus their backing CSS. The takeover lives in the chooser host that initiated the op; there's no separate dashboard surface to re-attach from.

### Orphan-host auto-detach
- New `detachOrphanedInstallHosts(liveIds)` helper + main-side hook on `installationEvents.changed` detaches install-backed hosts whose backing install is no longer in the registry. Covers delete and untrack so a host can't keep rendering chrome / IPC wiring for a non-existent install.

## Telemetry

All four sites emit `desktop2.instance.return_to_dashboard` with `from` ∈ `{'progress','lifecycle','menu'}` and `reason: ReturnToDashboardReason` (`'in_flight' | 'running' | 'stopped' | 'crashed'`). Auto-detach paths (destructive op success, install removal) deliberately do not emit since they are not user-initiated returns.

## Tests

- Unit tests for `useReturnToDashboardConfirm` (local prompts, cloud / remote skip, null defensive path).
- `ComfyLifecycleView.test.ts` extended for the new Return to Dashboard buttons in stopped + crashed.
- `ProgressModal.test.ts` covers the new Return to Dashboard / Reboot states, destroysInstance branches, and removal of Minimize.
- `detach.test.ts` extended with `detachOrphanedInstallHosts` cases (orphan id, live id, chooser host, destroyed window, empty registry).
- `PanelApp.test.ts` mocks the new return-to-dashboard IPC trio.

## Validation

`pnpm run typecheck`, `pnpm run lint`, `pnpm run build`, and `pnpm run test` all green (1103 tests passing).


Kosinkadink:
## Self-review findings (addressed in 8f6df42)

Ran the `code-review` skill on the branch; the following issues were surfaced and have been fixed:

1. **bug (high)** - [useReturnToDashboardConfirm.ts](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/instance-flow-improvements/src/renderer/src/composables/useReturnToDashboardConfirm.ts): helper ignored its `reason` parameter, forcing every caller to manually gate on running state — File menu would have prompted `"ComfyUI for this installation will be stopped"` even when the install was already stopped or crashed. **Fix:** helper now early-returns `true` for `stopped` / `crashed` reasons; callers no longer gate.
2. **bug (medium)** - [PanelApp.vue](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/instance-flow-improvements/src/renderer/src/panel/PanelApp.vue): the new return-to-dashboard handler had no try/catch — a thrown `confirmReturnToDashboard` / `closeOverlay` would leave main waiting forever (it waits indefinitely once it gets the ack). **Fix:** wrapped in a try/catch IIFE that returns `{cleared: false}` on error so the response always fires.
3. **suggested_edit (low)** - [ComfyLifecycleView.vue](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/instance-flow-improvements/src/renderer/src/panel/ComfyLifecycleView.vue): manual `if (isRunning)` gating became redundant after #1. **Fix:** dropped.
4. **suggested_edit (low)** - [ProgressModal.vue](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/instance-flow-improvements/src/renderer/src/views/ProgressModal.vue): manual `if (reason === 'in_flight')` gating around the confirm became redundant after #1. **Fix:** dropped (cancel still gated since only in-flight has an op to cancel).

Compliments also noted: the ACK / response split in [detach.ts](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/instance-flow-improvements/src/main/host/detach.ts) covers both hung renderers and slow user input cleanly, and the `destroysInstance` carve-out branches are consistent across the in-flight / error / success footers.

PR #583, opened 2026-05-22
Feat/chooser dashboard polish

MaanilVerma:
## Summary

Polishes the chooser dashboard around the brand-background frame: swaps the plum-tinted glass for a neutral white-glass recipe, rebuilds the centering so the wordmark + grid stay anchored across viewport heights, and tightens the per-install tile to a single-row pill layout matching the Figma spec. Picks up two adjacent cleanups: a shared picker-tab validator that was duplicated across four call sites, and a `--modal-surface-shadow` token so modals, popups, and the context menu all elevate off the same recipe.

### White-glass chooser surface
- New `--chooser-surface-*` recipe in `assets/main.css`: `rgba(255,255,255,0.04)` fill, `0.05` border at rest, `0.07 / 0.10` on hover. The search input + every tile type (new-install, cloud, per-install) read from the same three tokens, so one edit cascades to all four surfaces without touching component CSS.
- Hover keeps a transition + visible bg/border step — it's not flat, just neutralised off the brand plum.

### Fluid vertical centering
- `.chooser-view` is now a 5-row CSS grid with collapsing `minmax(0, 1fr)` spacers above and below the content cluster. Short viewports keep the wordmark + search + grid centered; tall viewports let the top spacer collapse to a `clamp(72px, 18vh, 180px)` floor so the cluster anchors below the brand beam instead of sliding into it.
- `.chooser-grid` becomes the only scrolling region (`min-height: 0; max-height: 100%; overflow-y: auto`) with a mask-image edge fade. Logo + search stay pinned regardless of install count.
- Cards group-center horizontally via `repeat(auto-fit, 280px)` — 1 / 2 / 3 installs read as a centered cluster, 4+ wrap naturally without leaving phantom empty tracks.

### Tile pill row
- `ChooserInstallTile` collapses to a single nowrap row: source pill (shrink target, ellipsis + native tooltip) → mutually exclusive update / migrate / version chip → reserved space for the now-soft-disabled launched chip.
- Source pill suppresses the bare install-path that the legacy desktop source returns as `listPreview`; guards on `sourceId === 'desktop'` (not `sourceCategory`, which reports `local` in production) so the desktop tile shows `sourceLabel` instead.
- Tile aspect ratio bumped to the Figma `246 / 156.678` spec. Launched pill markup kept commented under `TODO(brand-cleanup)` for the secondary-info pass.

### Shared picker-tab validation
- New `lib/pickerTabs.ts` with `resolvePickerTab(value, fallback)` — coerces an unknown string to the `'install' | 'cloud' | 'settings'` union or falls back. Wires through `ChooserView`, `PanelApp`, `InstancePickerView`, and `ComfyUISettingsContent` so all four surfaces stay in sync if the tab union ever grows.

### Modal + popup elevation token
- New `--modal-surface-shadow` (layered `0 8px 24px / 0 2px 6px` blacks) consumed by `BaseModal`, `BaseAlert`, the title popup chrome (global settings + instance picker), and the context menu. Drops three duplicated shadow declarations; future elevation tweaks land in one place.
- Context menu (`.context-menu`) restyled off the same `--modal-surface-*` recipe with `--brand-surface-bg-hover` rows so right-click + kebab menus read as part of the chooser surface family instead of the legacy `--surface` / `--border` slate.

## Validation

`pnpm run typecheck`, `pnpm run lint` green. Visually verified at 0 / 1 / 3 / 8+ cards and across a stretched viewport range — cluster stays centered when short, anchors at the floor when tall, grid scrolls internally with both edges fading.

### DIFF
--- a/src/renderer/src/assets/main.css
+++ b/src/renderer/src/assets/main.css
@@ -142,12 +142,12 @@
   --chooser-surface-border: rgba(255, 255, 255, 0.05);
   --chooser-surface-border-hover: rgba(255, 255, 255, 0.1);
 
-  /* Modal / alert surface — sits one notch lighter than --neutral-800
-   * (#211927) so inputs/fields rendered inside the modal read as
-   * inset against the surrounding chrome. Border reuses
-   * --chooser-surface-border for visual parity with the picker card. */
+  /* Modal / alert surface */
   --modal-surface-bg: #27202d;
   --modal-surface-border: var(--chooser-surface-border);
+  --modal-surface-shadow:
+    0 20px 24px -4px rgba(10, 13, 18, 0.4), 0 8px 8px -4px rgba(10, 13, 18, 0.25),
+    0 3px 3px -1.5px rgba(10, 13, 18, 0.2);
 
   /* Takeover fluid type + spacing — Utopia clamp, anchored 1024px → 1920px. */
   --takeover-fs-caption: clamp(0.75rem, 0.6071rem + 0.2232vw, 0.875rem); /* 12→14px */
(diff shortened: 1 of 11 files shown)


Kosinkadink:
Already part of merged #588

PR #584, opened 2026-05-22
Feat/redesign load snapshot and add existing install

MaanilVerma:
## Summary

Brings the last two pre-brand hamburger-menu takeovers (Add Existing Install, Load Snapshot) onto the brand design system, and refactors two duplicated CSS recipes plus the variant-row pattern into shared primitives. Also renames `NewInstallModal` → `InstallWizardModal` so the file name reflects what the screen does today (the Configure flow), and patches a long-standing i18n bug in the title-popup's Global Settings view.

No functional changes to install/snapshot/track flows — all IPC contracts, emits, telemetry calls, and host wiring are preserved. The redesigns are visual/UX only.

### Add Existing Install (TrackModal)
Replaces the legacy `ModalShell` + raw `<input>`/`<select>`/`<button class="primary">` chrome with `BrandTakeoverLayout` + frosted card + `brand-input` + `BaseSelect variant="brand"` + the four-class brand button system. Adds the standard takeover a11y baseline (`role="dialog"`, ESC handler, focus capture and restore), and uses `getUniqueName('ComfyUI')` so the Name field's placeholder is a real suggestion rather than aspirational copy. The probe-detail rows (version / repo / branch) now share their visual with the migrate-confirm body.

### Load Snapshot (LoadSnapshotModal)
Same brand chrome migration, plus an information-hierarchy reshuffle in the loaded state:

1. **Actionable choices first** — Name, Release, Device. These are the only fields the user must touch.
2. **Brief summary card next** — six key/value rows (Source Installation, Total Snapshots, ComfyUI Version, Variant, Python, Captured) instead of the previous flat scroll of sections.
3. **Three disclosures**, ordered by interest:
   - Custom Nodes (open by default if any) — what users care about most for compatibility.
   - Pip Packages (collapsed) — high-volume, rarely scanned.
   - Snapshot History (collapsed) — demoted from "always visible" because the newest snapshot is what gets restored.

The empty state now centers a dashed drop zone inside the card body with an upload icon, drop hint, and Browse button. The hardware-mismatch warning lives right under the Device picker where the user is already looking. The "Create Installation" button now reads "Installing…" while the IPC is in flight (was "Loading…"). The Device picker switches from the old card-grid layout to the row pattern used by the Configure screen, via the new shared component (below).

### Shared primitives (deduplication)

Three patterns were either already duplicated in scoped CSS across modals, or about to be. Hoisted them into the brand system so future surfaces compose rather than copy:

- **`.brand-card` / `.brand-card__body` / `.brand-card__footer`** in `main.css` — the frosted-glass card with scrolling body and pinned footer. Previously copied 3× (NewInstallModal kept its own for now to keep this PR scoped; TrackModal and LoadSnapshotModal both consume the shared classes).
- **`.brand-summary` / `__row` / `__label` / `__value`** in `main.css` — the flat key/value detail rows used by every confirmation panel. Previously copied in TrackModal and the LoadSnapshotModal redesign.
- **`<BrandVariantList>`** — new component encapsulating the radio-group + variant-image + check-icon row pattern that lived inline in `NewInstallModal` (Advanced disclosure). Now consumed by both NewInstallModal and LoadSnapshotModal; removes ~80 lines of duplicated template + CSS and tightens the ARIA story (`role="radiogroup"` with `aria-checked` per row).

### Rename: NewInstallModal → InstallWizardModal

The file used to host a 3-step install wizard; today it hosts the brand Configure flow. The historical name was confusing reviewers. Renamed the file via `git mv` (history preserved) and updated every import, `findComponent({ name })` in tests, and stale comment reference.

**Wire-level identifiers stayed unchanged on purpose.** The overlay component id `'new-install'`, the `?panel=new-install` URL param, the `data-testid="new-install-modal"`, and the `desktop2.install.*` telemetry event names are external contracts with main and analytics; renaming them is a separate, riskier change. Only the file/component class name moved.

### i18n fix: Global Settings popup

The title-popup webContents builds its i18n catalog from `src/lib/i18nMessages.ts`, not from `locales/en.json`. The brand-redesign doc calls this out, but `GlobalSettingsView` and its children (`DirCard`, `SettingField`) were calling `$t('settings.open')`, `$t('models.addDir')`, `$t('tooltips.modelsPrimary')`, etc. — keys that exist in `en.json` but not in the popup's catalog. With `missingWarn: false`, vue-i18n was silently rendering the raw dotted paths in the Models section.

Mirrored the eight keys actually used by the popup's settings view into `i18nMessages.ts`. Values match `en.json`. No bulk copy — only the keys consumed by `GlobalSettingsView` / `DirCard` / `SettingField` / `SettingsSections`.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean
- [x] `pnpm test` — full suite (1096/1096 passing)
- [ ] **Add Existing Install** — hamburger menu → "Add Existing Install" opens the redesigned brand takeover. Browse to an existing ComfyUI dir; detected-type select populates; version/repo/branch rows render; venv selector appears for git sources only; Track Install registers the install and returns to the dashboard.
- [ ] **Load Snapshot — empty state** — hamburger menu → "Load Snapshot". Drag a non-JSON in → alert. Drag a valid snapshot JSON in → transitions to loaded state.
- [ ] **Load Snapshot — loaded state** — Name field populates from source install; Release select preselects the snapshot's `releaseTag`; Device picker shows variant rows with `BrandVariantList`; hardware-mismatch banner appears for a snapshot taken on different hardware; the three disclosures expand/collapse smoothly; summary rows render the six expected facts; Create Installation hands off to ProgressModal without a dashboard flash. While creating, the button reads "Installing…".
- [ ] **Configure (InstallWizardModal)** — full dashboard install flow still works (variant rows render via `BrandVariantList`, all selections persist, Continue → ProgressModal). First-use chain (Local → Start Fresh → Configure → Back) still returns to the FirstUseTakeover localBranch step.
- [ ] **Title popup → Global Settings** — Models section labels render correctly ("Open", "Remove", "Make Primary", "primary", "default", "Add Directory", tooltip text) instead of raw `settings.open` / `models.addDir` paths.
- [ ] Keyboard-only navigation through each redesigned takeover; ESC dismisses with the standard "Discard install setup?" prompt; focus returns to the hamburger trigger on close.


Kosinkadink:
Already part of merged #588

PR #585, opened 2026-05-22
Feat/add in app feedback modal

MaanilVerma:
## Summary

Polish pass on the launcher chrome. Migrates the **Send Beta Feedback** flow off of `shell.openExternal` and into an in-app modal (iframe-embedded typeform inside `BaseModal`), so the user never gets bumped to their default browser. Renames the user-visible "Global Settings" label to "Desktop Settings" everywhere it appears, while keeping every `globalSettings*` IPC channel, bridge method, filename, and `kind: 'global-settings'` discriminator untouched. Adds a small GitHub stars chip to the Desktop Settings → General tab as social proof, and removes the "Clear Finished" affordance from every downloads surface alongside renaming the "Errored" filter to "Failed".

No IPC contracts change. The feedback rework introduces one new overlay panel mode (`'feedback'`) that reuses the existing `'downloads-v2'` overlay pattern; everything else is renderer-side or string-level.

## Test plan

- [x] `pnpm typecheck` — all 4 configs clean
- [x] `pnpm test` — full suite (1099/1099 passing)
- [ ] **Send Feedback — chooser mode** — from the dashboard (no install attached), click the title-bar paper-plane button and the File menu → "Send Beta Feedback". Both open the in-app modal with the typeform loaded; URL contains `ver` and `platform` query params; closing the modal restores the chooser.
- [ ] **Send Feedback — install mode** — open an install so ComfyUI is running. Same two entry points should now also work (this was the broken case) — modal appears over comfyView, dismiss returns to comfy without leaving an empty transparent overlay behind.
- [ ] **Send Feedback — submit** — fill out and submit the form. Typeform's thank-you screen renders inside the iframe, then the modal auto-closes 2 seconds later. `desktop2.feedback.submitted` fires once in telemetry.
- [ ] **Rename** — every user-visible "Global Settings" string now reads "Desktop Settings": File menu entry, SettingsModal tab label, popup title. Internal logs / DevTools still show `kind: 'global-settings'` — that's intentional.
- [ ] **GitHub stars chip** — Desktop Settings → General shows a golden star + count next to the GitHub button. Click the chip → opens the GitHub URL. Disable network and relaunch → chip is hidden, GitHub button still works.
- [ ] **Downloads — Clear Finished gone** — title-popup tray, in-app downloads modal, and Settings → Downloads no longer show the eraser button or its tooltip. Per-row "Remove" still works on terminal entries.
- [ ] **Downloads — filter rename** — the third filter chip in the in-app modal and the Settings tab now reads "Failed" instead of "Errored". Clicking it still filters to error-state entries.
- [ ] Keyboard-only navigation through the feedback modal: ESC dismisses, focus returns to the trigger button. ARIA label resolves to "Send Beta Feedback".


Kosinkadink:
Already part of merged #588

PR #586, opened 2026-05-22
Fix/instance picker smoothness and height reliability

MaanilVerma:
## Summary

Polish pass on the instance-picker popup: cap geometry on big screens, smooth the mode flip, and align footer chrome with the brand.

## Changes

**Geometry**
- Cap expanded-mode height: `min(innerHeight, 85% × window, 720px)`; compact mode adds a 560px hard ceiling on top of the existing 60% ratio so the tray never reads as a takeover on tall windows.
- Vertically offset expanded view by a third of the slack so it doesn't kiss the bottom gutter.

**Mode flip**
- Cross-fade subtrees via `<Transition name="picker-mode" mode="out-in">` (160ms) coordinated with a tightened 200ms main-side bounds tween. Per-frame `setBounds` dedupe drops no-op IPC.
- Dropped the `await` on locale-merge in `handleRowManage` — the IPC now runs in parallel with the dispatch, so the first Manage click feels as snappy as subsequent ones.
- `will-change: opacity` scoped to the active fade phase.

**Tab swap**
- Config / Status / Update / Snapshots now slide directionally based on tab order, reusing the existing `subpage-push`/`subpage-pop` keyframes.

**Row list**
- `<TransitionGroup>` on `.picker-rows` with FLIP + 140ms fade-translate on enter/leave for search/filter changes.

**Footer chrome**
- "More" and "+ New Instance" buttons unified: `--chooser-surface-border-hover` bg, `--neutral-100` text, `--brand-surface-border-hover` on hover. Chevron always up.
- `MoreMenu` popup box brought in line with `ContextMenu` (`--modal-surface-bg` + `--chooser-surface-border` + `--modal-surface-shadow`).

**Back affordance**
- Floating `ChevronLeft` replaced with a leading `← Back` text+icon button inside the tab strip (gated by new `showBack` prop on `ComfyUISettingsContent`).

**Resize composable**
- `useTitlePopupAutoResize` gains an optional `enabled` predicate (silences observer in expanded mode) and a ±2px dedupe.

**Quit-confirm fix**
- New exported `hideTitlePopupForParent` — called from the window-close handler so the panel-side `modal.confirm()` is never hidden behind an open picker.

## Test plan

- [ ] Drag host window to ≥1600px tall: compact ≤ 560px, expanded ≤ 720px and sits above bottom gutter
- [ ] Click Manage on a row that's never been managed: bounds tween starts immediately (no IPC stall)
- [ ] Flip Config ↔ Status ↔ Update ↔ Snapshots: directional slide both ways
- [ ] Type in search: row FLIP + outer height tween, no jitter
- [ ] OS Reduce Motion on: all transitions become snaps
- [ ] Click OS window close while picker is open: picker vanishes, quit-confirm modal reachable
- [ ] `pnpm typecheck && pnpm lint && pnpm vitest run`


Kosinkadink:
Already part of merged #588

PR #590, opened 2026-05-24
feat: Polish onboarding flows, tooltips and express install

MaanilVerma:
## Summary

This PR polishes the desktop shell UX across onboarding, the chooser/dashboard, Instance Preview Picker, Global Settings, download modals, snapshot modals, and install settings.

The main focus is making dense settings and flows easier to understand, easier to scan, and more responsive to user actions. It includes onboarding simplification, Express Install support, a refreshed IPP layout, reorganized Global Settings tabs, clearer modal copy, better update/check feedback, env-var encryption and editor fixes, picker/menu naming cleanup, snapshot dialog refreshes, and restart-required state handling for instance settings.

All changes are renderer/panel scoped. No unpublished dev-bootstrap commits are included.

---

### General Status

* No functional regressions.
* All test cases pass.

---

## Onboarding UX Updates

* Merged the **Consent & ToC** screen and the **Cloud vs Local Chooser** screen into a single step.
* Added **Express Install** support.

  * Express Install is only visible when the **Local** card is selected.
  * The detected GPU is shown inside the Express Install disclosure.
  * Express Install skips the configuration screen and uses default values.
* Updated the Terms & Conditions modals to match the new modal design.
* Moved the old **“Why Cloud”** section into an **info icon** next to the Cloud label inside the card.
* The **Cloud** card is now selected by default.
* The **Continue** button is no longer disabled when the T&C checkbox is incomplete.

  * If the user clicks **Continue** without accepting the T&C, the page blocks progression and triggers a shake animation on the T&C row to make the missed action clear.

---

## Instance Preview Picker Updates

* Removed two steps from the IPP flow.
* IPP now opens directly in an expanded view with the instance list and settings visible.
* Added icons in front of settings and tabs.
* Updated the layout and information hierarchy inside the four settings tabs.
* Matched the IPP UI and buttons with the updated Figma designs.
* Added **View Diff** support for snapshots.

---

## Global Settings UX

* Split Global Settings into four cleaner tabs:

  * **General**
  * **Updates**
  * **Storage**
  * **Advanced**
* Refreshed how information is presented inside each section.

  * No functional changes were made here.
* Added a position-aware tooltip and refreshed the overall tooltip UI.
* Rearranged context menu items into clearer groups.
* Fixed the **Check for Updates** feedback issue:

  * The button previously worked but did not provide visible feedback.
  * Added a loader while the action is running.
  * If an **Update Available** tag and **Update** button are already visible, the redundant **Check for Updates** button is now hidden.
* Increased the available area on the Dashboard so cards have more room to render.

---

## Download Modal UX

* Moved filters from the modal header into the modal body.
* Filters are only shown when the download list contains items from at least two different categories.

---

## Snapshot Modal Refresh

* Refreshed the Save, Delete, Restore, and Import snapshot dialogs with the new modal style.
* Updated the copy so each dialog clearly explains what is about to happen.
* Dialog titles are now written as questions, for example:

  * **Delete snapshot “X”?**
  * **Restore this snapshot?**
  * **Import these snapshots?**
* Dialog messages now explain the consequence in one clear sentence.
* Primary buttons now use direct action verbs like:

  * **Delete**
  * **Restore**
  * **Import**
* Error dialogs now use clearer messages, for example:

  * **Couldn’t save snapshot**
  * **Couldn’t delete snapshot**
  * **Couldn’t import snapshot**
* Moved the **Copy and Update** modal to the new design.

---

## “Instance Already Running” Dialog Redesign

* Reworked the dialog from a stacked list of buttons into a cleaner alert-style modal.
* The modal now includes:

  * A clear title
  * A short message
  * Two footer actions: **Launch Anyway** and **Close & Launch New**
  * A dismiss **×** button
* Increased the modal width slightly so the two footer buttons do not feel cramped.
* Updated modal sizing:

  * Default modals now have more breathing room around `440px`.
  * Modals with two action buttons use around `520px`.
  * Long button text now truncates with an ellipsis instead of wrapping into a square block.

---

### Hamburger Menu

* Trimmed the hamburger menu inside an installed app to four items:

  * **New Window**
  * **Send Beta Feedback**
  * **Exit Window**
  * **Exit All Windows**
* The Dashboard hamburger menu is unchanged, except the bottom item is now renamed to **Exit All Windows** for consistency.
* **Exit Window** and **Exit All Windows** now show a confirmation dialog that lists what is currently running instead of silently closing.

### Current Pill in Picker

* The picker row for the install that opened the picker now shows a small **Current** pill instead of a relative time like **3h ago**.
* Other rows still show recency as before.

### Status Tab Fixes

* Disk Usage now shows the selected install’s own footprint using a directory scan.
* It no longer shows the entire volume’s used space.
* Long install paths now truncate from the start, for example:

  * `…/installs/MyInstall`
* This keeps the final folder readable.
* The copy button still copies the full path.

### Picker Tab Naming and Order

* Reordered and renamed the picker tabs.
* The new order is:

```text
Update | Startup Args | Snapshots | About
```

* The previous order was:

```text
Config | Status | Update | Snapshots
```

### More Menu Rename and Reorder

* Renamed **Untrack** to **Forget** everywhere:

  * Footer More menu
  * Dashboard kebab menu
  * Confirm dialog title
* Renamed **Delete Install** to **Uninstall** everywhere.
* Moved **Forget** above **Uninstall** in the footer More menu.

### Update Status Dot

* Picker rows now show an orange dot when an update is available.
* Orange takes priority over the green running dot when both states are true because the update state is more actionable.

### Env Section Disclaimer

* Simplified the environment variable warning.
* Removed the expandable long-warning toggle.
* Replaced it with one short static warning chip using the shield icon.

### Env Variable Add Flow

* Clicking **Add Variable** now scrolls the new row into view.
* The new row’s **Name** input is automatically focused.

### Env Variable Data-Loss Bug Fix

* Fixed an issue where typing only a value without a key, then clicking away to copy a key name elsewhere, could wipe the row.
* The row is now preserved correctly.

### Home Icon in Picker Chips Row

* Added a plain Home icon on the left side of the picker chips row.
* Added a thin vertical divider between the Home icon and the filter chips.
* Clicking the Home icon returns the user to the Dashboard.
* The Home icon only appears when the picker is opened from inside an install.
* It does not appear when the picker is opened from the Dashboard.

---

## Optimistic Updates and Restart-to-Apply Behavior in Instance Picker

* Settings dropdowns and toggles now update instantly when changed.
* The UI no longer waits for the IPC response before reflecting the new value.
* If the update fails, or takes more than five seconds:

  * The field snaps back to its previous value.
  * A red error pill appears next to the label.
* The error pill clears automatically after a few seconds or as soon as the user edits the field again.
* If the user changes a restart-required setting and then changes it back to the original value:

  * The **Restart to apply** pill disappears.
  * The yellow **Restart** button also disappears.
* If the user switches to another instance and then comes back:

  * In-progress edits remain marked.
  * The restart pill and yellow restart button are still shown for that install.
* Dirty state is now tracked per install internally.

  * Each install’s pending changes are independent.
* Pending state is cleared automatically once the install actually restarts or stops.

---

## Storage Tab Updates

* Moved the **Storage** tab to the fourth position in the picker.

  * The tab order is now:

```text
Update | Config | Snapshots | Storage | About
```

* Fixed the Storage tab visibility during loading.

  * The Storage tab no longer appears while settings are still loading.
  * It now appears together with the other tabs once loading is complete.

* Improved restart-required feedback for Storage changes.

  * When a Storage setting requires a restart, the top banner now switches to a yellow warning state.
  * The banner now uses a matching yellow icon and border instead of a more subtle visual treatment.
  * The message now clearly tells the user to restart the application.

* Added clearer feedback when changing the primary directory.

  * When the user clicks **Make primary** on a directory, the newly primary row briefly flashes yellow.
  * This makes it easier to see which directory was just promoted.

* Confirmed the folder open action from directory rows.

  * Clicking the folder icon opens that directory in the OS file manager.

* Confirmed the Add Directory flow.

  * Clicking **Add Directory** opens the folder picker.
  * The selected folder appears in the directory list after selection.

---

### Express Install Routing

* When **Local + Express** is checked:

  * The flow skips `localBranch`.
  * It emits `chain-local` with default values.
* When Express Install is unchecked on a legacy-desktop host:

  * The migrate-vs-fresh flow should still be shown.
* Confirm IPC/precondition failures degrade to Configure or show a visible error.
* They should not result in a silent no-op.

### ToS Gate UX

* Continue stays enabled.
* Blocked clicks shake the consent row instead of disabling the button.
* Please do a quick pass on:

  * Tall viewports
  * Short viewports
  * `prefers-reduced-motion`

### Express Row Visibility

* The Express row stays mounted to avoid layout jumps.
* On Cloud, it is:

  * Faded
  * `aria-hidden`
  * `tabindex=-1`
* Please verify keyboard and screen-reader behavior when toggling between Cloud and Local.

### Env Var Encryption

* Values are encrypted at rest through `safeStorage` when available.
* Keyring-less Linux falls back to plaintext.
* Legacy plain-string records still load.
* Decrypt failures are dropped instead of being shown as ciphertext.
* Please confirm mask/reveal behavior in the settings editor for sensitive key names.

### Global Settings vs IPP Scope

* Global Settings popup is desktop-only.
* It does not include install-scoped Update Channel cards.
* Per-install channel and update actions remain in IPP.
* Worth confirming nothing regressed for install-backed hosts opening Global Settings from the title bar.

### IPP Default Selection

* Install-less hosts, such as Dashboard or Chooser, now auto-select the first install.
* This prevents the settings pane from opening empty.
* Selection is persisted to main on mount.

### Update Channel and Check Feedback

* Switching `updateChannel` triggers a background `check-update`.
* Inline action buttons show a spinner.
* A minimum busy duration of around `700ms` is used so very fast responses still feel acknowledged.
* The redundant **Check for updates** button is hidden when an update is already surfaced.

### `blurOverlay` on `BaseModal`

* `blurOverlay` is now opt-in.
* Only modals that pass `blur-overlay` get the frosted scrim, for example:

  * Terms
  * Downloads
* Any modal that previously relied on blur implicitly now needs an explicit flag.

### Tooltip Behavior

* Tooltips use capture-phase scroll and resize handling.
* Please sanity-check repositioning inside nested scroll surfaces, including:

  * Takeovers
  * Settings panes
  * IPP

---

### Menu

* From an install:

  * Hamburger menu shows exactly four items.
  * **Exit Window** prompts before closing.
  * **Exit All Windows** prompts even when only one window is open.
* From Dashboard:

  * Full menu is still shown.
  * Bottom item reads **Exit All Windows**.

### Current Pill

* Open picker from inside install A.
* Row A should show **Current**.
* Other rows should show relative time like **Xh ago**.
* Click row B.
* Row A should still show **Current** because the pill follows the host, not the selected row.

### Status

* Disk Usage should match the actual install footprint.
* Cross-check with `du -sh`.
* Long install paths should show a leading ellipsis while keeping the last folder readable.

### Tabs

* Picker preview should show tabs in this order:

```text
Update | Startup Args | Snapshots | About
```

### More Menu

* Open the footer More menu.
* **Forget** should appear above **Uninstall**.
* Right-click an install on the Dashboard.
* Labels should also read **Forget** and **Uninstall** there.
* Both confirm dialogs should use the new labels.

### Orange Dot

* Pick an install with a known old version.
* The row should show an orange dot.
* Launch it.
* The dot should stay orange instead of turning green.
* Launch a different up-to-date install.
* That install should show the green running dot.

### Env Disclaimer

* Open the Config/Startup Args env section.
* Only the short warning chip should appear.
* There should be no expand/collapse warning.

### Auto-scroll

* Add five env vars in a row.
* Each new row should scroll into view.
* The Name input should be focused.

### Env Bug

* Add a row.
* Type only into the Value field.
* Click away to a different app to copy a key.
* Return to the app.
* The row should still be present and the value should be preserved.

### Home Icon

* Open picker from inside an install.
* Home icon should appear on the left with no border/background.
* Divider should appear between Home and chips.
* Click Home.
* It should return to Dashboard.
* Open picker from Dashboard.
* Home icon should not appear.

### Restart-to-Apply State

* Open the picker on a running install.
* Change `Launch Mode` from Window to Console.
* Dropdown should update instantly.
* **Restart to apply** pill should appear next to the field.
* Footer button should turn yellow.
* Change it back to Window.
* Pill and yellow button should disappear.
* Repeat the same check for:

  * `Browser Partition`
  * `Port Conflict`
  * `Use Shared Paths`
  * Env Vars
  * Startup Args
* Change a setting.
* Click another install in the left list.
* Click back to the original install.
* Pill and yellow button should still be present.
* Click another install with no edits.
* That install should not show a pill or yellow button.
* Click Restart.
* After the install relaunches, pill and yellow button should clear.
* If possible, simulate failure:

  * Kill main while editing, or make the IPC throw.
  * Confirm the field snaps back.
  * Confirm a red error pill appears with the error message.
* Change a setting on a stopped install.

  * No pill should appear.
  * No yellow button should appear.
  * The next launch should pick up the value normally.

-----

### Storage Tab

* Open the picker and confirm the tab order reads:

```text
Update | Config | Snapshots | Storage | About
```

* Open a slow-loading instance.

  * Storage tab should be hidden during **Loading...**
  * It should appear with the other tabs once loading finishes.

* On the Storage tab, add a directory or toggle **Use Shared Directories**.

  * The top banner should turn yellow.
  * The icon and border should also use the yellow warning state.
  * The message should switch to the restart-required application message.

* With two or more model directories, click **Make primary** on a non-primary directory.

  * The row that moves to the top should briefly flash yellow.

* Click the folder open icon on any directory row.

  * It should open that folder in the OS file manager.

* Click **Add Directory**.

  * The folder picker should appear.
  * The selected folder should show up in the list.


src/main/lib/envVarsCrypto.ts
deepme987, line 1: > Windows: Encryption keys are generated via [DPAPI](https://learn.microsoft.com/en-us/windows/win32/api/dpapi/nf-dpapi-cryptprotectdata). As per the Windows documentation: "Typically, only a user with the same logon credential as the user who encrypted the data can typically decrypt the data". Therefore, content is protected from other users on the same machine, but not from other apps running in the same userspace.

It sounds "good" but I'd be vary of adding it NOW because it may fail

deepme987 (dismissed):

src/main/lib/envVarsCrypto.ts
deepme987: One suggestion: remove from this PR ; we can have it in our follow-up bucket for post-GA

Kosinkadink (approved):

PR #5, opened 2026-02-20
proposal(12): Sentry

christian-byrne:
## Proposal #12 — Sentry Error Tracking

Adds `@sentry/electron` (v7.8.0) for crash reporting and error tracking across main and renderer processes. **This is error tracking only — NOT telemetry or usage analytics.**

### What This PR Does

1. **`lib/sentry.js`** — New module that initializes Sentry only when the user has opted in via settings. Uses `beforeSend` to scrub filesystem paths for privacy.
2. **`main.js`** — Requires `lib/sentry.js` as the very first line (before `app.whenReady()`) so unhandled errors are captured from startup.
3. **`lib/ipc.js`** — Adds `errorReporting` toggle to Settings → General (line 449), using the existing boolean field renderer.
4. **`locales/en.json` / `zh.json`** — Adds i18n key: "Send anonymous error reports (requires restart)".
5. **`package.json`** — Adds `@sentry/electron` dependency.
6. **`.github/proposals/proposal-sentry.md`** — Full migration proposal with architecture, privacy analysis, compatibility, bundle size, and phased rollout plan.

### Key Design Decisions

- **Off by default** — `errorReporting` defaults to `false`. Sentry.init() is never called unless the user explicitly enables it.
- **Privacy-first** — `sendDefaultPii: false`, `tracesSampleRate: 0`, `beforeSend` strips absolute filesystem paths from error messages.
- **No bundler required** — SDK ships CJS entry points; works with the current plain-script architecture.
- **Placeholder DSN** — Uses `https://examplePublicKey@o0.ingest.sentry.io/0`. Must be replaced with a real DSN before shipping.
- **Electron v40 compatible** — SDK supports `electron >= v23`.

### What IS / IS NOT Collected

| Collected (when enabled) | NOT collected |
|---|---|
| Stack traces from unhandled errors | ❌ No usage analytics or telemetry |
| Electron/OS version, architecture | ❌ No user-identifiable information |
| Breadcrumbs (console, navigation) | ❌ No filesystem paths (scrubbed) |
| App version for release tracking | ❌ No workflow data or model names |

### Bundle Size

~1–2 MB added to asar — negligible vs Electron baseline (~150+ MB). No native modules.

### Dependencies

None (independent proposal). Benefits from Proposal #1 (electron-vite) for future source map uploads.

### Files Changed

| File | Change |
|---|---|
| `lib/sentry.js` | **New** — Sentry init with opt-in guard and path scrubbing |
[...]

PR #18, opened 2026-02-20
proposal(2): TypeScript + Typed IPC

christian-byrne:
## Summary

Migration proposal to add TypeScript and type-safe IPC to ComfyUI Launcher.

**Depends on:** Proposal #1 (electron-vite)

## What's Included

- **Proposal document** (`.github/proposals/proposal-typescript-typed-ipc.md`) — full analysis of all 36 IPC channels (27 invoke + 9 event), approach comparison (electron-trpc vs @electron-toolkit/typed-ipc vs manual shared types), migration plan, effort estimates, and tradeoffs
- **TypeScript configuration** — `tsconfig.json` (root) + `tsconfig.{main,preload,renderer}.json` for each Electron process
- **Shared IPC types** (`src/shared/ipc-types.ts`) — typed channel definitions for every IPC channel in `preload.js` and `lib/ipc.js`, with domain types and type-safe helper types (`TypedInvoke`, `TypedHandle`, `TypedSend`, `TypedOn`)
- **Renderer declarations** (`src/renderer.d.ts`) — `window.api` type declarations derived from the shared types
- **PoC conversion** (`settings.ts`) — `settings.js` converted to TypeScript as a working demonstration
- **Typecheck script** — `npm run typecheck` passes with zero errors

## Approach

**Shared type definitions with zero runtime dependencies.** No electron-trpc (runtime overhead, 7-layer abstraction for simple getters), no third-party typed-ipc libraries. Just TypeScript interfaces that enforce correctness at compile time and vanish at runtime.

## Key Decisions

| Decision | Rationale |
|---|---|
| Reject electron-trpc | 28KB runtime overhead, 133% complexity tax for simple IPC. [Detailed analysis](https://seed.hyper.media/hm/z6MkuBbsB1HbSNXLvJCRCrPhimY6g7tzhr4qvcYKPuSZzhno/tech-talks/the-case-against-electron-trpc-when-type-safety-becomes-a-performance-tax) |
| Reject @electron-toolkit/typed-ipc | 360 weekly downloads, requires replacing preload pattern |
| Manual shared types | Zero dependencies, zero bundle impact, full type safety |
| Renderer deferred | Can't convert `renderer/*.js` until bundler (Proposal #1) replaces `<script>` tags |
| tsgo mention | TypeScript 7 native compiler (~10× faster) is a free upgrade once TS is in place |

## Files Changed

| File | Purpose |
|---|---|
| `.github/proposals/proposal-typescript-typed-ipc.md` | Full proposal |
[...]

PR #591, opened 2026-05-25
fix(#582): UX rollup — stop-confirm modal, in-place relaunch, copy/update destination, settings staleness, delete latency, cloud zip capture, progress error overflow

### LINKED ISSUE #582: issues to fix
We need o catch up on the funcionality of desktop being broken. On version 0.6.4 things are okay, but since we started changing the UX a lot of functionality is broken.

The main things are, can users update as expected? Bot hfor nomrla copy and copy & update. Are the flows of modals smooth? Do we show enough about errors?

- When clicking on tabs in the center pill drawer, there is always delay to show the settings on hte right side for local installs. On first click there is a Loading..., but when clicking on the other lcaol sintalls while that right side thing is open, it no longer shows Loading... so it looks like the UI freezes
- Opening new window on WIndows (not suire about amc) takes 1.5 seconds, feels bad
- Download zip from cloud tries to open up browser for the download for some reaosn. Is there n aissue wiht capuring downloads in certain cases?
- Pressing Delete on dashboard has like a 2 secoand delay before modal pops up
- Updating issues; when trying to update across update channels (switchign form one to another), the update does nothing. Additionally, if an instnace is running, the first modal that asks to turn off the isntance makes the modals close themsleves, os you cant complete an update without clicking the central pill drawer to see the modals again. THe work on https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/pull/563 can be incorporated to smooth this out, but we probably should avoid the modals closing when not expected.
- Lifecycle tests are in progress being added on https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/pull/581 . Anything we fix should also be added to tehse tests; this PR should be merged before additional work is done, such that we can add tests as needed.
- More code is added in https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/pull/590, but I have no idea what it does, or if anything further is broken.
- When errors are displayed, there is the 'log' section and the 'red text that shows whats wrong section'. The 'red text what's wrong' before the console does not have a max height, so it can stretch the page in an ugly way. This section should have a limit and support scrollbar most likely.

Kosinkadink:
Fixes #582.

Rolls up the seven sub-fixes called out in #582 plus the supporting test-infra (`@lifecycle` Playwright suite, renderer e2e hooks) and an audit-driven refactor of the REQUIRES_STOPPED action flow.

## Fixes

| # | Issue body bullet | Fix commit(s) |
|---|------|------|
| 1 | Right pane shows stale sections when clicking tabs / switching installs ("UI freezes" after the first `Loading…`) | `5bc681c` `47a8564` `fb63a96` |
| 2 | Cross-channel `update-comfyui` was a no-op (Global Settings allowlist typo) | `5bc681c` `70f2b04` |
| 3 | Cloud zip download bounces out to the OS browser | `e95f12f` |
| 4 | Dashboard Delete ~2s latency before the confirm appears (`get-detail-sections` was running on the kebab path) | `e95f12f` `a819b82` |
| 5 | Update modal disappears mid-interaction when the install is running; no clear feedback | `d1881a7` `7d0f55e` `02fbd10` |
| 6 | Stop-confirm modal closes itself when the picker dispatches the action (popup blur tore down the modal) | `7fae0f4` `5f8643c` `d1881a7` `7d0f55e` |
| 7 | ProgressModal error block has no max-height — long stderr stretched the page | `5bc681c` `ad8f0d7` |

## Audit-driven refactor (FLOW 1 / 2 / 3)

The audit found that no REQUIRES_STOPPED action's `confirm` / `prompt` copy mentioned stopping ComfyUI; the standalone "Stop ComfyUI?" modal in `useActionGuard.checkBeforeAction` was the only user-visible warning, and it was the same one fix #6 routed around.

* **FLOW 1 — Self-stopping `apiCall` + in-place relaunch (`7d0f55e` `9944fe6`)**
  REQUIRES_STOPPED actions on a running install now wrap the `apiCall` to `stopComfyUI → wait → run-action`, and `IN_PLACE_RELAUNCH` (`update-comfyui`, `snapshot-restore`) appends `run-action('launch')` on success so the user lands back on a live ComfyUI. Centralised in `src/renderer/src/lib/stopWarning.ts`. The `cancelOperation` no-op for finished ops (`9944fe6`) closes the silent takeover→takeover swap race that otherwise tore down the relaunched session.
* **FLOW 2 — Copy & Update destination (`02fbd10`)**
  `copy` / `copy-update` / `release-update` now populate `ActionResult.newInstallationId`; `ProgressModal.handleDone` consumes it via a new `open-install-window` IPC so the destination install opens in its own window instead of swapping the source host. Misleading `copyAndUpdateMessage` copy fixed in the same change.
* **FLOW 3 — Modal stack collapse (`d1881a7`)**
  Removed the standalone Stop ComfyUI? modal from `useActionGuard`; added `errors.willStopRunning` i18n string and use `augmentActionWithStopWarning` to prepend the warning to the action's own confirm/prompt dialog (synthesising one if needed). The picker fix #6 bridge (`requiresStopped` flag in IPC payload) is dropped — the panel derives the need to stop from `actionId` + session state.

## Tests

* Unit tests for `useComfyUISettings.runAction` covering FLOW 1 + FLOW 3 behaviour (message augment, IN_PLACE_RELAUNCH stop+run+launch, no-relaunch on `ok:false`, no-op when not running, inline self-stop path).
* `progressStore.cancelOperation` regression test for the finished-op no-op.
* `@lifecycle` Playwright suite:
  * `picker-stop-confirm.test.ts` — picker forwards a REQUIRES_STOPPED action; panel-side apiCall self-stops + dispatches the original action; skip-self-stop branch covered.
  * `copy-update-destination.test.ts` — `ProgressModal.handleDone` consumes `newInstallationId` and fires `open-install-window`.
  * Plus per-fix coverage: `picker-settings-staleness` (fix #1), `global-settings-update-action` (fix #2), `dashboard-delete-flow` (fix #4), `progress-error-overflow` (fix #7).

## Pre-merge checks

* `pnpm run typecheck` — green
* `pnpm run lint` — green
* `pnpm run build` — green
* `pnpm run test` — 1133 tests pass
* `pnpm playwright test --project=lifecycle` — 39 tests pass
* `pnpm playwright test --project=windows` — 36 tests pass

### DIFF
--- a/e2e/copy-update-destination.test.ts
+++ b/e2e/copy-update-destination.test.ts
@@ -1,0 +1,114 @@
+/**
+ * Copy / copy-update destination focus (issue #582 FLOW 2).
+ *
+ * `handleCopy` / `handleCopyUpdate` / `handleReleaseUpdate` now return
+ * `{ ok: true, navigate: 'list', newInstallationId }` and
+ * `ProgressModal.handleDone` reads `newInstallationId` to call
+ * `window.api.openInstallWindow(newInstallationId)` — opening the
+ * newly-created destination install in its own window without
+ * swapping the source host.
+ *
+ * Test strategy: drive the panel's existing show-progress chain with
+ * a synthetic apiCall that resolves to a success payload carrying
+ * `newInstallationId`. The actual copy handler is bypassed (it's covered
+ * by integration tests + the field is populated unconditionally on
+ * the success path of all three handlers); this exercises the
+ * renderer-side handleDone branch + the main-side
+ * `open-install-window` IPC + the focus-existing-or-open-new
+ * behaviour.
+ */
+
+import os from 'node:os'
+import path from 'node:path'
+import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises'
+import { test, expect } from '@playwright/test'
+import { launchApp, type AppContext } from './launchApp'
+import { expectChooserVisible } from './support/chooserHelpers'
+import {
+  getIpcInvocations,
+  resetIpcInvocations,
+} from './support/devHooks'
+
+let ctx: AppContext
+let sourcePath: string
+
+const SOURCE_ID = 'inst-copy-source-test'
+const SOURCE_NAME = 'Source Install'
+const DEST_ID = 'inst-copy-destination-test'
+const MARKER_FILENAME = '.comfyui-desktop-2'
+
+test.describe.configure({ mode: 'serial' })
+
+test.beforeAll(async () => {
+  sourcePath = await mkdtemp(path.join(os.tmpdir(), 'comfyui-launcher-copy-dest-e2e-'))
+  await mkdir(sourcePath, { recursive: true })
+  await writeFile(path.join(sourcePath, MARKER_FILENAME), SOURCE_ID)
+
+  ctx = await launchApp({
+    settings: { firstUseCompleted: true, telemetryEnabled: false },
+    installations: [
+      {
+        id: SOURCE_ID,
+        name: SOURCE_NAME,
+        installPath: sourcePath,
+        sourceId: 'standalone',
+        status: 'installed',
+      },
+    ],
+  })
+  await expectChooserVisible(ctx.panel)
+})
+
+test.afterAll(async () => {
+  await ctx?.cleanup()
+  if (sourcePath) await rm(sourcePath, { recursive: true, force: true })
+})
+
+test.beforeEach(async () => {
+  await resetIpcInvocations(ctx.app, 'open-install-window')
+})
+
+test('Copy success opens the destination install in a new window @lifecycle', async () => {
+  // Drive ProgressModal with a synthetic copy result carrying the
+  // destination install id. The renderer's `handleDone` consumes
+  // `op.result.newInstallationId` and calls `openInstallWindow`.
+  await ctx.panel.evaluate<void>(
+    `window.__e2eRenderer.injectProgressSuccess({
+      installationId: ${JSON.stringify(SOURCE_ID)},
+      title: 'Copy — ' + ${JSON.stringify(SOURCE_NAME)},
+      newInstallationId: ${JSON.stringify(DEST_ID)},
+    })`,
+  )
+
+  // The brand loader auto-closes 700ms after `finished`, then
+  // handleDone fires. Poll the recorded IPC invocations.
+  await expect
+    .poll(async () => (await getIpcInvocations(ctx.app, 'open-install-window')).length, {
+      timeout: 5_000,
+      intervals: [100, 250],
+    })
+    .toBeGreaterThanOrEqual(1)
+
+  const calls = await getIpcInvocations(ctx.app, 'open-install-window') as
+    { installationId?: string }[]
+  expect(calls.length).toBe(1)
+  expect(calls[0]?.installationId).toBe(DEST_ID)
+})
+
+test('No newInstallationId → no open-install-window call @lifecycle', async () => {
+  // Same shape but no newInstallationId — should NOT trigger
+  // openInstallWindow. Guards the conditional branch in handleDone
+  // against accidentally firing for non-copy success paths.
+  await ctx.panel.evaluate<void>(
+    `window.__e2eRenderer.injectProgressSuccess({
+      installationId: ${JSON.stringify(SOURCE_ID)},
+      title: 'Other op — ' + ${JSON.stringify(SOURCE_NAME)},
+    })`,
+  )
+
+  // Wait for handleDone's auto-close window to elapse (≥700ms).
+  await new Promise((r) => setTimeout(r, 1_500))
+
+  const calls = await getIpcInvocations(ctx.app, 'open-install-window')
+  expect(calls.length).toBe(0)
+})
--- a/e2e/dashboard-delete-flow.test.ts
+++ b/e2e/dashboard-delete-flow.test.ts
@@ -1,0 +1,122 @@
+/**
+ * Dashboard Delete fast-path E2E (regression for issue #582 fix #4).
+ *
+ * Drives the full chooser-kebab → Delete confirm → ProgressModal →
+ * directory removed flow and asserts:
+ *
+ *   1. The confirm modal renders quickly after Delete is clicked
+ *      (no perceptible stall while a `get-detail-sections` payload is
+ *      rebuilt). The bug shipped a ~2s freeze on Windows because the
+ *      composable round-tripped through `get-detail-sections` just to
+ *      look up the `deleteAction()` shape; the fast path bypasses it.
+ *   2. `get-detail-sections` is NOT invoked during the Delete dispatch
+ *      — asserted via the test-only `__e2e.getIpcInvocations()`
+ *      counter so a future regression that re-introduces the
+ *      roundtrip fails this test rather than re-introducing the
+ *      latency.
+ *   3. The install directory is gone from disk after Confirm, and the
+ *      tile disappears from the chooser.
+ */
+
+import os from 'node:os'
+import path from 'node:path'
+import { access, mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises'
+import { test, expect } from '@playwright/test'
+import { launchApp, type AppContext } from './launchApp'
+import { expectChooserVisible } from './support/chooserHelpers'
+import { byTestId, TID } from './support/testIds'
+import { getIpcInvocations, resetIpcInvocations } from './support/devHooks'
+
+let ctx: AppContext
+let installPath: string
+
+const INSTALL_ID = 'inst-dashboard-delete-test'
+const INSTALL_NAME = 'Delete Fast-Path Me'
+
+/** Mirrors `MARKER_FILE` in `src/main/lib/ipc/shared.ts`. The delete
+ *  action refuses to touch a directory whose marker file is missing. */
+const MARKER_FILENAME = '.comfyui-desktop-2'
+
+async function pathExists(p: string): Promise<boolean> {
+  try {
+    await access(p)
+    return true
+  } catch {
+    return false
+  }
+}
+
+test.describe.configure({ mode: 'serial' })
+
+test.beforeAll(async () => {
+  installPath = await mkdtemp(path.join(os.tmpdir(), 'comfyui-launcher-dash-delete-e2e-'))
+  await mkdir(installPath, { recursive: true })
+  await writeFile(path.join(installPath, MARKER_FILENAME), INSTALL_ID)
+
+  ctx = await launchApp({
+    settings: { firstUseCompleted: true, telemetryEnabled: false },
+    installations: [
+      {
+        id: INSTALL_ID,
+        name: INSTALL_NAME,
+        installPath,
+        sourceId: 'standalone',
+        status: 'installed',
+      },
+    ],
+  })
+  await expectChooserVisible(ctx.panel)
+})
+
+test.afterAll(async () => {
+  await ctx?.cleanup()
+  if (installPath) await rm(installPath, { recursive: true, force: true })
+})
+
+test('chooser shows the seeded tile @lifecycle', async () => {
+  await ctx.panel.waitForSelector(byTestId(TID.dashboardTile(INSTALL_ID)), { timeout: 10_000 })
+  expect(await ctx.panel.textOf(`${byTestId(TID.dashboardTile(INSTALL_ID))} .chooser-tile-name`)).toBe(INSTALL_NAME)
+})
+
+test('Delete from kebab opens confirm without invoking get-detail-sections @lifecycle', async () => {
+  // Clear any cumulative invocation history so the assertion measures
+  // only what fires during the Delete dispatch itself.
+  await resetIpcInvocations(ctx.app, 'get-detail-sections')
+
+  // Open the kebab menu on the seeded tile.
+  const kebabClicked = await ctx.panel.click(byTestId(TID.dashboardTileKebab(INSTALL_ID)))
+  expect(kebabClicked, 'kebab button click dispatched').toBe(true)
+
+  // The shared ContextMenu portals to <body>; its Delete item carries
+  // a `context-menu-item-delete` test id.
+  await ctx.panel.waitForVisible(byTestId(TID.contextMenuItem('delete')), { timeout: 5_000 })
+  const deleteClicked = await ctx.panel.click(byTestId(TID.contextMenuItem('delete')))
+  expect(deleteClicked, 'delete menu item click dispatched').toBe(true)
+
+  // The confirm modal must appear — it's a `BaseAlert` simple-confirm.
+  await ctx.panel.waitForVisible(byTestId(TID.baseAlertAction), { timeout: 5_000 })
+
+  // Fast-path assertion: the Delete dispatch must NOT have invoked the
+  // `get-detail-sections` IPC. The bug we're guarding against
+  // re-introduces a roundtrip there to look up the action's confirm
+  // payload, costing ~2s on Windows before the modal can paint.
+  const invocations = await getIpcInvocations(ctx.app, 'get-detail-sections')
+  expect(invocations, 'Delete dispatch must not call get-detail-sections').toEqual([])
+})
+
+test('Confirm removes the install directory and tile @lifecycle', async () => {
+  // Confirm the delete via the BaseAlert primary button.
+  const confirmClicked = await ctx.panel.click(byTestId(TID.baseAlertAction))
+  expect(confirmClicked, 'confirm button click dispatched').toBe(true)
+
+  // Tile disappears from the chooser once the action completes.
+  await ctx.panel.waitFor(
+    async () => !(await ctx.panel.exists(byTestId(TID.dashboardTile(INSTALL_ID)))),
+    { timeout: 30_000, message: 'deleted tile never disappeared from chooser' },
+  )
+
+  // Install directory was removed on disk.
+  await expect
+    .poll(() => pathExists(installPath), { timeout: 30_000, intervals: [250, 500, 1000] })
+    .toBe(false)
+})
--- a/e2e/global-settings-update-action.test.ts
+++ b/e2e/global-settings-update-action.test.ts
@@ -1,0 +1,101 @@
+/**
+ * Global Settings → update-comfyui allowlist regression (issue #582
+ * fix #2).
+ *
+ * The title-bar Global Settings drawer gates install-action IPCs via
+ * `GLOBAL_SETTINGS_ALLOWED_ACTIONS` in `src/main/popups/titlePopup.ts`.
+ * Before the fix the allowlist carried a bare `'update'` id that no
+ * source produces — every click on Update Now returned
+ * `{ ok: false, message: "Action 'update' is not available." }` and
+ * the popup silently swallowed the result. The fix swapped that for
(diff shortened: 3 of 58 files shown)


Kosinkadink:
## Read-only audit on top of #591 (branch `fix/issue-582-test-infra-and-fixes`)

Pure inventory — nothing fixed here, just calling out the next tier of friction and the lifecycle-coverage holes I noticed while validating #591 against the action-runner / overlay / IPC inventory. Anything #591 already addresses (picker right-pane staleness, allowlist typo, cloud zip download capture, kebab Delete latency, cross-channel update, stop-confirm modal removal, ProgressModal error overflow, FLOWs 1/2/3, `cancelOperation` finished-op safety) is excluded.

### Annoying / simplifiable flows

1. **Snapshot Restore from picker Settings → Snapshots tab double-confirms when the install is running**
   - *Surface*: `src/renderer/src/views/comfyUISettings/SnapshotsView.vue` `handleRestore` (`@run-action` emit consumed by `useComfyUISettings.runAction`).
   - *Current*: SnapshotsView shows its own diff-preview confirm, then emits `run-action` with no `confirm` of its own. `useComfyUISettings.runAction` step 3 runs `augmentActionWithStopWarning`, which **synthesizes** a second confirm (`{ title: label, message: willStopRunning }`) because the emitted action has neither `confirm` nor `prompt`. Net effect on a running install: "Restore Snapshot? (diff …)" → click Restore → "Restore Snapshot — Will stop running ComfyUI" → click Restore again.
   - *Why annoying*: the user already confirmed the destructive op; the second modal feels like the app forgot the first click. Same pattern would bite any other surface that confirms before emitting `run-action`.
   - *Suggested simplification*: have `SnapshotsView` build its diff confirm with the willStopRunning sentence prepended (using `augmentMessageWithStopWarning`) and pass it through the emit as `confirm`, so `augmentActionWithStopWarning` mutates an existing confirm in place rather than synthesizing a new one. Alternatively skip the renderer-side confirm entirely and put the diff into `confirm.messageDetails`.

2. **Dashboard kebab "Copy Installation" and "Untrack" bypass the entire shopping-list pipeline**
   - *Surface*: `useInstallContextMenu.triggerAction` ids `copy-install` and `untrack`.
   - *Current*: both call `window.api.runAction(inst.id, 'copy' | 'remove')` directly, no confirm, no prompt, no fieldSelects, no disk-space check, no busy guard, no apiCall self-stop, no progress overlay. The in-file comment claims "source-side action def owns its native confirm dialog" — but those action defs declare `prompt`/`confirm` for the **renderer** to render. Main's handler just sees `actionData = undefined` and bails (handleCopy returns `{ ok: false, message: 'No name provided.' }`).
   - *Why annoying*: kebab Copy looks like a no-op to the user (and the error path silently swallows the `result.message` in the `catch {}` block — see #9 below). Kebab Untrack triggers a destructive action without any confirm dialog.
   - *Suggested simplification*: route both through the Delete pattern that already exists in the same composable (`opts.onShowProgress` / `opts.onManage(inst, { autoAction: 'copy' | 'remove' })`) so the source-action def's confirm/prompt/disk-check chain actually runs.

3. **`useActionGuard.checkBeforeAction` hard-codes `setTimeout(500)` after `cancelOperation`**
   - *Surface*: `src/renderer/src/composables/useActionGuard.ts:30`.
   - *Current*: After the user confirms cancelling an in-flight op, the guard calls `cancelOperation(id)` then `await new Promise(r => setTimeout(r, 500))` before returning `true`. On Windows (taskkill / Restart Manager) cancel teardown is often slower than 500 ms; on fast paths it's needless friction.
   - *Why annoying*: same family as fix #4 — an arbitrary sleep instead of a readiness signal. The next action races a still-pending op and either gets `result.running` back or hits the same REQUIRES_STOPPED guard in main.
   - *Suggested simplification*: poll `!progressStore.getProgressInfo(id) && !sessionStore.isStopping(id)` with a small loop and a deadline (mirrors `stopAndWaitForExit`); drop the sleep.

4. **`runAction` is duplicated ~280 lines across `useComfyUISettings.ts` and `DetailModal.vue`**
   - *Surface*: `useComfyUISettings.runAction` (steps 1–10) and `DetailModal.runAction` (same steps, near-verbatim).
   - *Current*: FLOWs 1/2/3 had to be applied to both; they were. But any subsequent dispatcher change (a new disk-check action, a new IN_PLACE_RELAUNCH entry, an extra augmentActionWithStopWarning callsite) has two places to keep in sync. `useListAction` is a third partial copy (subset — no fieldSelects/prompt/select/disk-check, but same self-stop + relaunch wrapping).
   - *Why annoying*: high drift risk; the audit thread already had to call out divergences between these three. The next bug fix will land in one and miss the other two.
   - *Suggested simplification*: extract one `runActionWithShoppingList(action, ctx)` helper from `useComfyUISettings.ts` (parameterised on `showProgress` callback, `onNavigateList`, `inst`) and have `DetailModal.runAction` and `useListAction.executeAction` call it. The picker-show-progress branch of `useDeepLinkRouter` rebuilds the same apiCall yet a fourth time — would also benefit.

5. **First-use Migrate branch briefly exposes the dashboard between Tier 3 dismiss and Tier 2 progress mount**
   - *Surface*: `useFirstUseChain.handleFirstUseChainMigrate` — `opts.dismissTakeoverDirect()` runs synchronously before `await opts.handleShowProgress(...)` mounts the Tier 2 progress overlay.
   - *Current*: between the dismiss and the next overlay mount the dashboard underneath is visible, the file menu flips to first-use `'none'`, then gets re-asserted to `'post-consent'` after the await. Tiny window but visible on slow machines.
   - *Suggested simplification*: open the Tier 2 overlay first and rely on the existing Tier 2-over-Tier 3 collision rule to silently kill the takeover (today the rule prompts on a takeover-with-`onCancel`; first-use's takeover doesn't set one, so it'd swap silently). Or mount migration as a Tier 3 `'update'`-style takeover the way update-while-running already does.

6. **`useInstallContextMenu.triggerAction` swallows action failures**
   - *Surface*: kebab `reveal-in-folder` / `copy-install` / `untrack` / fallback `delete` all use `try { await window.api.runAction(...) } catch {}` with comments like "Source action surfaces its own error path".
   - *Current*: the main-side handlers return `{ ok: false, message }` (rejected promise only on truly exceptional failure — not on action errors). The `catch {}` only catches the rejected-promise case; the `{ ok: false, message }` branch is dropped on the floor and the user sees nothing.
   - *Suggested simplification*: surface `result.ok === false` via `modal.alert({ title: action.label, message: result.message })` — the same shape `useListAction` already uses.

7. **`useListAction.executeAction` only busy-guards REQUIRES_STOPPED actions**
   - *Surface*: `useListAction.ts:39` — `if (requiresStoppedGuard) actionGuard.checkBeforeAction(...)`.
   - *Current*: Non-REQUIRES_STOPPED actions (snapshot-save, snapshot-export, snapshot-import, open-folder, dismiss-error) fire without consulting the busy guard. A snapshot-save fired mid-launch races the launch op without any user feedback.
   - *Suggested simplification*: gate every action runner on busy state (`progressStore.getProgressInfo(id) !== null` is the existing signal), regardless of REQUIRES_STOPPED. Or have main reject conflicting ops uniformly and surface `result.running` consistently.

8. **`useFirstUseChain.handleFirstUseComplete` opens Cloud install but doesn't dismiss takeover on `'missing-action'`**
   - *Surface*: `handleFirstUseComplete` Cloud branch — only the `'focused-running'` outcome explicitly calls `dismissTakeoverDirect()`. The `'launched'` outcome relies on the launch action's `showProgress: true` swapping the takeover, and `'missing-action'` relies on `onMissingLaunchAction = dismissTakeoverDirect`.
   - *Current*: works today, but the comment-vs-code separation means if `performChooserLaunch` ever returns a fourth outcome the first-use takeover would strand. Lower priority.
   - *Suggested simplification*: dismiss unconditionally before `performChooserLaunch` (the takeover would just be re-mounted as part of the launch's Tier 3 connect-progress takeover) — or guard via `if (outcome !== 'launched') dismissTakeoverDirect()`.

9. **Picker "Open Folder" hides the popup before firing the action**
   - *Surface*: `titlePopup.ts` `comfy-titlepopup:open-install-action` handler (`hideTitlePopup(...)` before `bindings.sendToPanelDeferred(...)`).
   - *Current*: every More-menu action hides the popup, including non-modal ones like Open Folder where the user might still want to keep poking at the picker. Reveal-in-folder also loses the user's mouse anchor.
   - *Suggested simplification*: keep popup visible for actions with no modal/progress side-effect (`reveal-in-folder`, `dismiss-error`).

10. **`useChooserHandoff.performPickerLaunch` short-circuits `'focused-running'` without dismissing the picker popup itself**
    - *Surface*: `useChooserHandoff.performPickerLaunch:203` — focuses the running window but does nothing about the picker popup (which has already auto-blurred but may have a forwarded progress request mid-flight if the user double-clicked tabs).
    - Lower-priority follow-up — confirm via the IPC-trace fixture whether double-firing is reachable.

11. **`progressStore.startOperation` reuses an existing in-flight op, but `cancelOperation` semantics changed under it**
    - With the `cancelOperation` finished-op no-op landing in this PR, the comment on `startOperation` ("reuses an existing in-flight op if present") becomes load-bearing for retry paths (`handleReboot`, `handleKillProcess`, `handleUseNextPort`) — they call `startOperation` with the SAME `installationId` while the previous op is finished. Make sure tests pin this behaviour so a future "always create a fresh op" refactor can't silently break port-conflict retry.

12. **Migration confirm has two render surfaces (`'modal'` / `'takeover'`) with subtly different copy paths**
    - *Surface*: `useMigrateAction.confirmMigration({ surface })`. First-use uses `'takeover'`, DetailModal and the in-dashboard MigrationBanner default to `'modal'`. The shopping-list (`useComfyUISettings.runAction` step 2) goes through the modal path because it doesn't pass `surface`.
    - *Why annoying*: same data, two render paths. If the takeover variant ever drifts (an extra checkbox, different default), three callsites have to pick which one to invoke.
    - *Suggested simplification*: make the surface a hard-coded property of the entrypoint, not a per-call opt — or merge the two variants behind one component that picks its chrome from the calling tier.

13. **Cloud zip download capture (#3) lives only on the cloud install path**
    - *Surface*: download-shelf wiring around the cloud zip download.
    - *Current*: snapshot-import / release-update / standalone install all do long-running downloads through the same `download()` primitive but don't surface in the downloads shelf. Worth confirming via a search across `makeSendProgress` / `download` callsites.
    - *Suggested simplification*: lift the capture into the shared `download()` helper rather than wiring it per-action.

---

### Missing `@lifecycle` test coverage

Inventory based on the 14 existing `e2e/*lifecycle*`, `dashboard-delete-flow`, `global-settings-update-action`, `picker-stop-confirm`, `picker-settings-staleness`, `copy-update-destination`, `progress-error-overflow` tests. Each gap below is something the IPC handler inventory or renderer composable inventory exposes but no `@lifecycle` test drives end-to-end.

1. **Plain `copy` action (no update)** — `handleCopy` in `src/main/lib/ipc/sessionActions/copy.ts`. Today only `copy-update` has a lifecycle test (`copy-update-destination.test.ts`). Assert: after Copy, the new install directory exists on disk, the new install id appears in `getInstallations()`, `open-install-window` IPC fires with the new id, and the source install's directory is unchanged.

2. **`release-update` success path** — `handleReleaseUpdate` (download → install → postInstall → `migrate-from`). Assert: new install dir exists with the picked release's marker; `migrate-from` copied customNodes + models + input + output across; `newInstallationId` opens in a new window; source install untouched.

3. **`release-update` failure cleanup path** — same handler's `migrateError` branch that calls `installations.remove(entry.id)` + `deleteDir(destPath)`. Assert: when migrate-from throws, the partial install dir is removed from disk AND the registry, source install untouched, ProgressModal shows the error message.

4. **`migrate-to-standalone` full progress op** — only the *preview* is tested today (`lifecycle-migrate.test.ts`). The actual `runAction('migrate-to-standalone', ...)` against the Legacy Desktop install isn't exercised end-to-end. Assert: legacy install marked migrated, standalone install registered, files copied, target install launchable.

5. **`migrate-from` invoked directly** (not via `release-update`'s wrapper) — there's no direct UI surface today but the source.handleAction is reachable via the cross-channel update flow. Assert via the `runAction('migrate-from', { sourceInstallationId, ... })` IPC.

6. **Port-conflict resolution (kill-port + use-next-port)** — `handleKillProcess` / `handleUseNextPort` in `ProgressModal.vue:604,619`. Seed a process listening on the install's port, drive `launch`, assert the port-conflict footer renders; then assert (a) Use Next Port re-launches at `portOverride`, (b) Kill Process invokes `killPortProcess` then re-runs the original `op.apiCall`.

7. **Launch failure → Reboot retry** — `handleReboot` in `ProgressModal.vue:534`. Inject a launch failure via `__e2eRenderer.injectProgressError`, assert the error banner; click Reboot; assert `op.apiCall` re-runs and the second attempt succeeds.

8. **Synthetic `restart` id (stop → launch)** — `useComfyUISettings.runAction` / `DetailModal.runAction` `pinBottomActions` swap. Launch ComfyUI, click the swapped Restart button, assert one continuous "Restarting ComfyUI" ProgressModal title and that the resulting `instance-started` carries a new pid (or new session id) for the same installationId.

9. **Snapshot export** — `runAction('snapshot-export', { file })` and `runAction('snapshot-export-all')`. No test drives the saveDialog → file-write path. Assert: chosen file exists on disk afterwards and contains valid zip headers.

10. **Snapshot import** — `LoadSnapshotModal.vue` + the corresponding IPC. Assert: importing a snapshot file creates a new install or restores into an existing one (whichever the modal path supports).

11. **Picker-driven launch** (`picker-pick-install` branch) — `useDeepLinkRouter` + `useChooserHandoff.performPickerLaunch`. `picker-stop-confirm.test.ts` covers picker-show-progress for action dispatch but NOT a plain pick → launch. Assert: clicking a picker row for a stopped install fires `runAction('launch')`, instance-started arrives, host doesn't close (picker variant skips `prepareChooserHostHandoff`).

12. **Picker "+ New Install"** — `comfy-titlepopup:open-new-install` → `bindings.setActivePanel(parentEntryId, 'new-install')`. Assert: the new-install Tier 3 takeover mounts on the current host (regardless of host kind), and dismissing it returns to the original host body.

13. **Picker More-menu install action** (`picker-install-action` branch for Open Folder / Copy / Untrack / Delete) — forwards to `useInstallContextMenu.triggerAction` on the panel. Assert at least one of these end-to-end, e.g. Delete from picker fires the same confirm + `onShowProgress` path as the dashboard kebab.

14. **Deep links** — `comfy://install-update/<id>` → picker on Update tab; `comfy://open-settings?tab=comfy` → picker on Config tab; `comfy://open-settings?tab=global` → openGlobalSettings. Assert: each deep link, replayed via the `onPanelTriggerOverlay` IPC, lands the correct overlay.

15. **First-use Local branch end-to-end** — `useFirstUseChain.handleFirstUseChainLocal` → new-install takeover → install op → auto-launch watcher. `lifecycle.test.ts` does the Local + install pieces but not as a first-use *chain* (no `chainingFirstUseToNewInstall` flag, no pending-auto-launch id capture). Assert: `markFirstUseCompleted` is called exactly once, `setFirstUseMode('post-consent')` then `'none'`, and the install ends in a running ComfyUI without an extra click.

16. **First-use Migrate branch end-to-end** — `handleFirstUseChainMigrate`. Assert: legacy install detected, migrate confirm in takeover surface, migration op runs, auto-launch fires on the resulting standalone install.

17. **File-menu Skip Onboarding** — `onFirstUseSkip` → `completeFirstUseAndDismiss`. Assert: clearing first-use bookkeeping plus dismiss of takeover; `firstUseMode === 'none'` and the chooser body is visible.

18. **`update-comfyui` IN_PLACE_RELAUNCH against a running install** — `lifecycle.test.ts:339` explicitly STOPS before running update, so FLOW 1's self-stop + auto-relaunch wrapper isn't exercised. Assert: with the install running, click Update; one continuous progress; `runAction('launch')` fires after `runAction('update-comfyui')` returns ok; instance-started arrives at the end.

19. **`snapshot-restore` IN_PLACE_RELAUNCH against a running install** — same gap as #18 for snapshot-restore.

20. **ProgressModal Return-to-Dashboard from an in-flight cancellable op** — `returnToDashboard('in_flight')` + `progressStore.cancelOperation`. Assert: confirm shown, op cancelled main-side, takeover closes, host flips back to chooser without orphaning the backend process.

21. **Cross-channel update flow** (fix #5) — picking a different channel in Settings → Update tab and asserting the update preview re-fetches without a separate click. The composable's `field.onChangeAction` branch runs `runAction('check-update')` but no test pins this.

22. **`handleCopyUpdate` post-copy update failure branch** — copy succeeded but the chained `update-comfyui` failed. Assert: the new install still exists; user-facing output banner contains "retry the update from the new installation"; the new install opens in its own window (still `newInstallationId` returned).

23. **`dismiss-error` from the chooser kebab** — `sessionStore.clearErrorInstance` path. After seeding an error instance, click Dismiss in the kebab and assert `errorInstances` no longer contains the id and the kebab item disappears.

24. **Restart of an errored install** — seed an error, click Reboot in ProgressModal, assert recovery. Distinct from #7 (failure-then-retry) because the error already exists before the action begins.

25. **`useActionGuard.checkBeforeAction` confirm-cancel flow** — driver: seed an in-flight launch op, fire another REQUIRES_STOPPED action, assert the "Operation in progress" modal appears, click Cancel Operation, assert `cancelOperation` IPC fires and the new action proceeds. None of the picker/kebab/settings tests exercise this branch today.

---

Happy to drill into any item before opening fix PRs — most of the "Annoying" entries are small isolated fixes; the lifecycle gaps are independent so they can be picked up one-by-one without ordering constraints.

Kosinkadink:
### Follow-up UX nits (kebab styling + naming)

Three more items I missed on the first pass:

1. **Dashboard kebab — Delete and Untrack should render as destructive (red)** — both items today land in `useInstallContextMenu.getMenuItems` as plain `ContextMenuItem` entries with no danger styling. The type itself (`src/renderer/src/types/context-menu.ts`) doesn't even carry a `danger`/`style` field. Both actions destructively change the registry (and Delete the disk), so they should match the danger-button visual treatment the modal `confirmStyle: 'danger'` uses elsewhere.
   - *Suggested simplification*: add `danger?: boolean` to `ContextMenuItem`; have the ContextMenu component render `danger` items with the red destructive tint; set `danger: true` on `id: 'delete'` and `id: 'untrack'` in `getMenuItems`.

2. **"Untrack" displayed string should be "Forget"** — `actions.untrack: 'Untrack'` in `src/renderer/src/lib/i18nMessages.ts:140`. "Untrack" reads as jargon; "Forget" tells the user what's actually happening (the app forgets the install, the files stay on disk). The action id stays `untrack` (and the source-side `'remove'` runAction stays untouched) — only the user-facing label changes. Worth re-checking the confirm-message copy too so it doesn't still say "untrack" inside the modal.

3. **"Add Existing Install" / "Track Existing Install" naming mismatch** — chooser entry CTA reads "Add Existing Install" (`i18nMessages.ts:75 addExistingInstall: 'Add Existing Install'`) but the wizard it opens reads "Track Existing Install" in both its title and `grandTitle` (`locales/en.json:297-298`). Same rename rationale as #2 — "Add Existing Install" is the surface the user clicked, the wizard heading should match. Also worth a sweep of `DESIGN_PROCESS.md` / docs that say "Track Existing" for any user-facing strings that still leak through.

PR #593, opened 2026-05-25
fix: Progress Bar Unification

MaanilVerma:
## Summary

Unifies the brand progress takeover so install→launch chains read as one continuous 0→100% journey instead of the install bar saturating at 100% and stalling through launch. Adds `chainSpan` plumbing from `useFirstUseChain` through the progress store into `ProgressModal`, maps install to 0–70% and launch to 70–100%, and replaces the rolling launch caption with a five-step stepper plus polished substatus formatting.

## General Status:

* **No functional regressions + all test cases pass.**

## Progress Bar Unification:

* Added **`chainSpan`** on `ShowProgressOpts` / `Operation` (`'install' | 'launch' | null`) so chained ops can share one visual progress arc.
* **Install leg** (`chainSpan: 'install'`) maps the existing stepped/global progress to **0→70%** — the bar never visually completes before launch begins.
* **Launch leg** (`chainSpan: 'launch'`) maps launch progress to **70→100%**, picking up where the install left off.
* **Standalone ops** (no `chainSpan`) keep their existing 0→100% behaviour unchanged.

## First-Use / Auto-Launch Chain:

* `useFirstUseChain` stamps `chainSpan: 'install'` on:
  * Express first-use install ops
  * Any op with `autoLaunchOnFinish: true`
* When the auto-launch watcher fires `performChooserLaunch`, the subsequent launch op is stamped `chainSpan: 'launch'`.
* `pendingChainedLaunch` is cleared in `.finally()` so a launch that never reaches `show-progress` (missing-action / focused-running) cannot leak the flag into unrelated ops.
* `usePanelOverlays` forwards `chainSpan` into `progressStore.startOperation`.

## ProgressModal UX:

* **`unifiedPercent`** — single bar math across install, chained launch, and standalone launch.
* **Launch stepper** — five rows (security scan → mount libraries → GPU → custom nodes → starting server) with done / active / pending states and spinner on the active row; replaces the old rolling caption for brand launch ops.
* **`launchPercent`** — slot-weighted bar interpolation within each launch step so the bar moves between step transitions even when advancement is timer/stdout-driven.
* **Substatus polish** — locale-groups large byte counts (`2100 MB` → `2,100 MB`), trims collapsed `0s remaining` tails.
* Launch ops show the **stepper + unified bar**; non-launch ops keep the **determinate bar + percent label** (indeterminate animation suppressed for chained install legs).

## Tests:

* `progressStore` — `chainSpan` persistence and no regression in `globalProgressFor`.
* `useFirstUseChain` — install/launch stamping, flag cleanup on missing-action outcomes.
* `ProgressModal` — unified bar mapping, launch stepper states, substatus formatting, updated success/cancel assertions for the redesigned footer.

## Review Focus

- **70 / 30 split** — Install caps at 70%, launch starts at 70%. Worth a feel-check on express install and auto-launch-on-finish flows: does the seam between the two takeover legs feel continuous, or does the bar jump/stall at the handoff?
- **`pendingChainedLaunch` lifecycle** — Flag is set before `performChooserLaunch` and consumed in `onShowProgress`; `.finally()` clears it if launch never emits `show-progress`. Confirm no mis-stamping on a second unrelated op after a failed/missing launch.
- **Install bar vs launch stepper** — Chained install still renders the horizontal bar (0–70%); chained launch swaps to the five-row stepper (70–100%). Verify the UX reads as one journey, not two disconnected surfaces.
- **`launchPercent` interpolation** — Bar motion inside a step is time-budgeted (`LAUNCH_STEP_BUDGET_MS`), while step advancement is still timer + stdout-driven. Quick check that the bar never outruns the active step label.
- **`globalProgressFor` unchanged** — `chainSpan` remapping lives in the view only; stepped install ops still report raw weighted percent from the store. Confirm no double-scaling if main-side phase weights shift.
- **Standalone ops** — Delete / update / snapshot / generic ops with no `chainSpan` should behave exactly as before (including indeterminate phases).
- **Substatus formatter** — Byte-count grouping is scoped to unit suffixes (`MB`, `GB`, etc.); sanity-check rich download strings don't mangle ports, PIDs, or ratios.


Kosinkadink:
superceded by another pr

PR #611, opened 2026-05-27
fix(install): refresh R2 manifests and tighten post-install version c…

### LINKED ISSUE #610: Latest Stable option in new install dropdown does NOT always get the latest stable version
- Experienced when no other installs were present on Windows. In the dropdown of Latest Stable, the git-obtained latest tag was not included, as if it was not requested at all. Seems to not even try to get latest when executing in that scenario.
- Latest Stable option should always attempt to get latest from git; suspicion that there are cases where the app might think there is nothing newer so it doesn't even attempt to get latest stable tagged at install time.

MaanilVerma:
Closes https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/issues/610

## Summary

Fixes standalone installs getting stuck on old ComfyUI versions.

- **Install wizard** — Always fetches fresh R2 manifest data (`latest.json` and `releases.json`) instead of using a cached copy. A stale cache could hide a new release, so new installs would only see older versions.
- **Post-install update check** — Compares the actual installed version (`ComfyVersion`) instead of formatted tag strings. If the latest version can't be fetched, it now says "Skipped — could not verify latest version" instead of wrongly showing "Already up to date".

## Test plan

- [ ] Open the install wizard after a new standalone release ships — the new version appears in the list
- [ ] Run a fresh standalone install when GitHub is unreachable during the post-install check — shows "Skipped", not "Already up to date"
- [ ] Run a fresh install when already on latest — still shows "Already up to date"

### DIFF
--- a/src/main/lib/fetch.ts
+++ b/src/main/lib/fetch.ts
@@ -70,9 +70,13 @@ function _headerString(value: string | string[] | undefined): string | undefined
   return Array.isArray(value) ? value[0] : value
 }
 
-export function fetchJSON(url: string): Promise<unknown> {
+export function fetchJSON(url: string, opts?: { refresh?: boolean }): Promise<unknown> {
   _ensureLoaded()
-  const cached = _cache.get(url)
+  // When `refresh` is set, ignore the persisted ETag so the response is
+  // never served from cache. Used by the install wizard for R2 manifests
+  // that decide which standalone env release a user can pick — stale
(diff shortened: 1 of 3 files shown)


deepme987:
superseeded by https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/pull/612, closing

PR #612, opened 2026-05-28
fix(install): unstrand new installs from bundled ComfyUI on first launch

deepme987:
Stacks on #611 (the first commit on this branch is Maanil's commit from that PR). Together they fix the "Latest Stable" install landing on v0.20.1 instead of v0.22.3 that we hit while QA-ing the prerelease.

## Summary

A fresh standalone install via "Latest Stable" silently stranded users on the bundled ComfyUI version (currently v0.20.1) while showing "Already up to date" in the post-install update step. Repro confirmed on a clean state of this Windows machine — bug reproduces deterministically when no git backend is available at app startup.

## Root cause (trace)

1. App startup runs the git-backend cascade in `src/main/lib/ipc/index.ts:111–160`:
   - `tryConfigureBootstrapPygit2()` — needs `resources/bootstrap-python/<platform>/python.exe`. The 0.6.4 installer doesn't ship one (the startup log says so verbatim: `Bootstrap pygit2 not available — bootstrap-python/<platform>/ is missing. Run "pnpm run bootstrap"`).
   - Falls back to per-install pygit2 via `installations.list()` — empty on a fresh machine.
   - Falls back to system `git` — many Windows users don't have it on PATH.
   - Final branch logs `No git backend available (bootstrap pygit2, standalone pygit2, and system git all missing)`.
2. Line 167 then calls `await getLatestStableTag()` to pre-warm the install-wizard label. With no backend, `lsRemoteLatestTag` resolves `undefined` (it doesn't throw — `execFile('git', ...)` ENOENT goes to the success path). `getLatestStableTag` caches `{ tag: null, expiresAt: now + SUCCESS_TTL_MS (10 min) }`.
3. User clicks "New install → Latest Stable". `tryConfigurePygit2Fallback` in `install.ts:118–120` configures pygit2 against the just-extracted standalone-env python, so the lookup *would now work* — but `fetchLatestRelease('stable')` at line 149 reads the still-valid poisoned `null` from the in-memory cache. `latestTag = undefined` → `!latestTag` branch → "Already up to date" → install finalizes on v0.20.1.

Captured live cascade from the test run:

```
[ipc] Bootstrap pygit2 not available — bootstrap-python/<platform>/ is missing. …
[ipc] No git backend available (bootstrap pygit2, standalone pygit2, and system git all missing)
```

## Fix (layered)

**Root cause** — `comfyui-releases.ts`: cache a `null` tag with `FAILURE_TTL_MS` (30 s), not `SUCCESS_TTL_MS` (10 min). A null result means the call resolved without throwing but produced no value, which in practice only happens when no git backend was available; treating it as a brief failure lets the next caller refetch (post-install runs minutes later, by which time pygit2 is configured against the extracted env).

**Defense-in-depth** — `install.ts`: pass `{ refresh: true }` from the post-install update path so any future caller that warms the cache before post-install can't strand an install. Adds `opts?: { refresh?: boolean }` to `fetchLatestRelease` (passthrough to `getLatestStableTag`, which already supports it).

**Honesty + structural comparison** — from #611, retained on this branch: the post-install update now distinguishes "could not verify latest version" (network/git unavailable) from "already up to date" (structurally on the latest tag, via `current.baseTag === latestTag && current.commitsAhead === 0` instead of formatted-string equality).

**R2 manifest refresh** — from #611, retained: `fetchJSON({ refresh: true })` for the install-wizard's `latest.json` / `releases.json` reads so a freshly-shipped vendor build can't be hidden by a stale ETag.

## Follow-up (not in this PR)

The 0.6.4 installer is missing `resources/bootstrap-python/<platform>/python.exe`. This is a release-engineering issue — `pnpm run bootstrap` (or `bootstrap:fetch`) needs to run before the packaging step, and the directory needs to be included in the asar-unpacked output. With it present, the startup cascade succeeds at step 1 and this whole class of bug becomes much harder to hit. Worth a separate ticket.

## Test plan

- [x] Repro on a clean Windows state: clear `installations.json` / `release-cache.json` / `fetch-cache.json`, strip `git` from PATH, launch app, click "New install → Latest Stable". Without this fix the update step shows "Already up to date" and ComfyUI is v0.20.1; with this fix it updates to v0.22.3.
- [ ] Run with a real git backend (system git on PATH) and confirm Latest Stable still installs on latest and the message is "Up to date" once on latest, not "Skipped".
- [ ] Run with the network down during post-install (interface offline) and confirm the message is "Skipped — could not verify latest version" rather than a false "Already up to date".
- [x] Vitest: two new tests pin the failure-TTL behavior and the `refresh` plumbing.

### DIFF
--- a/src/main/lib/comfyui-releases.test.ts
+++ b/src/main/lib/comfyui-releases.test.ts
@@ -165,6 +165,39 @@ describe('fetchLatestRelease', () => {
       expect(result!.tag_name).toBe('v1.19.5')
       expect(mockedLsRemoteLatestTag).toHaveBeenCalledTimes(1)
     })
+
+    it('a null tag uses the short failure TTL, not the long success TTL', async () => {
+      // A null result (e.g. no git backend configured) used to be cached
+      // for SUCCESS_TTL_MS = 10 min, indistinguishable from a confirmed
+      // empty repo. The post-install update step then read that poisoned
+      // null on every subsequent call within 10 minutes and falsely showed
+      // "Already up to date", stranding new installs on the bundled
+      // ComfyUI version. Pin behavior: a null result must expire on the
+      // failure cadence so the next caller refetches.
+      vi.useFakeTimers()
+      try {
+        mockedLsRemoteLatestTag.mockResolvedValueOnce(undefined).mockResolvedValueOnce('v1.19.5')
+        const a = await getLatestStableTag()
+        // Advance past the failure TTL (30 s) but well within the success
+        // TTL (10 min) so we can tell the two apart.
+        vi.setSystemTime(Date.now() + 60_000)
+        const b = await getLatestStableTag()
+        expect(a).toBeNull()
+        expect(b).toBe('v1.19.5')
+        expect(mockedLsRemoteLatestTag).toHaveBeenCalledTimes(2)
+      } finally {
+        vi.useRealTimers()
+      }
+    })
+
+    it('fetchLatestRelease forwards refresh through to the tag lookup', async () => {
+      mockedLsRemoteLatestTag.mockResolvedValueOnce('v1.19.5').mockResolvedValueOnce('v1.19.6')
+      const a = await fetchLatestRelease('stable')
+      const b = await fetchLatestRelease('stable', { refresh: true })
+      expect(a!.tag_name).toBe('v1.19.5')
+      expect(b!.tag_name).toBe('v1.19.6')
+      expect(mockedLsRemoteLatestTag).toHaveBeenCalledTimes(2)
+    })
   })
 
   describe('mirror setting', () => {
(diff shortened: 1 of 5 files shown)


Kosinkadink:
Ah, if pygit2 is missing in the release, then that's the bug! And explains why cannot be reproduced in dev if "pnpm run init" is ran as expected before running (which creates the env).

The environment with pygit2 should always be shipped, should follow this up immediately with that fix. We never want to depend on system git for ANY operation in desktop; the at some point we should probably kill those paths so any issues with the built in env can't be missed.

Kosinkadink:
https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/issues/610

Kosinkadink (approved):

PR #665, opened 2026-05-28
fix(window): native close (X) shows its confirm on Windows

deepme987:
> **Status: parked / unconfirmed.** This was seen **once** on Windows (native ✕ on a running instance did nothing) but could **not be reproduced again** — possibly a stale build at the time. Parking this PR rather than merging. **If the symptom shows up again, this PR is the fix** — verify against plain `main` on Windows first to confirm the bug is real, then merge.

## Summary

Follow-up to #661. On **Windows**, clicking the native OS close (✕) on a running instance did **nothing** — the "Close window? This stops the running ComfyUI instance" confirm never appeared.

Root cause: the confirm was `await`ed from inside the preventDefault'd `'close'` event handler. On Windows the window sits in a "closing" limbo until that handler returns, so a child-view modal shown from within the continuation never surfaces. The hamburger **Close Window** works because it's invoked from an IPC handler, fully decoupled from any close event.

Fix: when the close needs the no-overlay confirm, hand off to the exact menu handler (`confirmAndCloseHostWindow`) on a fresh tick via `setImmediate` and return from the `'close'` handler immediately — so the modal is shown outside the close event, the same way the working menu does. The Tier 2/3 in-flight-operation cancel-prompt still runs inline (it's a renderer modal and was never affected).

Behavior is unchanged on macOS; this only changes *when/where* the confirm modal is invoked.

## Test plan

- [x] `pnpm typecheck` (node/web/e2e/integration)
- [x] `pnpm lint`
- [x] `pnpm test` (1349 unit tests)
- [x] macOS: native ✕ on an instance still confirms → closes / returns to dashboard; menu Close Window + Quit unchanged
- [ ] **Windows**: native ✕ on a running instance now shows the confirm → closes (stops instance) / returns to dashboard on the last window; never quits except the lone dashboard
- [ ] Windows: ✕ mid-install/update still shows the operation cancel-prompt

### DIFF
--- a/src/main/host/createHostWindow.ts
+++ b/src/main/host/createHostWindow.ts
@@ -79,13 +79,13 @@ export interface HostWindowFactories {
   consultPanelRendererClose: (
     panelView: WebContentsView | null | undefined,
   ) => Promise<'cleared' | 'aborted' | 'defer'>
-  /** Shell-level "Close Window" confirm, owned by main so a hidden panel
-   *  renderer can't swallow it. Shared with the Close Window menu entry. */
-  confirmCloseInstanceWindow: (
-    window: BrowserWindow,
-    isLastWindow: boolean,
-    theme: { bg: string; text: string },
-  ) => Promise<boolean>
+  /** The instance menu's "Close Window" handler. The OS ✕ hands off to it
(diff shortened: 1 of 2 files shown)


deepme987:
couldn't replicate, marking closed

PR #667, opened 2026-05-28
fix(snapshots): surface the pip failure reason on a failed restore (#633)

### LINKED ISSUE #633: Restore failed -- error not shown
Image

deepme987:
## Summary

Closes #633. A failed snapshot restore surfaced only a **generic** error — "Failed to install `<spec>`" / "Pip restore failed — reverted to pre-restore state" — with no explanation of *why*. The real pip/uv error (version conflict, no matching distribution, etc.) was streamed to the terminal log but never made it into the message shown in the progress modal's error row.

This captures each failing single install/uninstall command's output and extracts the meaningful error line(s) (`extractPipFailureReason`), appending them to `result.errors`. That detail already propagates through `actions.ts` → the operation's `error` → the progress modal, so the user now sees the actual cause.

- Prefers error-looking lines (`error`, `no matching distribution`, `conflict`, …); falls back to the last non-empty line(s); capped at ~300 chars.
- The streamed live log is unchanged — output is still forwarded as before.

## Notes

The original failure isn't currently reproducible locally, so this is verified by unit tests (a simulated failing `uv pip install` that emits a real error line) rather than a live repro.

## Test plan

- [x] `pnpm typecheck` (node/web/e2e/integration)
- [x] `pnpm lint`
- [x] `pnpm test` — 1354 unit tests, incl. new `restorePipPackages` reason-capture case + `extractPipFailureReason` unit tests
- [ ] If a real restore failure can be reproduced: confirm the progress-modal error row shows the underlying pip reason, not just the generic headline

### DIFF
--- a/src/main/lib/snapshots.test.ts
+++ b/src/main/lib/snapshots.test.ts
@@ -23,7 +23,7 @@ vi.mock('./telemetry', () => ({
   capture: vi.fn(),
 }))
 
-import { buildExportEnvelope, validateExportEnvelope, importSnapshots, diffSnapshots, listSnapshots, restoreComfyUIVersion, buildPostRestoreState, restorePipPackages, formatSnapshotVersion } from './snapshots'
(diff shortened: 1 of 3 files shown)


deepme987:
doesn't address real concern - a snapshot failure could be much more than just pip / uv failure

PR #668, opened 2026-05-28
feat(picker): copy-update via picker — product code only (Desktop 2.0)

Kosinkadink:
## Summary

Adds the picker-driven **Copy-Update** path: from the title-bar instance picker, the user can clone an existing install and run the updater on the copy in a single flow. The cancel branch surfaces as `'Cancelled.'` in the picker's progress card, and Return-to-Dashboard falls back to a system modal when the install's `panelView` has been destroyed.

Slim scope: **only the product code**. The picker-driven `@real` lifecycle test (test 13) and the `@ci`/`@real` tier-rename infrastructure it depends on are *not* included here; they will land in their own PR.

## Commits (4)

| Commit | Change |
|---|---|
| Return-to-Dashboard: system-modal fallback when panelView is destroyed | `src/main/host/detach.ts` |
| fix(ipc): route session-level actions through dispatcher from picker | new `src/main/lib/ipc/sessionActions/index.ts` + `src/main/index.ts` picker background-op path. Fixes *"Action not yet implemented"* when picker drives `copy` / `copy-update` / `delete`. |
| test(ids): expose basePrompt* test IDs for picker-driven prompts | `src/shared/testIds.ts`, `src/renderer/src/components/ui/BasePrompt.vue` |
| fix(copy-update): retry transient Windows delete errors; map cancellation to 'Cancelled.' | `src/main/lib/delete.ts` (retry `rmdir` / `unlink`), `src/main/lib/ipc/sessionActions/copy.ts` (cancel branches return `{ ok: false, cancelled: true }`), `src/main/index.ts` (picker background-op maps `cancelled` → `'Cancelled.'`), `src/renderer/src/views/ProgressModal.vue` (consumes the `'Cancelled.'` string) |

## Validation

- `pnpm run typecheck` — ✅
- `pnpm run lint` — ✅
- `pnpm run build` — ✅
- `pnpm run test` — ✅ 1354 / 1354

The `@real` picker copy-update lifecycle test that exercises this end-to-end will follow with the tier-rename PR.

## Known issues (deferred — will be tackled separately)

1. **"Updating" status shown for non-update operations.** The picker's progress overlay (`ComfyUISettingsContent.vue` `opTitleLabel`) hardcodes `"Updating…"` for every actionId except an `update-comfyui` downgrade. Affects `copy`, `copy-update`, `delete`, `release-update`, `snapshot-restore`, etc. — they all read as "Updating…" while in flight and "Update complete" on success.
2. **Copy-update cancellation UX** not manually verified end-to-end (banner reads "Cancelled" rather than "Update successful").


Kosinkadink (commented):
## Code review — feature/picker-copy-update (6 commits)

Reviewed the branch against `origin/main`. Three findings beyond the two already filed (#671 post-install cancel rollback, #672 in-progress op tab capture).

### 🐛 High — `titlePopup.ts` deletes the AbortController before the handler's catch can read it

`src/main/popups/titlePopup.ts` (≈L2507):

```ts
abort.abort()
_operationAborts.delete(installationId)
```

`pickerRunBackgroundOp`'s outer catch (`src/main/index.ts` ≈L1248) maps a thrown abort to the `'Cancelled.'` string by looking the controller back up:

```ts
const abort = _operationAborts.get(installationId)
result = { ..., error: abort?.signal.aborted ? 'Cancelled.' : ((err as Error).message ?? 'Failed.') }
```

For handlers that don't internally catch the abort (e.g. `handleDelete` → `deleteDir` throws `'Delete cancelled'`), the race is:

1. User clicks Cancel → `titlePopup` calls `abort()` and immediately `_operationAborts.delete(installationId)`.
2. The handler's in-flight `fs.readdir`/`fs.rmdir` callback then throws.
3. `pickerRunBackgroundOp` catches it, does `_operationAborts.get(installationId)` → `undefined`.
4. `abort?.signal.aborted` is `false` → result becomes `error: 'Delete cancelled'` instead of `'Cancelled.'`.

The renderer (`PickerInlineProgress.vue` L29) gates `isCancelled` on the literal string `'Cancelled.'`, so the user sees a red "Something went wrong / Delete cancelled" banner instead of the cancelled state.

The copy / copy-update / release-update handlers happen to dodge this because they reference the *local* `abort` variable in their own catch and return `cancelled: true` before bubbling — but the safety-net path is broken for everything else.

**Fix:** drop `_operationAborts.delete(installationId)` from the cancel IPC. `AbortController.abort()` is idempotent against a second click, and the handler's own `finally`/catch path already clears the map entry. Alternatively, check the abort *flag* (e.g. capture `abort.signal.aborted` into a local at cancel time and read off `abort` directly in the outer catch) instead of round-tripping through the map.

### 🔧 Medium — duplicated lifecycle boilerplate across `handleCopy` / `handleCopyUpdate` / `handleReleaseUpdate`

`src/main/lib/ipc/sessionActions/copy.ts` — all three handlers repeat the same shape:

- name/path validation
- `_operationAborts.has(installationId)` guard
- `new AbortController()` + `_operationAborts.set(installationId, abort)`
- `try { … } catch { _operationAborts.delete(...); if (abort.signal.aborted) return { ok:false, cancelled:true, message:'Cancelled' }; return { ok:false, message: err.message } }`

The `'Cancelled'` mapping in particular *must* stay in sync across all three sites (and with `src/main/index.ts`'s `'Cancelled.'` writer and `PickerInlineProgress.vue`'s match string — note the trailing-period mismatch, see below).

**Fix:** extract a `withAbortableSessionAction(ctx, fn)` helper that owns the precondition checks, controller registration, map cleanup, and abort → `cancelled: true` mapping. Inner handlers just receive a live `signal` and return their success result.

### ⚠️ Low — `'Cancelled'` vs `'Cancelled.'` string drift

- `sessionActions/copy.ts` returns `message: 'Cancelled'` (no period) inside `{ cancelled: true, … }`.
- `src/main/index.ts` writes `error: 'Cancelled.'` (with period) into `PickerOperationStatus`.
- `PickerInlineProgress.vue` matches `props.operation.error === 'Cancelled.'`.

Today this works because `pickerRunBackgroundOp` overwrites `actionResult.message` with the hardcoded `'Cancelled.'` literal when `actionResult.cancelled === true`. But the handler's `message: 'Cancelled'` is dead bytes that look authoritative — a future caller (e.g. the `registerSessionHandlers` `run-action` IPC path that now also routes through `dispatchSessionAction`) will surface the unmapped no-period form to the modal renderer, which gates on a different string.

**Fix:** centralize the cancelled-string constant (`MSG_CANCELLED = 'Cancelled.'`) in `shared.ts`, use it on both the handler return *and* the outer-catch mapper, and have the renderer match against the same constant via the preload bridge.

### 🔧 Low — `dispatchSessionAction(actionId: string)` is untyped

`src/main/lib/ipc/sessionActions/index.ts` — the new central router accepts `actionId: string`, so the switch has no exhaustiveness check. Adding a new session-level action won't produce a compile error if the case is missed; it silently falls through to `handleDelegateToSource` and gets pushed at the source's `handleAction`, which surfaces as `'Unknown action: …'` only at runtime.

**Fix:** introduce `type SessionActionId = 'remove' | 'open-folder' | 'delete' | 'copy' | 'copy-update' | 'release-update' | 'migrate-to-standalone' | 'launch'`, narrow the parameter, and split the catch-all delegate into an explicit `delegateActionId: string` second overload (or callsite) so the type system distinguishes session-level vs source-level dispatch.

### ✅ Nits / non-blocking

- `progressStatusLabel.ts` exhaustiveness is solid and the test matrix (`progressStatusLabel.test.ts`) locks the mapping, including the `op.title → 'Working…'` fallback.
- `delete.ts` retry set (`ENOTEMPTY` / `EBUSY` / `EPERM` / `EACCES`) plus capped exponential backoff (50→800 ms, 5 attempts) is the right shape for the Windows handle-release window. Worth noting the retry loop does **not** check `signal.aborted` between attempts — a single hot file could keep retrying up to ~1.55 s after the user cancels. Cheap to fix: short-circuit `retryFsOp` if `signal?.aborted`.
- `showOpOverlay` gate change reads correctly: snapshot-restore on the snapshots tab is the only carveout, and the new test cases (`overlay routing` describe block) lock both the carveout and the symmetric "show on non-snapshots tab" case.
- Static-import fix in `titlePopup.ts` is the right call — no other `import(…)` dynamic imports remain in that file (`Select-String` came back empty).

---

Not re-reporting: #671 (copy-update / release-update post-install cancel rollback), #672 (in-progress op tab capture). The High finding above is the same shape of bug — a renderer state string going wrong because of a swallowed/lost signal — so a unified `withAbortableSessionAction` helper plus a shared `MSG_CANCELLED` constant would address that family of issues at once.

PR #687, opened 2026-05-28
fix(titlebar): keep title padding at narrow widths

deepme987:
### Issue

From internal desktop testing: "At certain widths, padding around title is non-existent. Should have some padding." When the window narrows, the left cluster (`min-width` mirroring the trailing cluster + the 128px Win control reservation) and the trailing cluster squeeze the centre `minmax(0, 1fr)` grid track toward zero. The install/title pill then ends up flush against the adjacent chrome with no horizontal breathing room.

### Fix

Reserve a symmetric inline-padding floor on `.title-center`:

```css
.title-center {
  box-sizing: border-box;
  padding-inline: 8px;
}
```

This guarantees a minimum horizontal gap on both sides of the title pill at every width. `box-sizing: border-box` keeps the padding inside the declared `max-width: 100%` so the track never overflows.

### Preserves the centered pill

The padding is symmetric, so the `justify-self: center` anchor and the ResizeObserver-driven true-window-centre mirror (PR #647) are untouched — only the slack the pill can grow into shrinks equally on each side. The centred-pill layout is preserved.

### Testing

- Added a regression assertion in `TitleBarApp.test.ts` (the layout itself isn't computable under happy-dom, so it asserts the scoped-CSS contract: symmetric `padding-inline`, `box-sizing: border-box`, and that the `justify-self: center` anchor remains).
- `pnpm test` scoped to `comfyTitleBar`: 50/50 pass.
- `pnpm typecheck` and `pnpm lint`: clean.

Visual verification needed at narrow widths (320 / 768 / 1024 / 1440px) on both Windows and macOS to confirm the gap and that the pill stays at true window centre.

### DIFF
--- a/src/renderer/src/comfyTitleBar/TitleBarApp.test.ts
+++ b/src/renderer/src/comfyTitleBar/TitleBarApp.test.ts
@@ -1,3 +1,5 @@
+import { readFileSync } from 'node:fs'
+import { resolve } from 'node:path'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { mount, flushPromises } from '@vue/test-utils'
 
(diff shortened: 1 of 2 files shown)


deepme987:
Closing: on review, the reported title-padding issue appears to be about the Cloud UI's right-side padding, not the Desktop 2.0 title bar. The desk-2 title-bar padding here isn't the actual report. Leaving #696 open to reclassify.

PR #734, opened 2026-05-28
fix(installer): align variant-card ComfyUI version with selected release

### LINKED ISSUE #708: [P&P] Version numbers are not aligned
Reported in #temp-desktop-bug-dump (Prompt & Play QA) by Alexis Rolland.

Version numbers are not aligned here (see screenshot).

Slack: https://comfy-organization.slack.com/archives/C0B6R9Y4YDT/p1780007816022589

deepme987:
Closes #708

## Problem
On the New Install wizard (Standalone tab), the **Release** dropdown's "Latest Stable (Recommended)" showed one ComfyUI version (e.g. `v0.22.3`) while the hardware **variant cards** (NVIDIA / Intel Arc / CPU) showed a different one (e.g. `ComfyUI 0.20.1`). They are sourced from two independent places, so they could disagree:

- **Dropdown description** = `getLatestStableTag()` — the latest upstream `Comfy-Org/ComfyUI` git tag (what the post-install "update to stable" step resolves to).
- **Variant cards** = `release.comfyui_version` from the newest R2 standalone bundle — an older ComfyUI baked into the prebuilt env. With `autoUpdateComfyUI: true` set for the `latest` selection, the install then auto-updates past it.

So a user picking "Latest Stable" ends up on the upstream stable version, but the card advertised the bundle's older version.

## Fix
Thread the resolved latest-stable tag through the `latest` release option's `data`, and have the variant cards display that version when "Latest Stable" is selected — matching the dropdown and the version the user actually ends up with after auto-update.

- Pinned (non-latest) releases still show the version baked into that env (correct, by design).
- Falls back to the bundled version when the upstream tag can't be resolved (offline / git unavailable).
- Display-only: the frozen install-time fingerprint (`originalBuild`, `originalTorchVersion`) still tracks the actual downloaded bundle.

Isolated to `src/main/sources/standalone/index.ts`; does not touch `InstallWizardModal.vue` (no conflict with #700).

## Validation
- `pnpm typecheck` — pass (node/web/e2e/integration)
- `pnpm lint` — pass
- `pnpm vitest run src/main/sources/standalone/index.test.ts` — 14 pass (3 new covering latest/fallback/pinned version display)

### DIFF
--- a/src/main/sources/standalone/index.test.ts
+++ b/src/main/sources/standalone/index.test.ts
@@ -176,3 +176,63 @@ describe('standalone.getFieldOptions release', () => {
     expect(latestEntry.description).toBeUndefined()
   })
 })
+
+// --- getFieldOptions('variant') — version-display consistency (issue #708) ---
+
+describe('standalone.getFieldOptions variant version display', () => {
+  // The newest R2 standalone bundle ships an OLDER ComfyUI (0.20.1) than the
+  // upstream stable tag the wizard auto-updates to (v0.22.3). Set up that gap.
+  function setupVersionGap() {
+    const { latest, vendorReleases, vendorId } = makeR2Releases(
+      ['v0.20.1-env1'],
+      { comfyuiVersion: '0.20.1' },
+    )
+    mockedFetchJSON.mockImplementation((url: string) => {
+      if (url.includes('latest.json')) return Promise.resolve(latest)
+      return Promise.resolve(vendorReleases[vendorId]!)
+    })
+    return { vendorId }
+  }
+
+  async function getReleaseOption(value: string) {
+    const releaseOptions = await standalone.getFieldOptions!(
+      'release',
+      {},
+      { includeLatestStable: true },
+    )
+    return releaseOptions.find((o) => o.value === value)!
+  }
+
+  it('variant card shows the upstream stable version (not the bundled one) when "Latest Stable" is selected', async () => {
+    const { vendorId } = setupVersionGap()
+    mockedGetLatestStableTag.mockResolvedValue('v0.22.3')
+    const release = await getReleaseOption('latest')
+
+    const variants = await standalone.getFieldOptions!('variant', { release }, {})
+    const card = variants.find((o) => o.value === vendorId)!
+    // The card and the dropdown must agree: both surface 0.22.3.
+    expect(card.description).toContain('ComfyUI 0.22.3')
+    expect(card.description).not.toContain('ComfyUI 0.20.1')
+  })
+
+  it('variant card falls back to the bundled version when the upstream tag is unresolved', async () => {
+    const { vendorId } = setupVersionGap()
+    mockedGetLatestStableTag.mockResolvedValue(null)
+    const release = await getReleaseOption('latest')
+
+    const variants = await standalone.getFieldOptions!('variant', { release }, {})
+    const card = variants.find((o) => o.value === vendorId)!
+    expect(card.description).toContain('ComfyUI 0.20.1')
+  })
+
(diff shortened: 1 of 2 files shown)


deepme987:
Closing — mis-scoped. The report's screenshot is about custom-node *download* version mismatch, which lives in ComfyUI_frontend, not the desktop installer's variant cards. #708 should be re-routed to the frontend repo.

Kosinkadink (commented):
## Code review - verified ?

Re-reading the diff after merging latest main (no conflicts), and tracing the data path end-to-end, this PR does exactly what its title says.

### Behavior trace (confirms it does what it should)

The wizard surfaces the ComfyUI version in two places that previously came from independent sources:

| Surface | Source (before) | Source (after) |
|---|---|---|
| Release dropdown - "Latest Stable" description |  getLatestStableTag() (upstream tag, e.g.  v0.22.3) | unchanged |
| Variant card description | 
elease.comfyui_version (R2 bundle, e.g. .20.1) | latestStableTag when isLatest, else bundled |

The dropdown's "Latest Stable" sets utoUpdateComfyUI: true in the build args ([standalone/index.ts:114](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/deepme987/desktop2/fix-install-version-mismatch/src/main/sources/standalone/index.ts#L114)), and the post-install step at [install.ts:143-175](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/deepme987/desktop2/fix-install-version-mismatch/src/main/sources/standalone/install.ts#L143-L175) then runs 
unComfyUIUpdate({ channel: 'stable' }), which resolves to the same latestStableTag the dropdown advertises. So the card now correctly previews the version the user will actually end up on after install.

### Correctness checklist

- ? **Latest Stable selection** - card shows latestStableTag (with leading  stripped to match the bare-version format the card has always used).
- ? **Pinned (non-latest) release** - isLatest is alse, so the card falls through to 
elease.comfyui_version (the version actually baked into that R2 bundle). Correct by design.
- ? **Offline / getLatestStableTag() returns 
ull** - 
eleaseData.latestStableTag is falsy, so the card falls back to 
elease.comfyui_version. No regression vs. pre-PR behavior.
- ? **Install-time fingerprint untouched** - only the display string changes. The download URL, manifest.comfyui_ref, and the 
2Release object passed back through data still reference 
elease.comfyui_version, so originalBuild / originalTorchVersion still track the actual downloaded bundle.
- ? **Scope** - confined to standalone/index.ts; doesn't touch InstallWizardModal.vue, so no conflict with #700.

### Tests

The 3 new tests in standalone/index.test.ts cover the three paths above:
1. "Latest Stable" selected, upstream tag resolved ? card shows .22.3, not .20.1.
2. "Latest Stable" selected, upstream tag unresolved ? card falls back to .20.1.
3. Pinned release selected ? card shows bundled .20.1.

### Local validation (after merging main into the branch)

- pnpm typecheck - pass (node/web/e2e/integration)
- pnpm lint - pass
- pnpm build - pass (only pre-existing static/dynamic import warnings)
- pnpm test - **1514/1514 pass** (116 files), including the 3 new ones

### Nits (non-blocking)

- The data-type cast s { ...; latestStableTag?: string | null } at [line 245](https://github.com/Comfy-Org/ComfyUI-Desktop-2.0-Beta/blob/deepme987/desktop2/fix-install-version-mismatch/src/main/sources/standalone/index.ts#L245) widens to four occurrences of the same release-data shape; if this grows further, a single named interface would be nice. Not worth a separate commit here.

LGTM. Re-opening - the close looks accidental.

PR #737, opened 2026-05-28
fix(picker): reset source filter to All when the active chip disappears

### LINKED ISSUE #723: [P&P] Dashboard filter chips: 'All' hides local installs after toggling Cloud then back
Reported in #temp-desktop-bug-dump (Prompt & Play QA) by Robert Paige.

Dashboard filter chips: clicking "All" won't show local installs after you toggle to "Cloud" and back to "All".

Slack: https://comfy-organization.slack.com/archives/C0B6R9Y4YDT/p1780009703417739

deepme987:
## Summary

On the dashboard instance picker, a source-filter chip (Local / Cloud / Remote) only renders while at least one install backs that category. When the active category emptied — e.g. the cloud or remote install dropped out of a snapshot refresh — its chip disappeared, but `activeFilter` stayed pinned to the now-hidden key. The list was then stuck showing nothing, no chip was highlighted, and the surviving local installs never reappeared. This is the "'All' hides local installs after toggling Cloud then back" symptom from the report.

## Fix

Watch the visible chip set in `InstancePickerView.vue`; when the active filter key is no longer among the rendered chips, fall back to `'all'`. This mirrors the existing guard already used in `DownloadsModal.vue`, so the two filter surfaces behave consistently.

## Test plan

- [x] Added a focused regression test: activate Remote, remove the remote install, assert the filter resets to All (highlighted) and the surviving local install reappears.
- [x] `pnpm typecheck` (node/web/e2e/integration) passes
- [x] `pnpm lint` passes
- [x] `InstancePickerView` / `useInstallList` / `ChooserView` unit suites pass (58 tests)

Closes #723

### DIFF
--- a/src/renderer/src/comfyTitlePopup/InstancePickerView.test.ts
+++ b/src/renderer/src/comfyTitlePopup/InstancePickerView.test.ts
@@ -352,6 +352,43 @@ describe('comfyTitlePopup/InstancePickerView', () => {
       expect(rows[0]!.text()).toContain('RemoteThing')
     })
 
+    // Regression (#723): a category chip only renders while an install
+    // backs it, so when the active category empties (e.g. the cloud /
+    // remote install drops out of a snapshot refresh) its chip vanishes.
+    // `activeFilter` must fall back to 'all' instead of staying pinned to
+    // the now-hidden key — otherwise the list is stuck showing nothing and
+    // the surviving installs (locals) never reappear.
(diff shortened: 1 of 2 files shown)


deepme987:
Closing — not reproducible on local dev or the build. Most consistent with a transient snapshot drop-out (active chip vanished mid-refresh). The fall-back-to-All guard is sound defensive hardening, but holding rather than merging an unreproducible fix; will revisit if it recurs.

PR #741, opened 2026-05-28
fix(install): defer baseline snapshot until after first-launch auto-update

### LINKED ISSUE #707: [P&P] Fresh install shows a phantom snapshot upgrade (v0.20.0 to v0.22.3)
Reported in #temp-desktop-bug-dump (Prompt & Play QA) by Robin Huang.

On a fresh install, a snapshot still shows — implying an upgrade v0.20.0 -> v0.22.3 that never happened. Feels wrong for a clean install.

Slack: https://comfy-organization.slack.com/archives/C0B6R9Y4YDT/p1780007877286799

deepme987:
## Summary
On a fresh "Latest Stable" install, the snapshot history showed a phantom upgrade (e.g. v0.20.0 → v0.22.3) that the user never performed.

**Root cause:** `postInstall` saved a `boot` snapshot at the just-extracted *bundled* ComfyUI version, *then* ran the first-launch auto-update to the latest release — which writes its own `post-update` snapshot. The two snapshots left a diff in the history that reads as an upgrade, even though it was all one uninterrupted clean install. (Surfaced after #612 made that auto-update actually fire.)

**Fix:** Skip the `boot` snapshot when an auto-update will run; the `post-update` snapshot becomes the single baseline that reflects what actually landed. Fall back to a `boot` snapshot when no auto-update runs (pinned release / no git) or when the update is skipped/fails, so there's always a "Current" baseline.

## Test plan
- [x] New `install.test.ts` covers: successful auto-update (no boot snapshot), pinned release (boot snapshot), latest-unverifiable skip (fallback boot), and update failure (fallback boot)
- [x] `pnpm run typecheck:node` + `typecheck:integration` pass
- [x] `eslint` clean (pre-commit hook)
- [x] Full standalone + snapshot suites pass (119 tests)

Closes #707

### DIFF
--- a/src/main/sources/standalone/install.test.ts
+++ b/src/main/sources/standalone/install.test.ts
@@ -1,0 +1,193 @@
+// @vitest-environment node
+/**
+ * Unit tests for postInstall's snapshot-trigger decision.
+ *
+ * Regression guard for #707: a fresh "Latest Stable" install auto-updates the
+ * just-extracted bundled ComfyUI to the newest release, which writes its own
+ * `post-update` snapshot. Capturing a `boot` snapshot *before* that update
+ * recorded the bundled version (e.g. v0.20.x), leaving the snapshot history
+ * showing a phantom "upgrade" (bundled → latest) the user never performed.
+ *
+ * postInstall should therefore:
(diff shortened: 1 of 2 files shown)


deepme987:
closing - expected behavior.

Issue #702, opened 2026-05-28
[P&P] Add tooltips/explanations for core concepts (Instances, Snapshots)


deepme987:
Reported in #temp-desktop-bug-dump (Prompt & Play QA) by Robin Huang.

Some primitive concepts could use tooltips or a bit more explanation — e.g. Instances, Snapshots, etc.

Slack: https://comfy-organization.slack.com/archives/C0B6R9Y4YDT/p1780008237870569

Filed via Prompt & Play triage.

PR #742, opened 2026-05-28
fix(settings): trim redundant label-echo tooltips on settings tabs

### LINKED ISSUE #713: [P&P] Trim tooltips that add no info and obstruct content
Reported in #temp-desktop-bug-dump (Prompt & Play QA) by Alex Tov.

Nit: some tooltips aren't telling anything extra and get in the way of the content in the section. Trim tooltips that add no info.

Slack: https://comfy-organization.slack.com/archives/C0B6R9Y4YDT/p1780007733925789

deepme987:
## Summary

The settings tab strip (Update · Startup Args · Snapshots · Storage · About) wrapped every tab button in a tooltip whose text just echoed the tab's own always-visible label. Hovering e.g. the "Snapshots" tab popped a "Snapshots" tooltip — zero added info, and the bottom-anchored popover overlapped the content beneath the strip (reported in Prompt & Play QA).

## Fix

Gate each tab tooltip on whether its label is actually hidden, rather than removing the tooltip outright:

- The strip only hides labels in its narrow-container state (`@container settings-tabs (max-width: 520px)`), where inactive tabs collapse to icon-only. There the tooltip is the sole way to read the label, so it's genuinely useful and is kept.
- In the wide state (the reported case) every label is visible, so all tab tooltips are suppressed.
- The active tab always keeps its label, so its echo tooltip is suppressed in every state.

A `ResizeObserver` (mirroring the existing `TitleBarApp` pattern, with the same `typeof ResizeObserver !== 'undefined'` guard) tracks the collapse breakpoint in JS and drives the `Tooltip`'s `disabled` prop.

Scope is strictly the tab-bar echo tooltips; panel-body help (e.g. #702's "what is a Snapshot") is untouched.

## Files

- `src/renderer/src/components/settings/ComfyUISettingsContent.vue`
- `src/renderer/src/components/settings/ComfyUISettingsContent.test.ts`

## Validation

- `pnpm typecheck` (node/web/e2e/integration) — pass
- `pnpm lint` — pass
- `pnpm vitest run ComfyUISettingsContent.test.ts` — 19/19 pass (2 new cases lock the suppress-at-wide / keep-on-collapsed behaviour)

Closes #713

### DIFF
--- a/src/renderer/src/components/settings/ComfyUISettingsContent.test.ts
+++ b/src/renderer/src/components/settings/ComfyUISettingsContent.test.ts
@@ -335,6 +335,79 @@ describe('ComfyUISettingsContent', () => {
     })
   })
 
+  describe('tab tooltips (#713 — no redundant label echo)', () => {
+    type ResizeCb = (entries: ResizeObserverEntry[], obs: ResizeObserver) => void
+    interface RoHandle {
+      el: Element
+      fire(width: number): void
+    }
+    let roHandles: RoHandle[]
+    let originalRo: typeof globalThis.ResizeObserver | undefined
+
+    beforeEach(() => {
+      roHandles = []
+      class StubRo {
+        cb: ResizeCb
+        constructor(cb: ResizeCb) {
+          this.cb = cb
+        }
+        observe(el: Element): void {
+          roHandles.push({
+            el,
+            fire: (width: number) => {
+              this.cb(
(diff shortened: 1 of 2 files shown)


deepme987:
Superseded by #745 — the trim-redundant-tab-tooltips change was merged into #745, which now closes both #702 and #713 with the reconciled behavior (info tooltips always show; label-echo tooltips only when the tab label is collapsed).

PR #743, opened 2026-05-28
fix(host): recover comfy-view zoom-in (zoom no longer stuck after restart)

### LINKED ISSUE #698: [P&P] Cannot zoom back in with Ctrl/Cmd +/- (zoom stuck, persists across restart)
Reported in #temp-desktop-bug-dump (Prompt & Play QA) by Purz.

In Desktop you can only Zoom Out — you can't Zoom back In with Ctrl/Cmd +/-. Worse, it stays stuck that way even after restarting the app.

Slack: https://comfy-organization.slack.com/archives/C0B6R9Y4YDT/p1780006472553219

deepme987:
## Summary

Closes #698.

In Desktop 2.0 you could only zoom **out** with Ctrl/Cmd +/- on the comfy view — you couldn't zoom back **in**, and the stuck level survived an app restart.

Two root causes, both on the desktop side:

- **Persists across restart:** the comfy `WebContentsView` lives in a `persist:` session partition, and Electron persists the zoom level per-partition. Nothing reconciled it on load, so a zoomed-out level was restored on every launch.
- **Can't zoom back in:** the `before-input-event` handler incremented off a fresh `webContents.getZoomLevel()`, which drifts out of sync with the visually-applied persisted zoom right after a navigation (known Electron behavior). Zoom-out still visibly shrank, but zoom-in stalled.

## Fix

- Track the zoom level as the single source of truth in the attach closure; drive +/-/0 off it (clamped to a recoverable range so 1x is always reachable).
- Re-assert the tracked level on `dom-ready` so a persisted/stale level can't strand the view across reload or restart.
- Extract the pure level math to `src/main/host/zoom.ts` with unit tests.

## Test plan

- [x] `pnpm typecheck` (node/web/e2e/integration) passes
- [x] `pnpm lint` passes
- [x] `pnpm exec vitest run src/main/host/zoom.test.ts` — 5/5 pass (zoom-in recovers from a zoomed-out level, in/out are inverses, reset returns to 1x, clamping)
- [ ] Manual: zoom out with Ctrl/Cmd+-, then zoom back in with Ctrl/Cmd+= to 1x; restart app and confirm zoom is no longer stuck

### DIFF
--- a/src/main/host/attach.ts
+++ b/src/main/host/attach.ts
@@ -14,6 +14,7 @@ import {
   setLastFocusedInstallationId,
 } from './registry'
 import type { ComfyWindowEntry } from './registry'
+import { nextZoomLevel } from './zoom'
 
 const APP_VERSION = getAppVersion()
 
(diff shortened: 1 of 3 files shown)


deepme987:
closing - doesn't fix the issue. @MaanilVerma to takeover

PR #685, opened 2026-05-28
fix(launcher): preserve GPU vendor logo aspect ratio

deepme987:
## Issue
On the New Install wizard (GPU vendor selection step), the Apple MPS logo looked distorted/squished. The vendor logos render inside square icon boxes (96x96 in the card grid, 40x40 in the variant list), but the Apple MPS asset is 420x210 (2:1) while NVIDIA, AMD, and Intel are square. With `object-fit: cover`, the wide logo was scaled to fill the square and got cropped/zoomed, so it read as distorted next to the cleanly-fitting square logos.

## Fix
Switch the logo `` rules from `object-fit: cover` to `object-fit: contain` in both render paths:
- `.variant-card-icon img` (96x96 card grid) in `src/renderer/src/assets/main.css`
- `.brand-variant-row__icon img` (40x40 list row) in `src/renderer/src/components/BrandVariantList.vue`

`contain` preserves each logo's native aspect ratio and shows it in full (letterboxed within the box) rather than cropping. Square logos are unaffected.

## Validation
- `pnpm typecheck` — pass
- `pnpm lint` — pass

## Visual verification needed
A human should confirm the GPU vendor logos (especially Apple MPS) now display at the correct aspect ratio in the New Install wizard.

PR #754, opened 2026-05-29
fix(chooser): preserve Comfy wordmark aspect ratio on dashboard

### LINKED ISSUE #697: [P&P] Logo is squished
Reported in #temp-desktop-bug-dump (Prompt & Play QA) by Robert Paige.

The logo renders squished.

Slack: https://comfy-organization.slack.com/archives/C0B6R9Y4YDT/p1780006798568489

deepme987:
## Issue
Closes #697

The large yellow "Comfy" wordmark in the center of the dashboard / instance-chooser (above the "Search for and open an instance" bar, on the screen with the New Install / Comfy Cloud / Legacy Desktop cards) rendered horizontally squished — the wrong aspect ratio.

## Root cause
`ComfyWordmark.vue` was the only brand SVG that hardcoded `width="100" height="23"` presentation attributes. When the component is placed as a grid item in `ChooserView.vue` (`.chooser-wordmark` sets `width: clamp(120px, 8vw, 180px); height: auto`), Chromium could not cleanly resolve the intrinsic aspect ratio against the grid track and scaled the SVG non-uniformly. Its sibling brand SVGs (`ComfyCLogo`, `BrandProgressGlyph`) size purely from `viewBox` and render correctly, which is why this wordmark was the lone offender.

## Fix
- Remove the fixed `width`/`height` attributes from `ComfyWordmark.vue` (keep the `viewBox` so the 100:23 intrinsic aspect ratio is preserved).
- Add `preserveAspectRatio="xMidYMid meet"`, matching `BrandProgressGlyph`.

The wordmark now keeps its native aspect ratio in any layout context. No consumer CSS changed, so the existing ProgressModal / first-use install usage is unaffected.

## Scope note
This **replaces the closed #685**, which mistakenly fixed the GPU *vendor-card* logos in the New Install wizard. Those vendor logos are deliberately untouched here — this PR targets the **dashboard center wordmark** only.

## Validation
- `pnpm install` — ok
- `pnpm typecheck` — pass (node / web / e2e / integration)
- `pnpm lint` — pass

## Visual verification needed
A human should confirm the dashboard wordmark now displays at its correct aspect ratio.


deepme987:
Deferring for now — couldn't visually confirm the difference at this window size. The fix (ComfyWordmark aspect-ratio) is geometrically correct but unverified, so holding off rather than merging blind. #697 stays open to revisit later.

PR #795, opened 2026-06-03
fix(settings): always render the Chinese mirrors description (closes #779)

### LINKED ISSUE #779: Regression - the Use Chinese mirrors toggle does not explain what it does when its toggled
https://github.com/user-attachments/assets/695d33b9-a2c1-428e-be8b-66f155ad023f

Kosinkadink:
Fixes #779.

## What was wrong
The **Use Chinese Mirrors (Git & PyPI)** toggle in the Global Settings popup had no on-screen explanation at all. The two-layer regression along the new title-popup pipeline:

1. **Settings payload stripped the description.** `toDetailField()` in `src/main/popups/titlePopup.ts` didn't copy `f.description` from the main-side `SettingsField` into the loose-typed `DetailField` shape it ships to the renderer.
2. **Renderer didn't read it.** `SettingsSectionList.vue` (the shared field renderer used by the popup and the per-install settings) had no `v-if="field.description"` block, and `DetailField` didn't even declare a `description?` member.

The user had no way to know what flipping the toggle would do, in either state.

## Fix
- Add `description?: string` to `DetailField` in `src/types/ipc.ts`.
- Propagate it in `toDetailField()`.
- Render an inline `ShieldAlert` notice beneath the control in `SettingsSectionList.vue` for any field that carries one (styling matches the existing env-vars notice family).
- In `buildSettingsSections()`:
  - **Always** attach the explanation as the field's `tooltip`, so hovering the info-icon answers the question in **either** state (OFF or ON).
  - Keep `description` gated to the ON state. The inline copy ("Git repositories clone from `gitcode.com`.") reads as "here's what's now in use", so it only makes sense when the toggle is enabled. The OFF state is covered by the hover tooltip.

End result: tooltip on hover **always** explains the toggle; inline description appears below the control **once enabled** to confirm what's now active.

## Popup-payload dedupe fix (caught in self-review)
The popup-update broadcast paths (`broadcastGlobalSettingsUpdate()` / `broadcastInstancePickerUpdate()`) skip a send when the new payload JSON equals `entry.lastXxxBroadcastJson` - a per-popup cache of the most recently broadcast payload, used to suppress redundant re-renders.

That cache used to be populated **only** by live broadcasts, never by the initial payload shipped at popup-open time. Failure mode: popup opens showing payload **B** (cache still says **A** from a previous session), user changes a setting back to **A**, the live broadcast is deduped against the stale cache, renderer keeps showing **B**.

Fixed by seeding both caches inside `openTitlePopup()` whenever the initial config carries a payload, so the cache always reflects what the renderer is currently displaying.

(Note: "payload" here is the per-popup `GlobalSettingsSnapshot` / `InstancePickerSnapshot` object - unrelated to the user-facing snapshots feature.)

## Dead-code cleanup
The soft-deleted legacy settings chain mentioned in `SettingsModal.vue`'s own header comment was no longer reachable from any live code:

- `views/SettingsModal.vue`, `views/SettingsView.vue` (+ test), `views/DirectoriesView.vue`, `views/DownloadsView.vue` (+ test) - replaced by `GlobalSettingsView.vue` (in the title-popup) and `ManageInstallModal.vue`.
- `components/SettingsSections.vue`, `components/SettingField.vue` (+ test) - only consumed by the deleted chain.
- `components/AppUpdateAction.vue` - only imported by the deleted `SettingsView.vue`; `UpdatesSection.vue` already carries the equivalent UI.
- `composables/useGlobalSettings.ts` - orphaned after the legacy `SettingsView` deletion; was its own (incomplete) `toDetailField` mapper, so removing it kills the drift risk at the source.

Also dropped the now-orphan `vi.mock` calls + absence-assertions in `PanelApp.test.ts`, and refreshed stale `"legacy SettingsModal"` references in `DetailModal.vue`, `ManageInstallModal.vue`, `titlePopup.ts`, `comfyTitlePopupPreload.ts`, `ipc.ts`, and `UpdatesSection.vue` per the AGENTS.md "don't narrate history" rule.

Follow-up cleanup for the rest of the codebase is tracked in #796.

## Tests
Added `SettingsSectionList.test.ts` covering:
- description renders below the control when one is attached
- no `.settings-v2-field-description` node for fields without one
- description renders for non-boolean field types too
- `InfoTooltip` trigger appears when a field has a `tooltip`

## Verification
- `pnpm run typecheck` ?
- `pnpm run lint` ?
- `pnpm run test` - **1603 passed (+4 new)**
- `pnpm run build` ?

Net change: 21 files, **+163 / ?2442**.

Amp-Thread-ID: https://ampcode.com/threads/T-019e8b46-2619-779d-be9c-910757f3cb18
Co-authored-by: Amp <amp@ampcode.com>


Kosinkadink:
## Self-review (oracle-assisted)

Ran the oracle over the diff. It flagged two things worth addressing in this PR; both are now in `70e59e2`:

### 1. Snapshot-broadcast dedupe could swallow a real change
`broadcastGlobalSettingsUpdate()` / `broadcastInstancePickerUpdate()` skip a send when `JSON.stringify(snapshot) === entry.lastXxxBroadcastJson`. That cache used to be populated *only* by live broadcasts, never by the initial config we ship at popup-open time. Failure mode: popup opens with snapshot **B** (cache still says **A** from the previous session), user changes a setting back to **A**, the live broadcast is deduped against the stale cache, renderer keeps showing **B**. Fixed by seeding both caches inside `openTitlePopup()` whenever the initial config carries a snapshot.

### 2. Description rendering had no test
Added `SettingsSectionList.test.ts` with four cases:
- description present under a boolean control when value=false (the OFF state - the exact regression #779 reports)
- description present when value=true
- no `.settings-v2-field-description` node for fields without a description
- description renders for non-boolean (text) field types too

### Also addressed in 70e59e2
Deleted `src/renderer/src/composables/useGlobalSettings.ts` - the oracle pointed out it has its own (also incomplete) `toDetailField` mapper. After the legacy `SettingsView` deletion in this PR's first commit it has no live importers, so removing it kills the drift risk at the source rather than patching the orphan.

### Intentionally not in this PR
- **Readonly grid-column CSS guard for descriptions.** Defensive code for a state we never hit (no readonly status field carries a description). Add when something needs it.
- **Other stale `TODO(brand-cleanup)` headers** (`DetailModal.vue`, `PanelApp.vue`, `useOverlay.ts`, `i18nMessages.ts`). Tracked in #796 as a follow-up so this PR doesn't balloon past its stated scope.
- **Icon tone.** `ShieldAlert` reads slightly warning-ish for what is really informational. Matches the existing `.settings-v2-env-notice` family, so consistency wins for now; revisit if descriptions become general help text.

### Verification on 70e59e2
- `pnpm run typecheck` ?
- `pnpm run lint` ?
- `pnpm run test` - **1603 passed** (was 1599 pre-tests, +4 new)

PR #815, opened 2026-06-03
fix(settings): don't recreate ~/ComfyUI-Shared when user has custom paths (#699)

### LINKED ISSUE #699: [P&P] Don't auto-recreate ComfyUI-Shared / ComfyUI-Installs when custom paths are set
Reported in #temp-desktop-bug-dump (Prompt & Play QA) by Alexis Rolland.

Using custom folder paths to manage models and installs, but the app keeps recreating these folders every time:
- `C:\Users\Alexis\ComfyUI-Shared`
- `C:\Users\Alexis\ComfyUI-Installs`

Allow deleting them / don't recreate the default folders when custom paths are configured.

Slack: https://comfy-organization.slack.com/archives/C0B6R9Y4YDT/p1780006506781319

deepme987:
## Summary
Alexis Rolland reported in P&P QA that she'd set custom model / input / output paths and deleted `~/ComfyUI-Shared`, but the empty folder kept reappearing every time the app loaded settings.

Two compounding bugs in `src/main/settings.ts`:

1. **Force-injection.** `sanitizeModelsDirs` always appended the system default (`~/ComfyUI-Shared/models`) to the user's `modelsDirs` list. The caller `load()` did the same again via its `else if` branch — so a user who explicitly removed the default got it re-injected, twice.

2. **Unconditional mkdir.** `load()` then `mkdirSync(systemDefault)` + the model-type subfolders, recreating the empty `~/ComfyUI-Shared` tree. The same pattern applied to `inputDir` / `outputDir`: when the user's setting was unset, the code fell back to `defaults[key]` and `mkdir`'d that.

## Fix
- Dropped the force-append in `sanitizeModelsDirs` and in the `load()` `else if`. A non-empty `modelsDirs` reflects user intent — honour it. Empty / missing `modelsDirs` still falls back to `[systemDefault]` (unchanged).
- Gate the system-default `mkdir` on `usesSystemDefault` (whether `systemDefault` is actually in `result.modelsDirs`).
- Gate the input/output `mkdir` on either an explicit non-default user value OR `usesSystemDefault`. Windows's `sanitizeUserDefaultPath` replaces an undefined `inputDir`/`outputDir` with `defaults[key]`, so a non-empty value alone isn't a signal the user opted in to the default location — compare against `defaults[key]` to detect that case.

## Cross-platform
Same logic runs on Windows + macOS — both use `homeDir() / ComfyUI-Shared` and the fix touches platform-agnostic code paths. (#699 explicitly required this.)

## Tests
- **`rewrites copied foreign-user defaults on Windows only`** — updated `expectedModelsDirs` to drop the now-absent trailing system-default entry on both branches.
- **`appends system default when missing from user list`** — renamed to **`preserves user-only modelsDirs without re-injecting system default (#699)`**. It now asserts both halves of the bug: the user's array stays as-is, AND `~/ComfyUI-Shared` is not recreated on disk.
- `injects system default when modelsDirs is empty` still passes (empty → fallback behaviour unchanged).
- All 9 `src/main/settings.test.ts` + 4 `src/main/lib/oem.test.ts` tests green. `typecheck:node` + `lint` clean.

Closes #699


Kosinkadink:
For this one, the only case we want to create it is if some for some reason no other paths that the user had assigned as model paths exist anymore (so that the app is never in a state where there are no directories to handle models)

Kosinkadink (approved):

PR #829, opened 2026-06-03
feat(installer): wire BMP artwork via NSH (bypasses ToDesktop CLI schema reject)

deepme987:
## Summary

Wires the installer/uninstaller artwork BMPs that have been sitting orphaned in \`resources/\` since #614, without re-triggering the ToDesktop config-validation failure that killed builds in #810/#817.

## Why this route

PR **#810** added \`installerHeader\`, \`installerSidebar\`, \`uninstallerSidebar\` to \`todesktop.json\`. ToDesktop CLI 1.22.0 (pinned in \`build-release.yml\`) rejects those keys at config validation in <30s:

\`\`\`
todesktop.json invalid.
ADDITIONAL PROPERTY must NOT have additional properties
  > installerHeader is not expected to be here!
\`\`\`

PR **#817** stripped them out to unblock releases. Local electron-builder builds auto-discover the BMPs from \`directories.buildResources\` (= \`resources/\`), but ToDesktop reads its own config schema, not \`electron-builder.yml\`, so the auto-discovery never kicks in on ToDesktop builds.

\`todesktop.json\` still accepts \`nsisInclude: ./scripts/installer.nsh\` — and that NSH is the one input ToDesktop *does* feed to electron-builder's NSIS step. So we set the MUI bitmaps directly via NSIS-native \`!define\` directives inside the NSH. ToDesktop never sees the offending keys; NSIS resolves the paths.

## What's in the patch (\`scripts/installer.nsh\`, +24 / -0)

\`\`\`nsis
!ifndef MUI_HEADERIMAGE
  !define MUI_HEADERIMAGE
!endif
!ifndef MUI_HEADERIMAGE_BITMAP
  !define MUI_HEADERIMAGE_BITMAP \"\${BUILD_RESOURCES_DIR}\installerHeader.bmp\"
!endif
!ifndef MUI_WELCOMEFINISHPAGE_BITMAP
  !define MUI_WELCOMEFINISHPAGE_BITMAP \"\${BUILD_RESOURCES_DIR}\installerSidebar.bmp\"
!endif
!ifndef MUI_UNWELCOMEFINISHPAGE_BITMAP
  !define MUI_UNWELCOMEFINISHPAGE_BITMAP \"\${BUILD_RESOURCES_DIR}\uninstallerSidebar.bmp\"
!endif
\`\`\`

- Placed at the top of \`installer.nsh\` (after the existing \`!include\`s, before any MUI page macros are processed).
- \`\${BUILD_RESOURCES_DIR}\` is the same variable already used for \`vc_redist_version.nsh\` two lines above — proven resolution mechanism.
- \`!ifndef\` guards protect against a future electron-builder bump pre-setting any of these (which would otherwise trip the \`-WX\` \"macro redefined\" failure).

## BMPs already in place

\`resources/installerHeader.bmp\`, \`installerSidebar.bmp\`, \`uninstallerSidebar.bmp\` — added in #614, dimensions already match MUI2's expected defaults exactly: header ≈ 25.8 KB ≈ **150x57x24-bit**, sidebars ≈ 154.5 KB ≈ **164x314x24-bit**.

## What this does *not* touch

- \`todesktop.json\` — left untouched, ToDesktop CLI schema validation still passes.
- \`electron-builder.yml\` — \`nsis:\` block keeps only the behavior keys (\`oneClick\`, \`perMachine\`, etc.). No art keys re-introduced.

## Test plan

- [x] \`pnpm typecheck\` + \`pnpm lint\` clean (pre-commit hook).
- [ ] **Local \`pnpm run build:win\`** — running now; this proves NSIS compiles the NSH without errors. (Will post the result in a comment when it finishes.)
- [ ] **Manual install/uninstall visual smoke test on Windows** — confirm the header (top of install pages) and welcome/finish sidebars actually render.
- [ ] **ToDesktop ephemeral build** — the real gating check (since past attempts died there). Once local + visual pass, an ephemeral confirms config validation still clears + ToDesktop's NSIS pipeline picks up the BMPs.

Refs: #614 (BMPs added), #810 (broken approach via \`todesktop.json\`), #817 (cleanup that explicitly tracked this NSH route as the follow-up).


deepme987:
Local `pnpm run build:win` ✅ — exit code 0, no NSIS warnings, both installer + uninstaller produced and signed.

Confirms:
- NSH syntax is valid under `-WX`.
- `${BUILD_RESOURCES_DIR}` resolves and all three BMPs were found.
- electron-builder's NSIS template accepts the MUI directives via nsisInclude — no conflict.

Artifact: `dist/ComfyUI-Desktop-2.0-0.6.13-win-x64.exe` (~124 MB).

Remaining checks:
- Visual smoke test on a real Windows install — confirm the welcome/finish sidebars and header BMPs render correctly.
- ToDesktop ephemeral build — the actual gate (past attempts died at the CLI's config schema; this PR doesn't touch `todesktop.json`, so we expect it to clear).

Issue #843, opened 2026-06-04
Migration progress is weird


Kosinkadink:
The steps for the migration are not defiend well at all, they are just one word lowercase things. We used to have such nice names for operation steps, what happened?

Example, this "source" I am pretty sure is the step where it fetches the git repo information such that the migrated one can continue.
https://github.com/user-attachments/assets/56539f06-a050-4c93-abb9-1c0662746ba0

PR #845, opened 2026-06-04
fix(adopt): populate comfyVersion and register step labels for legacy migration

### LINKED ISSUE #842: MIgrated installs think they are not on latest when they are
Example: 0.24.0 instead of v0.24.0, and thinks it can update

Image

Kosinkadink:
Fixes #842 and #843.

## Issue #842 — Migrated installs always say "update available"

The Legacy Desktop adoption flow recorded only the bare `__version__` string from `comfyui_version.py` (e.g. `"0.24.0"`) on the resulting installation record. The release-cache `installedTag` derivation then fell back to that bare string, while `info.latestTag` from GitHub is `"v0.24.0"`. A literal `!==` comparison in `isUpdateAvailable` wedged every adopted install into a permanent "update available" state immediately after adoption.

**Two complementary fixes**:

1. **`desktopAdopt.ts`** — after `updateComfyToStable` checks out the latest stable tag, resolve a real `{commit, baseTag, commitsAhead}` via `resolveLocalVersion` (the same call the standalone fresh-install `postInstall` already makes) and persist it as `comfyVersion` on the record. `getEffectiveInfo` then renders `installedTag` via `formatComfyVersion(cv, 'short')`, which returns the real git tag (`v0.24.0`) and matches the remote.
2. **`release-cache.ts`** — make the `installedTag !== info.latestTag` / `!== info.releaseName` check tolerant of a leading `v`. Covers any other code path that persists a bare version string and any legacy records already on disk from previous adoptions, so the fix lands for existing users without a data migration.

## Issue #843 — Migration progress is weird

`runAdoption` emits `sendProgress('backup', ...)`, `sendProgress('venv', ...)`, `sendProgress('source', ...)`, … but never calls `sendProgress('steps', { steps: [...] })` to register human-readable labels. The renderer fell back to displaying the raw phase id ("source", "venv", …), which is what the user saw in the screenshot.

**Fix**:

1. **`desktopAdopt.ts`** — emit a single `sendProgress('steps', { steps: [...] })` at the top of `runAdoption`, listing every phase the flow goes through with i18n labels. The macOS-only `tcc` step is conditionally included.
2. **`locales/en.json`** — add short step labels under `desktop.adoptStep*` (rendered in the step timeline) and longer active-step captions under `progress.phaseLabel.*` (rendered as `friendlyCaption` while a phase is active). Mirrors how the standalone install flow registers its steps + captions.
3. **`progressWeights.ts`** — add fingerprint entries for the adoption phase set (mac and non-mac variants) so the global progress bar uses meaningful per-step weights instead of the equal-weights fallback. `source` + `comfy-update` + `requirements` dominate wall time, the rest are fast.

## Tests

- `release-cache.test.ts`: 4 new cases covering v-prefix tolerance (bare ↔ prefixed in both directions, real-update detection across prefix variants, cross-channel display).
- `desktopAdopt.test.ts`: 4 new cases — `comfyVersion` populated after successful adoption, `comfyVersion` omitted when no `.git` dir, adoption non-fatal when version resolution throws, and step labels registered at the top of `runAdoption`.

All existing tests still pass: typecheck/lint/build/test all green (1746 tests across 126 files).

Amp-Thread-ID: T-019e9063-1a8e-767c-b498-48f5c62eec58

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -340,7 +340,17 @@
       "migration": "Migrating your Legacy Desktop install…",
       "restore-nodes": "Restoring custom nodes from your snapshot…",
       "restore-pip": "Reinstalling Python packages from your snapshot…",
-      "launch": "Starting ComfyUI…"
+      "launch": "Starting ComfyUI…",
+      "backup": "Backing up Legacy Desktop state…",
+      "tcc": "Checking folder access…",
+      "venv": "Validating Legacy Python environment…",
+      "snapshot": "Capturing a forensic snapshot…",
+      "allocate": "Allocating install folder…",
+      "source": "Staging ComfyUI source…",
+      "comfy-update": "Updating ComfyUI to the latest stable release…",
+      "requirements": "Installing Python requirements…",
+      "settings": "Carrying over Legacy Desktop settings…",
+      "register": "Registering the adopted installation…"
     }
   },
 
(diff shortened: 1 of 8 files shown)


Kosinkadink:
Combined into #849 (same commits flattened onto a single linear branch). Closing.

PR #878, opened 2026-06-04
feat(mirrors): inject HF_ENDPOINT mirror when useChineseMirrors is on

deepme987:
## Summary

When a user opts into the China mirror flow (\`useChineseMirrors=true\`), inject \`HF_ENDPOINT=https://hf-mirror.com\` into every spawned ComfyUI process so HuggingFace downloads route through the mirror.

Current behaviour: opt-in only rewrites pip's \`--index-url\` and the Comfy-Org git clone URL. HuggingFace, the second-largest external dependency (model downloads, hub API), still points at huggingface.co with no fallback. This PR closes that gap.

## Why

Cohort analysis over a 24h launch window:

| Cohort | Users | Users hitting external-network errors |
|---|---|---|
| opted_in (mirror on) | 1,225 | 61 (5.0%) |
| no_prompt | 1,461 | 17 (1.2%) |

Even users who explicitly opt into the mirror still hit network failures at ~4x the baseline rate. The errors trace to dependencies the mirror doesn't currently cover — HuggingFace is one. \`HF_ENDPOINT\` is the documented hub env var (https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables) and \`hf-mirror.com\` is the standard community mirror.

## Behaviour

- \`useChineseMirrors=true\` and no user override → injects \`HF_ENDPOINT=https://hf-mirror.com\`
- \`useChineseMirrors=true\` but the user already set \`HF_ENDPOINT\` in their shell → keeps the user value, no override
- \`useChineseMirrors=false\` or unset → no-op

## Test plan

- [x] \`pnpm vitest run buildLaunchEnv.test.ts\` — 5/5 pass (covers off/unset/on-default/user-override/other-keys)
- [x] \`pnpm lint\` — clean
- [ ] Manual: build, set \`useChineseMirrors=true\`, start an instance, attach to the Python process and confirm \`os.environ['HF_ENDPOINT'] == 'https://hf-mirror.com'\`

### DIFF
--- a/src/main/lib/ipc/buildLaunchEnv.test.ts
+++ b/src/main/lib/ipc/buildLaunchEnv.test.ts
@@ -1,0 +1,77 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+vi.mock('electron', () => ({
+  app: {
+    isPackaged: false,
+    getPath: () => '/tmp',
+    getVersion: () => '0.0.0-test',
+    getLocale: () => 'en',
+  },
+  ipcMain: { handle: vi.fn(), on: vi.fn(), off: vi.fn() },
+  dialog: {},
+  shell: {},
+  BrowserWindow: { getAllWindows: () => [] },
+  nativeTheme: { on: vi.fn(), shouldUseDarkColors: false },
+}))
+
+const settingsState: Record<string, unknown> = {}
+vi.mock('../../settings', () => ({
+  get: (key: string) => settingsState[key],
+  set: (key: string, value: unknown) => {
+    settingsState[key] = value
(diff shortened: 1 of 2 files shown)


deepme987:
Closing per re-scoping discussion. The cohort error data shows zero `huggingface.co` errors in the top 15 — `raw.githubusercontent.com` dominates. PR #880 (Manager CDN channel) directly addresses the data-driven pain point; this one is forward-looking insurance for a downstream cliff most users aren't reaching yet. May re-open once #880 and the `desktop-assets.comfy.org` install-wizard fallback land and we see HF show up in the error distribution.

PR #880, opened 2026-06-04
feat(mirrors): seed Manager config with CDN channel when useChineseMirrors is on

deepme987:
## Summary

When a user opts into the China mirror flow (\`useChineseMirrors=true\`), seed ComfyUI-Manager's \`config.ini\` with a mirror-friendly channel URL and SSL bypass, so its node-list fetch at boot doesn't hang or fail on \`raw.githubusercontent.com\`.

Writes only when no \`config.ini\` exists at the target path — returning users with their own customized config keep full control.

## Why

Cohort analysis over a 24h launch window:

| Cohort | Users | Users hitting external-network errors |
|---|---|---|
| opted_in (mirror on) | 1,225 | 61 (5.0%) |
| no_prompt | 1,461 | 17 (1.2%) |

For the opted-in cohort, 83% of those network errors are \`aiohttp ... raw.githubusercontent.com\` — ComfyUI-Manager's node-list fetch. The existing mirror flow rewrites pip's index and the Comfy-Org git remotes, but Manager hardcodes the GitHub raw channel and bypasses any other mirror config. When that DNS hangs, boot stalls and users quit — strongly correlated with the boot→view drop in the activation funnel.

## How

\`ComfyUI-Manager\` already supports \`channel_url\` and \`bypass_ssl\` in its config.ini. This PR seeds those values for opted-in users:

- \`channel_url = https://cdn.jsdelivr.net/gh/ltdrdata/ComfyUI-Manager@main\` — jsdelivr's CDN proxy serves arbitrary GitHub raw paths and is reachable from regions where raw.githubusercontent.com is not. Keeps fresh node-list updates (no offline mode), just routes them through a working CDN.
- \`bypass_ssl = true\` — kills SSL cert chain failures from corporate proxies / outdated CA bundles (next-biggest error category after DNS).
- \`network_mode = public\` — explicit default, makes the config self-documenting.

The config is written at launch time, gated on \`settings.get('useChineseMirrors') === true\`, only when the target file doesn't already exist. Idempotent — the file-exists guard short-circuits every subsequent launch.

Targets the modern ComfyUI system-user-api path: \`<installPath>/ComfyUI/user/__manager/config.ini\`.

## Test plan

- [x] \`pnpm vitest run managerConfig.test.ts\` — 4/4 pass (writes config, creates intermediate dirs, does not overwrite existing config, correct path)
- [x] \`pnpm lint\` — clean
- [ ] Manual: build, opt into mirrors at first-use, launch an instance, confirm \`<installPath>/ComfyUI/user/__manager/config.ini\` exists and Manager pulls node-list from jsdelivr instead of github

### DIFF
--- a/src/main/lib/ipc/sessionActions/launch.ts
+++ b/src/main/lib/ipc/sessionActions/launch.ts
@@ -26,6 +26,7 @@ import { rotateLogFiles, getLogDir } from '../../logRotation'
 import { createExecutionTap } from '../../executionTap'
 import { clearCrash, recordCrash } from '../../crashBuffer'
 import * as telemetry from '../../telemetry'
+import { ensureManagerMirrorConfig } from '../../managerConfig'
 import type { WriteStream } from 'fs'
 
 // Feature flags injected on every spawned ComfyUI, gated by the running
@@ -115,6 +116,14 @@ export async function handleLaunch({ event, installationId, inst: instArg, actio
       } catch {
         // Schema not available — pass args as-is.
       }
+    }
+  }
+
+  if (settings.get('useChineseMirrors') === true) {
+    try {
+      await ensureManagerMirrorConfig(inst.installPath)
+    } catch (err) {
+      console.warn('Failed to seed ComfyUI-Manager mirror config:', err)
     }
   }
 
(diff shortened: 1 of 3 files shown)


deepme987:
Closing in favor of a single consolidated PR that bundles this + #881 + the now-provisioned GCS mirror URL + binary download fallback + review fixes from both prior PRs.

PR #881, opened 2026-06-04
feat(install): wire mirror-URL fallback for R2 manifest fetches

deepme987:
## Summary

When the primary R2 host is unreachable (connection-reset, throttling, regional blocks), retry the install-wizard JSON fetches once against a configured mirror host before falling back to the on-disk cache. Primary stays the source of truth and the ETag cache key — the mirror only catches the network-fail case.

\`R2_MIRROR_BASE_URL\` is empty by default; the fallback activates with zero further code change once infra populates it with a mirror bucket URL.

## Why

A user testing without VPN cannot load the install-wizard release dropdown — \`get-field-options\` fails with \`net::ERR_CONNECTION_RESET\`. The failing URL is \`https://desktop-assets.comfy.org/standalone-environments/latest.json\` (Cloudflare R2, no ICP-licensed CDN in front), which means the install wizard never populates and fresh installs are blocked.

This affects every JSON manifest fetch in the install path:
- Boot ETag pre-warm (\`fetchJSON\` in \`registerIpcHandlers\`)
- Release dropdown population (\`R2_BASE_URL/latest.json\`)
- Per-vendor release list (\`R2_BASE_URL/<vendor>/releases.json\`)

## How

- \`fetchJSON\` now accepts an optional \`mirrorUrl\`. On a primary network failure, it retries once against the mirror. Both paths share the same ETag cache key so subsequent revalidation continues to negotiate against the primary.
- New helper \`r2MirrorUrl(primaryUrl)\` builds the mirror URL by swapping the base prefix. Returns \`undefined\` when no mirror is configured or the URL isn't in the R2 namespace.
- Three call sites updated to pass \`mirrorUrl: r2MirrorUrl(url)\`.
- \`R2_MIRROR_BASE_URL\` is currently empty — no behavior change at runtime until infra populates it.

## Out of scope

Binary standalone bundles (the multi-GB \`.7z\` / \`.tar.gz\` files) go through a different download path, not \`fetchJSON\`. Those continue to hit the primary host only. Mirroring them is a follow-up that depends on the same mirror bucket being available.

## Test plan

- [x] \`pnpm vitest run fetch.test.ts updateSections.test.ts\` — 19/19 pass (5 new fetch-fallback tests + 14 existing)
- [x] \`pnpm lint\` — clean
- [ ] Once infra provisions a mirror, set \`R2_MIRROR_BASE_URL\` and verify on a network that connection-resets the primary

### DIFF
--- a/src/main/lib/fetch.test.ts
+++ b/src/main/lib/fetch.test.ts
@@ -1,0 +1,100 @@
+import { EventEmitter } from 'events'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+vi.mock('./paths', () => ({
+  cacheDir: () => '/tmp/desktop-test-cache',
+}))
+
+vi.mock('./safe-file', () => ({
+  writeFileSafe: vi.fn(),
+}))
+
+interface FakeRequest extends EventEmitter {
+  setHeader: ReturnType<typeof vi.fn>
+  end: ReturnType<typeof vi.fn>
+  __resolvedUrl: string
+}
+
+const requests: FakeRequest[] = []
+
+vi.mock('electron', () => ({
+  net: {
+    request: vi.fn((opts: { url: string }) => {
+      const req = Object.assign(new EventEmitter(), {
+        setHeader: vi.fn(),
+        end: vi.fn(),
+        __resolvedUrl: opts.url,
+      }) as FakeRequest
+      requests.push(req)
+      return req
+    }),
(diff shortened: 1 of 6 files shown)


deepme987:
Closing in favor of a single consolidated PR that bundles this + #880 + the now-provisioned GCS mirror URL + binary download fallback + review fixes from both prior PRs.

PR #883, opened 2026-06-04
test: revert productName to isolate macOS auto-update failure

deepme987:
**Test build, not for merge.**

## Why
Windows migrates App 1 (`com.todesktop.241012ess7yxs0e`, 0.9.x) to the new build in place cleanly; macOS does not. Dave's hunch is the `productName` change from #616 (`ComfyUI Desktop 2.0` → `Comfy Desktop`), which renames the `.app` bundle / CFBundleName and could break Squirrel.Mac's in-place replace.

## What
Reverts **only** `productName` back to the pre-#616 value (`ComfyUI Desktop 2.0`). Everything else is held constant:
- `appId` stays App 1's `com.todesktop.241012ess7yxs0e`
- signing / notarization unchanged
- userData path unaffected either way (runtime app name is package.json `name` = `comfyui-desktop-2`, no `productName` in package.json)

So this is a clean single-variable isolation.

## How to read the result
Build this branch in ToDesktop, offer it as an update to a Mac sitting on App 1:
- **Updates cleanly** → the rename is the cause; we decide bundle name vs Mac auto-update.
- **Still fails** → it's not productName. Next suspect is signing (same Apple Team ID between App 1 and the new build?). Pull the ShipIt log to confirm: `~/Library/Caches/com.todesktop.241012ess7yxs0e.ShipIt/ShipIt_stderr.log`.

## Caveat
This reverts to the beta's pre-rename value (`ComfyUI Desktop 2.0`). If App 1 actually shipped under a different bundle name (e.g. `ComfyUI`), the truest Squirrel.Mac continuity test should match that instead. Worth confirming App 1's installed productName with Dave.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -1,5 +1,5 @@
 appId: com.todesktop.241012ess7yxs0e
-productName: Comfy Desktop
+productName: ComfyUI Desktop 2.0
 directories:
   buildResources: resources
 files:
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -1,5 +1,5 @@
 appId: com.todesktop.241012ess7yxs0e
-productName: ComfyUI Desktop 2.0
+productName: ComfyUI
 directories:
   buildResources: resources
 files:
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "comfyui-desktop-2",
-  "version": "1.0.2",
+  "version": "1.0.2-mactest.1",
   "description": "Comfy Desktop",
   "author": {
     "name": "Comfy Org",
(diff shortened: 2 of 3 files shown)


deepme987:
Corrected the test target: App 1 (`com.todesktop.241012ess7yxs0e`) ships as **ComfyUI.app**, not the beta's `ComfyUI Desktop 2.0`. Squirrel.Mac replaces the bundle in place, so the continuity test needs the new build's bundle name to match what's installed on the user's machine. `productName` is now **`ComfyUI`** (was reverted to `ComfyUI Desktop 2.0` in the first commit, fixed in the second). Caveat in the description is resolved.

deepme987:
confirmed: the issue IS the name. unfortunately, there's no nice way to do it from our end. toDesktop is looking whether this is possible to fix. otherwise, it's just an inconvenience we tank

PR #899, opened 2026-06-04
feat: interactive Console in install Settings (supports_terminal)

### LINKED ISSUE #888: Add COnsole support to Settings (and comfyui_frontend feature flag to have it work in the bottom Logs of the frontend)
The previous console implementation in legacy desktop had a glaring issue: you could type 'exit' and kill the console, making it unaccessible until you restart ComfyUI desktop. We'd want a way to tell the user they killed their session, and that pulling it up again (on the frontend) will restart it; if we can get a button for restarting the console instance when we kill it. On the desktop side (not comfyui_frontend), we can render the button as we please

Kosinkadink:
Adds an interactive **Console** tab to each install's Settings, plus the backend that powers it.

## What this does
- New **Console** tab in the per-install Settings surface, backed by an `xterm` view.
- One shared shell **per installation** (keyed by `installationId`), visible from every window attached to that install, so the same session is never duplicated or confused across windows.
- The shell is **independent of the ComfyUI process** — it works whether ComfyUI is running or stopped (most console use happens while ComfyUI is off, since changes usually need a reboot to take effect).

## Fixes the legacy "typed `exit` and the console is dead" bug
In legacy Desktop, typing `exit` killed the PTY and it stayed dead until the app restarted. Now:
- `onExit` marks the session ended and notifies every subscriber instead of caching a dead PTY.
- Writes after exit are dropped.
- The pane shows a **"session ended"** banner with a **Restart console** button that respawns a fresh shell.
- Guards against the restart race where an old PTY's async exit could clobber the new session.

## Frontend feature flag (served frontend bottom panel)
So the served ComfyUI frontend can also surface its bottom-panel terminal, the launcher injects the `supports_terminal` CLI feature flag (validated against the install's `--list-feature-flags` registry). The interactive transport is the existing `__comfyDesktop2.Terminal` preload bridge; the flag only gates visibility.

This piece depends on two not-yet-merged PRs (branches pushed, PRs intentionally not opened yet):
- **ComfyUI** `feat/supports-terminal-flag` — registers `supports_terminal` as a bool CLI feature flag.
- **ComfyUI_frontend** `feat/desktop-terminal-feature-flag` — gates the bottom-panel terminal on `supports_terminal` and adds host-agnostic terminal transport + auto-restart.

The desktop Console tab in this PR works on its own (via `window.api.terminal*`) and does not require those to land first; they only enable the served frontend's bottom panel.

## Packaging
`node-pty` is externalized in the main build and `asar`-unpacked for distribution.

## Validation
`typecheck`, `lint`, `build`, and the full unit suite (1808 tests) pass.

Fixes #888

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -19,6 +19,7 @@ extraResources:
     to: lib
 asarUnpack:
   - node_modules/7zip-bin/**/*
+  - node_modules/node-pty/**/*
 afterPack: ./scripts/afterPack.mjs
 publish:
   provider: github
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -17,6 +17,11 @@ export default defineConfig({
   main: {
     build: {
       sourcemap: 'hidden',
+      rollupOptions: {
+        // node-pty is a native addon; it must stay external (its .node binary
+        // can't be bundled) and is loaded from the unpacked node_modules.
+        external: ['node-pty'],
+      },
     },
   },
   preload: {
--- a/locales/en.json
+++ b/locales/en.json
@@ -72,6 +72,7 @@
     "tabUpdate": "Update",
     "tabSnapshots": "Snapshots",
     "tabStorage": "Storage",
+    "tabConsole": "Console",
     "relaunch": "Relaunch",
     "more": "More",
     "diskUsage": "Disk Usage",
@@ -362,7 +363,9 @@
     "disconnect": "Disconnect",
     "connectedTo": "Connected to {url}",
     "processExited": "Process exited",
-    "processCrashed": "Process crashed (exit code {code})"
+    "processCrashed": "Process crashed (exit code {code})",
+    "sessionEnded": "This console session ended. Restart it to keep using the shell.",
+    "restartSession": "Restart console"
   },
 
   "settingsModal": {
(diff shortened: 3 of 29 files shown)


deepme987:
Reviewing this for the Day-3 launch feedback ("Why u delete cmd?" from Typeform users + similar Reddit complaints).

**Verdict from product (Deep): the desktop-side CSS-injection approach is the right call — which is exactly what `comfyTerminalContentScript.ts` already does.** So this PR isn't just a stopgap — it IS the right shape for the inline-terminal experience.

**Two suggestions for when you're back:**

1. **Promote the stopgap from "only when feature flag missing" to "always inject" (with frontend de-dupe).** Currently `attach.ts:390` only injects when `supports_terminal` is *not* available. The risk of always-injecting is duplicate tabs once the companion ComfyUI / ComfyUI_frontend PRs land. To avoid that:
   - In the injected script, check whether `bottomPanelTabs` already has an entry with `id: 'command-terminal'` (or similar) and bail out if so.
   - That way: new-frontend renders its own terminal natively, old-frontend gets our stopgap. No duplication, no gating on a feature flag the user may not have yet.

2. **Discoverability: surface the Terminal entry point in the launcher chrome too.** The bottom-panel tab is great for users who scroll down to find it; first-time users coming from the legacy Desktop don't know to look there. Cheap add: a per-install Console icon button in the title bar (next to the existing chrome icons) that calls the same backend bridge. We can either land it here as a follow-up commit OR I can spin it as a tiny standalone PR — your preference. Day-3 Typeform: "Why u delete cmd?" was the literal phrasing, so the affordance matters.

Will hold off on creating a duplicate PR for the inline injection — yours is the canonical implementation. Will follow up with Deep on the discoverability question once you're back to confirm direction.

Day-3 context: 30k DAU / 65k total, terminal-related complaints are top-tier in the Typeform feed.

deepme987:
closing -- superseeded by another PR

PR #938, opened 2026-06-05
feat(storage): per-instance model dirs + consistent clickable storage path rows

### LINKED ISSUE #909: [Feat] Per-Instance model directories
Allow users to add model directories per-instance.
Right now, you can do it via `--extra-model-paths-config` ; only need to get it as a UI feature

https://discord.com/channels/1218270712402415686/1512196069809852596/1512295549775122494

Kosinkadink:
Closes #909.

## What

Reworks the Storage UX so that **local (per-instance) vs shared** model/input/output directories are clear, and makes directory paths consistent and openable across the app.

The original ask was a **per-instance model directories** UI surfaced when an installation opts out of shared models (`Use Shared Models` off). Previously, turning shared models off left an install able to see only its own `models` folder; the only way to add extra model paths was the `--extra-model-paths-config` CLI flag. While implementing it, the storage UI was tightened up so models, inputs, and outputs all behave consistently.

## Per-instance model directories

- New per-install field `modelDirs: string[]` (extra/external model dirs) plus `modelDirsPrimary` to mark which dir is primary. These are only meaningful when `useSharedModels === false`.
- The install-owned `<comfy>/models` directory is **computed/displayed, not persisted** — only external extra dirs are stored. This keeps clones safe (a cloned install derives its own local path instead of pointing back at the original).
- The install-owned models row is always shown as a **locked, non-removable row** (it can never be removed), tagged with an **"Instance"** pill + tooltip explaining it is the instance's own unshared folder, and uses a `FolderLock` icon.
- The locked install-owned row renders at the **bottom** of the list unless it is primary (primary always stays on top).
- Adding/removing/make-primary for external dirs mirrors the global Shared Models list, persisting through the standard `update-field` -> `updateInstallation` path.
- At launch, when shared models is off the dirs are written to a **per-install YAML** under `dataDir()/instance-model-paths/<id>.yaml` and passed via `--extra-model-paths-config`. The primary dir is written `is_default: true` so newly downloaded models save there. The YAML is cleaned up when the install is deleted.
- The install-owned models dir is also shown as a locked instance row when **shared models is ON**, since ComfyUI still uses `<comfy>/models` in that mode — making it clear that folder is still active but not editable.

## Input / output directories

- Per-instance and shared input/output now render as **readonly path rows** (consistent with models / About Location) instead of untypable text fields.
- When shared input/output is toggled off, the install's default input/output locations are **computed and displayed** (not persisted), so clones derive their own local paths.
- `Directory` wording is preserved (no longer changes to "Folder").

## Consistent, clickable path rows

- New shared `StorageDirRow.vue` readonly path-row component used across storage UIs.
- Path **text** is clickable to open the folder in the OS file manager (Explorer / Finder / etc.). Only the path text is the click target (not the whole bar) for clarity.
- Applied across: per-instance models/input/output, shared input/output, Desktop Settings storage rows, About/Status **Location**, Advanced **Cache Directory**, and **Default Install Location**.

## Layout

- The `Use Shared Models` toggle is grouped with the models section, and the shared input/output toggle is grouped with the input/output section (previously the toggles were grouped together away from their relevant sections).

## Tests

- `src/main/lib/models.test.ts`: per-install YAML path, custom YAML target, primary/`is_default` placement, no global-YAML write.
- `StoragePane.test.ts`, `GlobalSettingsView.test.ts`, `StatusFactPanel.test.ts` extended for the new rows, locked instance row, clickable paths, and toggle grouping.
- `pnpm typecheck` / `lint` / `build` / `test` all green (2046 passing).

https://github.com/user-attachments/assets/906f5626-05ea-4a27-ae75-4ffc739dfc4f


Kosinkadink:
Note: this has **not been tested manually** yet — only verified via the automated suite (typecheck, lint, build, and unit tests). Manual end-to-end testing in the desktop app is still needed before merge.

deepme987 (approved):

Issue #902, opened 2026-06-05
Restore last-used ComfyUI instance on startup (allow closing last window without returning to dashboard)


Kosinkadink:
## Summary

On startup, restore the **last-used ComfyUI instance window** instead of always opening the dashboard — _assuming the last window the user had active was an instance window, not the dashboard_.

A prerequisite for this: today, pressing **X** on the last instance window **returns the user to the dashboard** rather than closing the window. To support "restore last instance on boot," we need a way to **close the last window without returning to the dashboard**, plus persisted state for which instance (if any) was last used.

## Current behavior

- On boot the app opens the dashboard/chooser window (`openChooserHostWindow` / `openOrFocusChooserHostWindow`).
- Pressing **X** on the last instance window detaches it back to the dashboard instead of closing — see the `close` handler + `shouldDetachLastInstallWindowToDashboard` in `src/main/host/createHostWindow.ts`, and `returnToDashboard` / `_detachInstallImpl` in `src/main/host/detach.ts`.
- We already track some MRU/last-used info: `lastFocusedInstallationId` (in-memory, `src/main/host/registry.ts`) and persistent `lastLaunchedAt` on installation records (`src/main/installations.ts`).

## Desired behavior

1. **Persist** the last-used instance and whether the last active surface was an instance window vs. the dashboard (must survive quit — in-memory MRU is lost on shutdown).
2. **On startup**, if the last active surface was an instance window, restore/open that instance's window instead of the dashboard. If it was the dashboard (or there is no last instance), open the dashboard as today.
3. **Allow closing the last instance window without returning to the dashboard**, so the "last window used" can legitimately be a closed instance window rather than always being coerced back to the dashboard.

## Open questions / considerations

- Should restoring **auto-launch** the ComfyUI process, or just open the window on its (stopped) lifecycle surface and let the user start it?
- How do we distinguish "user pressed X to close/quit" from "return to dashboard"? Options: change X semantics + add an explicit "Return to dashboard" action, or a setting.
- Multi-window: "last used" = most recently focused instance window (`lastFocusedInstallationId`)?
- Should this be behind a setting (e.g. **"Reopen last instance on launch"**), default on/off?

## Relevant code

- **Startup:** `app.whenReady` boot flow in `src/main/index.ts`; `openOrFocusChooserHostWindow` / `openOrFocusAnyHostWindow` in `src/main/host/registry.ts`.
[...]

PR #939, opened 2026-06-05
feat: restore last-used instance on startup (allow closing last window without forced dashboard) (#902)

Kosinkadink:
## Summary

Implements #902: on startup, restore the **last-used ComfyUI instance window** instead of always opening the dashboard — and, as the prerequisite, allow **closing the last instance window without being forced back to the dashboard**.

Previously, pressing ✕ on the last instance window always detached it back to the dashboard, so the window could never truly "go away" and we couldn't remember that the user last left an instance window.

### Closing the last window
- Closing the last instance window now **quits Desktop**, but the confirm is a **three-way choice** so quitting is never the only option:
  - **Quit Desktop** (primary/danger) → tear down + quit
  - **Return to Dashboard** (secondary) → flip the window in place to the chooser (old behavior, still available)
  - **Cancel** → keep the window open
- Applies to both the OS ✕ and the File-menu **Close Window**. The picker **Home** icon still returns to the dashboard as before.
- Non-last windows are unchanged (plain Close/Cancel confirm).

### Restore on startup
- New setting **"Reopen last instance on launch"** (Global Settings → General), **on by default**.
- The last active surface (dashboard vs. a specific instance) is persisted to a small state file in `stateDir()`, updated on window focus / attach / detach (skipped while quitting).
- On boot, when the setting is on and the user last left an instance window, that instance is **auto-launched and restored in place** on top of the freshly-opened chooser host — by reusing the dashboard's own launch path (the `picker-pick-install` deep link → `handleChooserPick`). That means it gets the same progress overlay, attach-claim, and in-place attach the user sees when clicking an install card.
- Falls back to the dashboard when: the setting is off, the remembered install was deleted, first-use isn't complete, or the launch fails.

## Open questions from the issue, resolved
- **Auto-launch vs. open-stopped-window** → auto-launch (reuses the existing launch path; no new host-construction surface).
- **Behind a setting / default?** → yes, `reopenLastInstanceOnLaunch`, default on.
- **X semantics** → ✕ on the last window quits, with an explicit "Return to Dashboard" escape in the confirm; Home icon unchanged.
- **Multi-window "last used"** → most-recently-focused surface wins.

## Implementation notes
- `src/main/lib/lastSession.ts` — new persisted last-active-surface state (mirrors `windowState.ts`).
- `systemModal` (main + preload + `SystemModalApp.vue`) — extended with an optional middle ("secondary") action, rendered via the existing `BaseAlert` three-button footer; new `openSystemModalChoiceAsync`.
- `createHostWindow.ts` / `detach.ts` — close handler + `confirmAndCloseHostWindow` now drive a `CloseWindowChoice` (`close` / `cancel` / `return-to-dashboard`); the old "always detach last window" branch is now only a fallback for paths that reach teardown without an explicit choice.
- `index.ts` — `openStartupSurface()` replaces the unconditional chooser open at boot.

## Validation
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run build`, `pnpm run test` all pass (1842 tests, incl. new `lastSession.test.ts` and updated close-handler unit tests).

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -396,6 +396,7 @@
     "closeQuitDownloads": "Downloading",
     "autoUpdate": "Check for updates on startup",
     "autoInstallUpdates": "Automatically install Desktop updates",
+    "reopenLastInstanceOnLaunch": "Reopen last instance on launch",
     "checkForUpdates": "Check for updates",
     "checkingForUpdates": "Checking…",
     "telemetry": "Telemetry",
--- a/locales/zh.json
+++ b/locales/zh.json
@@ -267,6 +267,7 @@
     "closeQuitConfirm": "退出",
     "autoUpdate": "启动时检查更新",
     "autoInstallUpdates": "自动安装桌面更新",
+    "reopenLastInstanceOnLaunch": "启动时重新打开上次的实例",
     "checkForUpdates": "检查更新",
     "checkingForUpdates": "检查中…",
     "telemetry": "遥测",
--- a/src/main/host/attach.ts
+++ b/src/main/host/attach.ts
@@ -7,6 +7,8 @@ import { TITLEBAR_BG } from '../lib/theme'
 import * as mainTelemetry from '../lib/telemetry'
 import { refreshCloudUserTier } from '../lib/userTier'
 import { forwardDatadogError } from '../lib/processErrorHandlers'
+import { isQuitInProgress } from '../lib/quit-state'
+import { recordInstanceSurface } from '../lib/lastSession'
 import { installationEvents, type InstallationRecord } from '../installations'
 import {
   dropInstallationIndex,
@@ -137,6 +139,10 @@ export function attachInstall(entry: ComfyWindowEntry, opts: AttachInstallOpts):
   // null) install on the next dock-icon click.
   if (comfyWindow.isFocused()) {
     setLastFocusedInstallationId(installationId)
+    // An attach onto the focused host makes this install the active surface;
+    // persist it so the next boot restores this instance (no fresh focus
+    // event fires for an in-place attach).
+    if (!isQuitInProgress()) recordInstanceSurface(installationId)
   }
 
   // OS-level window title is rebuilt whenever the page title or the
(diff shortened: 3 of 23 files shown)


deepme987:
> Behind a setting / default? → yes, reopenLastInstanceOnLaunch, default on.

Maybe it should be default off?

Kosinkadink:
⚠️ **Heads-up: the latest commit (`a90c851`) has not been runtime-tested.**

`feat: no dashboard flash when restoring last instance on boot` was verified only via `pnpm run typecheck`, `lint`, `test` (1854 passing), and `build` — I was not able to launch the actual Electron app to confirm the boot-into-instance flow visually.

Please manually smoke-test a real cold boot into a remembered instance before merging, specifically:
- Restoring a healthy last instance shows **no dashboard flash** (window stays hidden until the launch takeover is up).
- Fallback paths reveal the dashboard correctly: deleted install, missing launch action, console/external mode, launch error, and a slow/absent renderer (10s backstop).

deepme987:
Reviewed at a structural level for the Day-3 launch patch — leaving a quick read while you're out.

**What this is doing right:**
- Three-way close modal (Quit / Stay / Back to dashboard) is exactly the shape the Typeform complaints have been pointing at ("I want the program to CLOSE not return to dashboard"). This is the load-bearing fix for that whole category of feedback.
- `lastSession.ts` as a single-source module for the persisted last-launched-instance is the right call (vs. spreading state across the chooser).
- Setting handler in `registerSettingsHandlers.ts` + IPC plumbing in `api.ts` matches the existing snapshot-update pattern.

**Two small things to think about when you're back (not blockers):**

1. **Default-on vs default-off for the restore behavior?** Your comment thread on the issue (#902) said "default on, but toggleable." If the toggle is default-on AND the close-modal default is "Quit Desktop," the chained behavior for someone who never opens settings is: close last window → quits → relaunches into the same instance window. Wanted to surface this so you can sanity-check the combined default before merge.

2. **Telemetry on the three-way choice?** Knowing which of the three buttons users actually click in the first month would help us tune the default. A single `first_use.close_last_window_choice` event with `{choice: 'quit'|'stay'|'dashboard'}` would be cheap and answers a real question.

I'm shipping the **Hide Cloud from Picker** opt-out as a separate small PR (#961) since that's orthogonal and lands on a different cohort of complaints ("Dashboard is a nuisance, can't be disabled"). Day-3 patch can bundle both.

deepme987:
closing - superseeded by #964

PR #955, opened 2026-06-06
fix(menu): restore fullscreen accelerator across all OSes

### LINKED ISSUE #952: [Misc][UI] Allow F11 [Inspect window] if electron allows?
https://comfy-organization.slack.com/archives/C0ALBG79NFJ/p1780694806950789

MaanilVerma:
### Problem
Users reported **F11 does nothing** in the new build. The custom title bar
(`titleBarStyle: 'hidden'`) ships with the native app menu stripped — on
Win/Linux production the menu was set to `null`. That menu carried the
`togglefullscreen` role, whose default accelerator (F11 on Win/Linux,
Ctrl+Cmd+F on macOS) vanished with it. No menu → no accelerator → no
fullscreen. macOS only ever worked via the green traffic-light button, so
the breakage read as OS-specific.

### Fix
Restore fullscreen the canonical way (how VS Code/Chrome do it): keep the
role in the menu and let Electron bind the right key per platform. Works
window-globally regardless of focus, and is discoverable in the macOS
Window menu.

- **menu.ts** — add `{ role: 'togglefullscreen' }`; Win/Linux now install a
  minimal `View` menu (was `null`) so the accelerator has a carrier.
- **index.ts** — `browser-window-created` hook hides the menu bar on *every*
  window (incl. `window.open` checkout/OAuth popups) while keeping
  accelerators live, so no visible bar regresses anywhere.
- **menu.test.ts** — assert the fullscreen role per platform; drop the two
  tests that locked in the old `null`-menu (buggy) contract.

### Result
| Platform | Key | Status |
|---|---|---|
| Windows | F11 | ✅ fixed |
| Linux | F11 | ✅ fixed |
| macOS | Ctrl+Cmd+F | ✅ confirmed (F11 is OS-reserved) |

The existing `enter/leave-full-screen` handlers already keep the custom
title bar in sync, so no renderer changes were needed.

### Test plan
- [x] `pnpm test` (menu suite, 8 pass) · typecheck · lint
- [ ] Windows: F11 toggles fullscreen; Alt shows no menu bar; checkout/OAuth popups barless
- [ ] Linux: F11 toggles fullscreen
- [ ] macOS: Ctrl+Cmd+F toggles; traffic lights reflow correctly

closes #952

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1145,6 +1145,15 @@ if (app.isPackaged && !app.requestSingleInstanceLock()) {
       }
     })
 
+    // Hide the bar on every window (incl. window.open popups) while keeping
+    // its accelerators live — the app uses a custom title bar throughout.
+    if (process.platform !== 'darwin') {
+      app.on('browser-window-created', (_event, window) => {
+        window.setMenuBarVisibility(false)
+        window.autoHideMenuBar = true
+      })
+    }
+
     // Bring up main-process telemetry as early as possible so install/migrate
     // sub-step events can fire even before the renderer mounts.
     //
--- a/src/main/menu.test.ts
+++ b/src/main/menu.test.ts
@@ -21,18 +21,26 @@ beforeEach(() => {
 })
 
 describe('installAppMenu', () => {
-  it('clears the application menu on win32 without dev overrides', () => {
-    installAppMenu('win32')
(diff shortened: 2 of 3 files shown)


MaanilVerma:
@deepme987 @Kosinkadink 

Whoever reviews this, please verify it on Windows.

deepme987 (approved):

PR #961, opened 2026-06-06
feat(settings): add "Hide Cloud from Instance Picker" opt-out

deepme987:
## Summary

Day-3 launch feedback has a clear drumbeat from local-only users asking for a way to make the Cloud tile go away (Reddit, Typeform, Discord). Cloud stays a first-class peer of local installs — this is opt-out, not removal.

Default: **Cloud visible** (no behavior change for the 99%). Toggle: **Global Settings → General → "Hide Cloud from the Instance Picker"**.

## Changes

| File | What |
|---|---|
| `src/main/settings.ts` | New `hideCloudFromPicker?: boolean` in `KnownSettings`, schema entry |
| `src/main/lib/ipc/registerSettingsHandlers.ts` | Boolean toggle in the General settings section, next to `autoInstallUpdates` |
| `src/renderer/src/composables/useInstallList.ts` | `useInstallList` reads the setting once on mount and ANDs it into `showCloudCard`. Failure resolves to `false` so Cloud is never silently hidden on a missing IPC bridge |
| `locales/en.json` | New `settings.hideCloudFromPicker` + description strings |

## Behavior

- **Default (off)**: Cloud tile renders exactly like today — both the Try-Cloud CTA tile and the per-install cloud tile when a cloud install exists.
- **On**: `showCloudCard` always returns `false`. Both the CTA and the per-install tile disappear from the chooser. Cloud webview / login flow / IPP routing are untouched — re-enabling the toggle restores everything immediately on next dashboard render.
- **No live broadcast**: the setting is fetched once per dashboard mount. Toggling in Global Settings takes effect on the next dashboard load (matches the snapshot-update pipeline pattern already in use).

## Why we're shipping this

From the Day-3 Typeform raw quotes (paraphrasing — sample):

> "Fuck Comfy Cloud / Dashboard is a nuisance / cannot be disabled"
> "When I close the program, I want it to CLOSE, not return to Dashboard"
> "stop forcing cloud version, nothing works today after update"

These users aren't going to ever click Cloud. Giving them a settings toggle is cheap, keeps Cloud first-class for everyone else, and removes the loudest source of "Comfy is becoming SaaS" sentiment.

## Test plan

- [x] `pnpm typecheck:web` passes
- [x] `pnpm vitest run src/renderer/src/composables/useInstallList.test.ts` passes (28/28)
- [ ] Manual: toggle the new setting in Global Settings → close + reopen dashboard → Cloud tile gone
- [ ] Manual: toggle off → reopen dashboard → Cloud tile back, no other behavior change

(Node-side typecheck has a pre-existing `Cannot find module 'yaml'` error in `desktopAdopt.ts` that's unrelated to this change.)

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -397,6 +397,8 @@
     "closeQuitDownloads": "Downloading",
     "autoUpdate": "Check for updates on startup",
     "autoInstallUpdates": "Automatically install Desktop updates",
+    "hideCloudFromPicker": "Hide Cloud from the Instance Picker",
+    "hideCloudFromPickerDescription": "For local-only users — hides the Cloud tile (and the Try Cloud CTA) from the Dashboard. Cloud remains fully functional; you can re-enable this any time.",
     "checkForUpdates": "Check for updates",
     "checkingForUpdates": "Checking…",
     "telemetry": "Telemetry",
--- a/src/main/lib/ipc/registerSettingsHandlers.ts
+++ b/src/main/lib/ipc/registerSettingsHandlers.ts
@@ -55,6 +55,16 @@ export function buildSettingsSections(): SettingsSection[] {
           type: 'boolean',
           value: s.autoInstallUpdates !== false
         },
+        // Local-only users have asked us to make the Cloud tile go away.
+        // Cloud is still a first-class peer; this is an opt-out, not a
+        // removal — default off, surface as a plain toggle.
+        {
+          id: 'hideCloudFromPicker',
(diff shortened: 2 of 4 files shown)


deepme987:
Bundled into #964 (`feat(launch-week): bundle close-modal + restore-last + always-on terminal + hide-cloud`), which merged the `hideCloudFromPicker` setting + the renderer-side live opt-out into main. Closing as superseded.

PR #973, opened 2026-06-07
Refactor/926 state driven instance navigation

### LINKED ISSUE #911: Implement: Add button to open new instance in a new window
## Overview

Add a button to the instance manager modal that allows users to open a new instance in a new window.

## Current Behavior

Currently, users can create a new instance via the "New Instance" button at the bottom left, but there is no convenient way to directly open a new instance in a new window from this modal. The "new window" button (highlighted in red in the screenshot) is visible in the UI.

## Requested Feature

Add functionality to the "new window" button so users can easily open a new instance in a separate window directly from the instance manager.

Source: #temp-desktop-bug-dump thread

---
*Migrated from Linear: [DESK2-115](https://linear.app/comfyorg/issue/DESK2-115/implement-add-button-to-open-new-instance-in-a-new-window)*

### LINKED ISSUE #926: [Refactor] Tech Debt
Few key architectures to review:
- Global single source of truth for "Instance State"
- Single flow for "instance action" [update, copy, copy-update, snapshot, restore, etc]
- Functional logic on "if in [Dashboard | Instance | Cloud], and clicked [same | new] [Dashboard | Instance | Cloud] ; then do [same_window | new_window] [switch | restart | open-new] [Dashboard | Instance | Cloud]

MaanilVerma:
## Summary

Refactors instance/window navigation (#926) from logic scattered across three sites into one pure, table-driven decision function. Given the current view (Dashboard | Instance | Cloud) and a clicked target, `decideNavigation` returns whether to switch in place, restart, focus, or open a new window — and drives the footer CTA label, a new caret split-button, and an in-drawer confirm from that single source. Ships the CTO matrix's behavior changes (Instance→Instance 3-way, cloud/instance/remote always-new-window) and adds Remote handling the matrix never specified.

## Changes

**What**

- `src/shared/navigation/navDecision.ts` (new) — `decideNavigation(NavInput): NavDecision`, a total `ReadonlyMap` transition table over `ViewKind × TargetKind × TargetRun`. Pure, O(1), runs identically in renderer + main.
- `src/shared/viewKind.ts` (new) — single `ViewKind`/`Category`/`NavClass` vocabulary + `navClass` (folds `remote ⇒ cloud`) + `viewKindFor` (one classifier shared by `computeViewKind` and the snapshot builder).
- `src/renderer/src/composables/useInstanceNavState.ts` (new) — read-model deriving `NavInput` facts; reuses `useInstallCta` so the one-install-one-window invariant stays single-sourced. A **remote target routes via the cloud cells** (non-local URL backend).
- `src/renderer/src/composables/useInstanceActions.ts` (new) — single dispatcher routing a decision's verb onto the bridge, with cloud-capacity + local-kill gates; aborts (no unhandled rejection) if a confirm dialog rejects.
- `src/main/index.ts` — extracts `deliverPickToEntry` (shared by swap + new-window), adds the `openInstallInNewWindow` primitive (focus-existing else spawn a fresh chooser host), and threads a `confirmed` flag through `pickInstallFromPicker` so the renderer's in-drawer confirm replaces main's system modal.
- `src/main/popups/titlePopup.ts` — adds the `comfy-titlepopup:open-install-new-window` IPC + `currentView`/`currentCategory` on the picker snapshot; picker IPC wirings wrapped in `.catch()`.
- `src/renderer/src/components/settings/ComfyUISettingsContent.vue` — footer CTA + caret split-button driven by `decideNavigation`; leading icons on the CTA and both menus; per-category CTA wording/icon (`Open Cloud` vs `Open Remote`); reuses the existing `stop` action in the caret next to Restart.
- `src/renderer/src/views/comfyUISettings/MoreMenu.vue` — optional `heading` + per-item `icon` (the plain "More" menu is unchanged).
- `docs/instance-navigation-926-handoff.md` (new) — verbatim CTO matrix + Implemented? column, Remote rows (§1b), deviations (§1a), manual-verification + unit-vs-e2e coverage tables.

**Breaking**

- None. IPC channels are additive; the `primary-action` emit payload changed `boolean → NavDecision` but is internal to the picker. CTA labels for unchanged cells are byte-identical; existing telemetry (`instance.switched`, `instance.opened_new_window`) is preserved.

## Review Focus

- **The decision table** (`navDecision.ts`) is the whole behavior surface — each matrix cell is one entry. Review it as data against the CTO matrix in the handoff doc. A reachability test bans any no-op CTA for a reachable (host, target, run) combo.
- **Remote == Cloud (target):** a remote connection is a non-local URL backend, so it routes through the cloud cells; the component resolves wording/icon per raw category. This fixes the dead "Start"/wrong-"Switch" cases on remote targets.
- **Deliberate deviations (§1a/§1b):** Cloud→Dashboard ships new-window (the chip isn't table-driven yet); Cloud/Remote → self resolves to **Restart** instead of matrix row 16's "second window" (a true second view of one session isn't supported; the `allowDuplicate` plumbing is kept dormant/reserved).
- **Scope boundaries:** pillar-2 lifecycle-action consolidation (`useInstallContextMenu` thinning) is a deferred follow-up. The 3-way confirm *rendering* and focus-existing-for-a-real-window are unit/manual-only (e2e can't spawn a live ComfyUI process) — see the coverage table.

## Testing

State-driven navigation splits into a pure decision (unit-testable exhaustively) and side-effecting wiring (e2e via real bridge → IPC → window): unit pins *what* the decision is for every cell; e2e pins *that the wiring fires* for the risky paths. Un-e2e-able cells are documented with their reason in the handoff doc.

- `navDecision.test.ts` — every matrix cell + the full `ViewKind × TargetKind × TargetRun × class × intent` cross-product (totality), caret/new-window selection, and the **no-dead-no-op-CTA** reachability guard.
- `useInstanceActions.test.ts` — verb→bridge routing for all verbs; cloud-capacity + switch/restart gates; the 3-way (switch / open-new / cancel) outcomes; dispatch-reject and bridge-undefined guards; dormant `allowDuplicate` passthrough.
- `useInstanceNavState.test.ts` / `registry.test.ts` — run-state derivation; remote routes like cloud (incl. Instance→Remote ≡ Instance→Cloud); `computeViewKind` (null/unknown category).
- `e2e/nav-matrix-{dashboard,instance,cloud}.test.ts` — 8 tests asserting via recorded IPC + live `BrowserWindow` counts: same-window launch, focus-existing, new-window spawn (parent undisturbed), `allowDuplicate` primitive.

`pnpm run typecheck` (node/web/e2e/integration) + `pnpm run lint` clean; **1983 unit tests pass**; e2e `--project=lifecycle` green (`pnpm run build` first — e2e runs the built bundle).

### Manual verification matrix

| # | Current view | Click target | Target state | Proposed action | Primary CTA | Secondary | Implemented? |
|---|---|---|---|---|---|---|---|
| 1 | Dashboard | Dashboard | n/a | No-op | – | – | ✅ |
| 2 | Dashboard | Instance A | Not running | Launch + same window | Start | – | ✅ |
| 3 | Dashboard | Instance A | Running | Focus existing | Switch | – | ✅ |
| 4 | Dashboard | Cloud | Not running | Open same window | Open Cloud | Open in new window | ✅ |
| 5 | Dashboard | Cloud | Already open | Focus existing | Switch | Open in new window | ✅ |
| 6 | Dashboard | + New Instance | n/a | Install wizard | New Install | – | ✅ |
| 7 | Instance A | Dashboard | n/a | New window (A keeps running) | Open Dashboard | – | ✅ |
| 8 | Instance A | Instance A | Self | Restart in place | Restart | Stop | ✅ |
| 9 | Instance A | Instance B | Not running | Switch (in-drawer 3-way) | Switch | Open in new window → Switch / Open in new window / Cancel | ✅ |
| 10 | Instance A | Instance B | Running | Focus existing | Switch | – | ✅ |
| 11 | Instance A | Cloud | Not running | New window (A keeps running) | Open Cloud | – | ✅ |
| 12a | Instance A | Cloud | Already open | Focus existing | Switch | – | ✅ |
| 12b | Instance A | Remote | Already open | Focus existing | Switch | Stop | ✅ |
| 13 | Cloud | Dashboard | n/a | Switch same window | Open Dashboard | – | ⚠️ ships new-window (§1a) |
| 14 | Cloud | Instance A | Not running | New window (cloud keeps running) | Open in new window | – | ✅ |
| 15 | Cloud | Instance A | Already open | Focus existing | Switch | (Restart — deferred) | ✅ |
| 16 | Cloud | Cloud | Self | Restart in place | Restart | – | ⚠️ self → Restart, not second window (§1b) |
| 17 | Cloud | + New Instance | n/a | Install wizard in new window | New Install | – | ✅ |

### Remote rows (not in the original CTO matrix — Remote ≡ Cloud target)

| # | Current view | Click target | Target state | Behavior | Primary CTA | Implemented? |
|---|---|---|---|---|---|---|
| R1 | Dashboard | Remote | Not running | Open same window (= row 4) | Open Remote | ✅ |
| R2 | Dashboard | Remote | Already open | Focus existing (= row 5) | Switch | ✅ |
| R3 | Instance A | Remote | Not running | New window, A keeps running (= row 11) | Open Remote | ✅ |
| R4 | Cloud / Remote | different Cloud/Remote | Not running | New window, host keeps running | Open Remote / Open Cloud | ✅ |
| R5a | Any | Cloud | Running elsewhere | Focus existing | Switch | – | ✅ |
| R5b | Any | Remote | Running elsewhere | Focus existing | Switch | Stop | ✅ |
| R6 | Cloud / Remote | self | — | Restart in place (second view unsupported) | Restart | ✅ |

Closes #926 closes #911


deepme987:
## Review: backend + frontend pass

Reviewed the full changeset (25 files, +2557/−204) across two lenses — shared decision logic / main process / IPC (backend) and renderer composables / components / i18n (frontend). **No blocking issues.**

The refactor is well-structured: `decideNavigation` is genuinely pure and total (a missing key falls through to a frozen `NO_OP`, so no undefined-deref), the `TABLE` is frozen against shared-singleton mutation, the new title-popup IPC handlers are properly hardened (sender `entry.kind === instancePicker` guard, `installationId` string validation, `allowDuplicate === true` coercion, caught rejections), lifecycle teardown is clean (`ResizeObserver` / countdown interval / document listeners all in `onUnmounted`), and the dispatcher's abort path can't leak an unhandled rejection. The e2e negative assertions are done right (they sample the full settle window rather than `expect.poll().toBe(0)`, no `waitForTimeout`).

Findings below are all **medium and under**. The headline ones (missing i18n keys, dead `telemetry` field, unreachable table cells, dead `running-here` branch) were spot-verified against the code; the dead branch was independently flagged by both lenses.

| # | Sev | Lens | Location | Issue | Suggested fix |
|---|-----|------|----------|-------|---------------|
| 1 | medium | FE | `locales/{en,zh}.json` · `ComfyUISettingsContent.vue:556,1115,1124` | `instancePicker.openRemote`, `windowOptions`, `moreWindowOptions` are referenced at runtime but exist in **neither** locale — they survive only on the inline `t(key, 'fallback')` default, so every non-English locale renders English for the remote CTA + caret split-button. The PR's i18n guard only iterates `NAV_LABEL`, so it won't catch this. | Add the 3 keys to `en.json` + `zh.json`; extend the i18n test beyond `NAV_LABEL`. |
| 2 | medium | BE | `navDecision.ts` | `NavDecision.telemetry` is populated in 4 cells and snapshot-asserted, but **nothing reads it** — real telemetry is emitted imperatively in `main/index.ts`. The column looks authoritative but isn't, and can silently diverge (main emits `opened_new_window` only on the spawn branch; the cell marks the whole transition). | Either dispatch telemetry from `decision.telemetry` (make the table the real source) or delete the field + its assertions. |
| 3 | medium | BE | `navDecision.ts:126-265` | 6 cells (all `dashboard` / `new-instance` targets) are unreachable from the only production caller (`navInput` never yields those targets) yet are snapshot-tested as if live. The header says "edit the cell to change behavior" — false for these rows. | Mark them documentation-only in code + annotate the tests (as the `*->dashboard` cells already are). |
| 4 | low | BE | `InstancePickerView.vue` · `main/index.ts:1602` | `confirmSwitch` / `shouldConfirmKillForEntry` gate the "this stops the current instance" prompt on the host being a **local install**, not on it actually **running** — a stopped local host still triggers the warning. Internally consistent, but the "the current local instance is running" comments overstate the precondition. | Fix the comments to say "local host", or gate on an actual running-session check. |
| 5 | low | BE | `main/index.ts` (`openInstallInNewWindow`) | Takes `installationId` from the renderer and, with no existing window, spawns a fresh chooser host + deep-link **without verifying the install exists** — a stale id leaves a stray empty window (not a security issue; the id is only a map key). | Guard with a `getInstallation(id)` existence check before spawning (mirror the `openStartupSurface` raced-delete fallback). |
| 6 | low | FE | `navDecision.ts:82-85` | `NavInput.currentClass` / `targetClass` are populated on every input but `decideNavigation` never keys on them (the `remote ⇒ cloud` fold is already applied upstream into view/target). Dead decision inputs that invite a reader to assume they matter. | Drop them from `NavInput`, or note inline they're carried for diagnostics only. |
| 7 | low | BE + FE | `useInstanceNavState.ts:76-81` | The `running-here` branch is unreachable: `isTargetRunningHere` (`cta.runningInThisWindow`) is true only when `target.id === activeInstallationId` — the exact condition as `isTargetCurrentHost` — so the earlier `self` branch always returns first. (Flagged by both lenses.) | Drop the branch (+ the `TargetRun` member if unused), or comment that it collapses into `self`. |
| 8 | low | BE | `navDecision.test.ts:216` | `en as unknown as Record<…>` is a double assertion — the `as unknown as` pattern the repo rules forbid alongside `as any`. The JSON import is already structurally compatible. | Single assertion `en as Record<…>`, or type the import. |
| 9 | nit | BE | `ComfyUISettingsContent.vue:616-695` | `caretActions` keys menu items by `alt.primaryLabel` (the i18n key); if a cell ever carried two secondaries sharing a label, `find()` makes the second unselectable. Latent (every cell has ≤1 secondary today). | Key by index / a synthetic id. |
| 10 | nit | FE | `ComfyUISettingsContent.vue:156,543` | `useInstallCta` is instantiated twice for the same install (once directly, once inside `useInstanceNavState`) — two computed graphs deriving identical run-state. Correct + cheap, but duplicates the thing the refactor set out to centralize. | Have `useInstanceNavState` accept/expose the `InstallCta` so the component reuses one. |

**Net:** ship-ready after the i18n keys (#1) and a decision on the dead `telemetry` field (#2). The rest is cleanup that can land here or in a follow-up. Nice work on the totality test and the IPC hardening.

Kosinkadink:
## Review notes (non-blocking) — from a post-merge review

Nice refactor — the single table-driven `decideNavigation` is a real improvement and the unit coverage is excellent. None of the below blocks merge, but they're worth addressing here or in a fast follow-up.

### 1. Cold/stale picker cache can skip the local-switch confirm (highest priority)
`buildInstancePickerSnapshot` derives the host's `currentView`/`currentCategory` by looking the active id up in `args.installs` (`src/main/popups/titlePopup.ts` ~L333-L336). The initial snapshot is built from the cached install list and only refreshed after open. If that cache is empty/stale when the picker opens on a **local instance** host, the active install isn't found → `currentCategory: null` → `currentView: 'cloud'`. The renderer then treats the host as non-local in `confirmSwitch` and returns `'switch'` **without** the in-drawer 3-way confirm, and because it sends `confirmed: true`, main also skips its system-modal fallback (`src/main/index.ts` ~L1582). Net effect: a local ComfyUI process can be stopped with no prompt.

Suggested fix: source the host category/view from `parentEntry.sourceCategory` / `parentEntry.installationId` directly rather than the install-list payload (keep the list lookup only as a fallback). A cold-cache unit test (local host id, empty `installs` ⇒ `currentView: 'instance'`, `currentCategory: 'local'`) would lock it in.

### 2. `focus` cells carry a dead `secondary` (table-purity)
`dashboard|cloud|running-elsewhere` and `instance|cloud|running-elsewhere` are `verb: 'focus'` but still include `OPEN_NEW_WINDOW_SECONDARY` (`src/shared/navigation/navDecision.ts`). The component hard-overrides the caret for any `focus` verb (`ComfyUISettingsContent.vue` ~L613-L615), so it's not user-visible — but it means `decideNavigation({ intent: 'new-window' })` returns `open-new` for an already-running target, which dents the "table is the single source of truth" guarantee. (Note: the dashboard cell matches matrix row 5; the instance cell contradicts row 12a.) Suggest dropping those two `secondary` entries and adding an invariant test: every `focus` decision has empty `secondary`.

### 3. Cloud-capacity gate blocks a harmless focus
The footer disables the CTA whenever `sourceCategory === 'cloud'` and capacity is disabled, and `useInstanceActions.dispatch` runs the capacity gate before checking the verb (`useInstanceActions.ts` ~L54). For an already-running cloud target the decision is `focus`, so this blocks focus-existing — whereas `ChooserView` focuses running windows *before* applying the capacity gate. Suggest skipping the block/gate when `decision.verb === 'focus'`.

### 4. Telemetry: `instance.opened_new_window` is new, not "preserved"
The PR description lists `instance.opened_new_window` among preserved events, but `git grep` finds it nowhere in `main` — it's introduced here (`src/main/index.ts` ~L1562). Mechanically harmless (event names are free-form; consent-gated + rate-limited, no name allowlist), but: it won't appear in existing dashboards until it's registered server-side, and the description wording should be corrected. `instance.switched` is genuinely preserved and fires correctly (the new-window choice returns before that emit, so no false swap events).

### 5. Telemetry: focus-existing emits nothing
`opened_new_window` fires only on the spawn path, after the `willFocusExisting` early-return (`src/main/index.ts` ~L1542). So "Open in new window" that focuses an already-open window emits no telemetry — a blind spot if you want to measure that interaction.

(Auto-updater is untouched and unaffected — flagging only because it was in scope of the question.)

Kosinkadink (approved):

Kosinkadink:
@deepme987 can you please accept the CLA 😭 otherwise i will have to bypass repo reqs to merge

MaanilVerma:
> @deepme987 can you please accept the CLA 😭 otherwise i will have to bypass repo reqs to merge

He said "I have accepted 40-50 times already, it still asks me for the same, not sure if it works" you might need to bypass I think for now, obviously if we can.

PR #974, opened 2026-06-07
fix(standalone): force pip-sync on first auto-update to cure ModelMMAP/aimdo drift

deepme987:
## What
On a fresh standalone install, force `uv pip install --upgrade -r requirements.txt` to run after the post-install auto-update, even when the requirements.txt content is byte-identical pre/post.

Also tightens the existing update flow: re-sync deps when git HEAD moves (not just when requirements.txt content changes), and add `--upgrade` to all three pip-install paths so uv's satisfaction check can't no-op against a stale lock entry.

## Why
Field reports of:

- \`AttributeError: 'ModelMMAP' object has no attribute 'get_file_handle'\` (Comfy-Org/ComfyUI#14227, #14295)
- \`ModuleNotFoundError: No module named 'comfy_aimdo.vram_buffer'\` (Discord report attached above)

trace back to the standalone install bundle's pre-extracted venv shipping \`comfy-aimdo\` at a version older than what ComfyUI's source imports. ComfyUI logs a \`WARNING: Installed comfy-aimdo version X is lower than recommended Y\` and keeps going, so the user only finds out when their first workflow crashes (usually a VAE or checkpoint load).

The post-install auto-update was running (\`install.ts:168\`) but the pip sync was gated on \`preReqs !== postReqs\`. When ComfyUI hadn't changed its \`requirements.txt\` between the bundle freeze and the first git pull, that gate was false → pip install skipped → bundled aimdo stayed lagged forever.

Two related gaps fixed alongside:

1. **HEAD-movement was ignored.** Source-only bumps (new \`comfy_aimdo.*\` imports without a requirements.txt change — exactly what \`model_management.py\` did between aimdo \`0.4.x\` releases) slipped through without a re-sync.
2. **No \`--upgrade\` flag.** uv's satisfaction check could no-op against a lock-pinned version even when \`requirements.txt\` had drifted.

## How
- \`installFilteredRequirements\` takes an \`upgrade\` flag → adds \`--upgrade\` to the uv pip command.
- \`runComfyUIUpdate\` takes a \`forceDepsSync\` flag; trigger pip sync when reqs changed OR HEAD moved OR caller forces. All three pip-install paths (dry-run, real, post-install fallback) pass \`--upgrade\`.
- Post-install auto-update in \`install.ts\` sets \`forceDepsSync: true\`.

## Safety
- PyTorch protection (\`PYTORCH_RE\` filter) is preserved — \`--upgrade\` only applies to the filtered set, so \`torch\` / \`torchvision\` / \`torchaudio\` / \`torchsde\` are never touched.
- Existing updates that legitimately don't move HEAD or change reqs (i.e., already up to date) still skip the pip step — no UX regression on no-op updates.
- All 70 existing standalone + pip tests pass; no test changes required (the change is gate-widening, not gate-removal).

## Test
- [ ] Fresh install on a system without a global ComfyUI; verify the post-install auto-update step now runs \`uv pip install --upgrade -r requirements.txt\` even when ComfyUI master hasn't bumped reqs
- [ ] Verify the first workflow load (VAE or CheckpointLoaderSimple) doesn't throw \`ModelMMAP\` errors
- [ ] On an existing install with a known-lagged \`comfy-aimdo\`, trigger Update → confirm aimdo gets upgraded
- [ ] Confirm no-op updates (already up to date, no reqs/HEAD change, no force flag) still skip pip — no regression


deepme987:
Combined with #975 into #976 to keep the snapshot-import + deps-sync fixes under a single review (both surface during the same restore-from-snapshot workflow). Closing as superseded.

PR #975, opened 2026-06-07
fix(snapshot-import): pass includeLatestStable so Create Installation enables

deepme987:
## What
Four call sites of `standaloneSource.getFieldOptions('release', ...)` were missing `{ includeLatestStable: true }` in the context arg. With the flag absent, the standalone source returns an empty release list silently — every downstream UI / flow then either disables a button or bails with "No releases available."

Fixed call sites:

- `src/renderer/src/views/LoadSnapshotModal.vue` — Create Installation button now enables after a snapshot is loaded
- `src/renderer/src/composables/useMigrateAction.ts` — migrate flow auto-picks release/variant correctly
- `src/main/lib/ipc/registerSnapshotHandlers.ts` (`handleCreateFromSnapshot`) — server-side create no longer falsely reports "No releases available."
- `src/main/lib/standaloneMigration.ts` — legacy-desktop migration progresses past release selection

## Why
The standalone source's release branch gates the entire option list on the flag (`standalone/index.ts:328`):

```ts
if (tags.length > 0 && context?.includeLatestStable) {
  options.push({ value: 'stable', recommended: true, ... })
  options.push({ value: 'latest', ... })
}
return options
```

Field report (etheruk via Discord): "Load Snapshot shows all the stuff but Create Installation is disabled." Snapshot file imports fine, preview renders, but the button never activates because `selectedVariant` stays `null` (no releases → no variant fetch → no variant pick).

`InstallWizardModal` and `QuickInstallModal` already pass `{ includeLatestStable: true }` — this PR just mirrors that.

## Test
- [ ] Open Comfy Desktop dashboard
- [ ] Click "Load snapshot" / drag a snapshot file
- [ ] Verify release dropdown auto-populates with "Stable" + "Latest"
- [ ] Verify variant cards render and one is auto-selected
- [ ] Verify Create Installation button enables
- [ ] Click Create → install proceeds without "No releases available." error
- [ ] (Bonus) legacy-desktop migration: trigger from a legacy install → confirm it picks a release/variant instead of falling into the "no auto-pick" branch


deepme987:
Combined with #974 into #976 to keep the snapshot-import + deps-sync fixes under a single review (both surface during the same restore-from-snapshot workflow). Closing as superseded.

PR #997, opened 2026-06-09
feat(close): always confirm closing local install windows; remove last-window quit toggle

Kosinkadink:
## What

Closing a **local install instance window now always shows the Close Window / Cancel confirm**, including the last window. Confirming the last window tears it down and quits Comfy Desktop via `window-all-closed`.

The friction is intentional: it prevents accidental closes of a ComfyUI that can take a minute or more to launch.

Returning to the dashboard is **no longer a close-time choice** - the three-way *Quit Desktop / Return to Dashboard / Cancel* modal is gone. Going back to the dashboard remains a deliberate user action via the title pill's **Open Dashboard** / **New Window**.

## Why

The previous model bypassed the confirm on the last window and instead consulted a `closeDirectlyOnLastWindow` setting (default off ? silently return to dashboard). That setting was confusing and, combined with the return-to-dashboard default, made the accidental-close protection inconsistent. Simplifying to "always confirm, close = quit" is clearer and makes `reopenLastInstanceOnLaunch` (auto-restart) actually meaningful: closing the last window now quits with the instance as the active surface, so the next launch boots straight back into it.

## Changes

- `confirmCloseInstanceWindow` (detach.ts): always renders the 2-way Close Window / Cancel modal; copy is tailored for the last window (".quits Comfy Desktop") and for whether a local session is stopped. `CloseWindowChoice` narrowed to `'close' | 'cancel'`.
- Close handler (createHostWindow.ts): the last install window now goes through the same confirm as any other; removed the `return-to-dashboard` branch and the `shouldDetachLastInstallWindowToDashboard` fallback (+ helper + tests).
- Removed the `closeDirectlyOnLastWindow` setting: schema (`settings.ts`), Global Settings field (`registerSettingsHandlers.ts`), and `en.json` strings. The legacy key is cleaned from existing `settings.json` on load.
- `confirmAndCloseHostWindow` (menu path): dropped the `return-to-dashboard` handling.

Deliberate Return-to-Dashboard flows (title pill, File menu, IPC) are untouched.

## Notes

- The 3-button native modal helper `openSystemModalChoiceAsync` is now unused but left in place - it's general cross-process (main + preload + renderer) infrastructure; removing the whole path is out of scope here.

## Verification

- `pnpm run typecheck` ?
- `pnpm run lint` ?
- `pnpm run build` ?
- `pnpm run test` ? (1931 tests)

(Repo-wide `format:check` fails on pre-existing untouched files - e.g. `tsconfig.json`, `package.json` - so it was not run as a gate; it is not part of the husky pre-commit hook.)

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -406,8 +406,6 @@
     "autoInstallUpdates": "Automatically install Desktop updates",
     "appBehavior": "App Behavior",
     "reopenLastInstanceOnLaunch": "Auto-restart last instance on launch",
-    "closeDirectlyOnLastWindow": "Close Comfy Desktop on last window (don't return to Dashboard)",
-    "closeDirectlyOnLastWindowDescription": "When you close the last ComfyUI instance window, quit Comfy Desktop immediately instead of returning to the Dashboard.",
     "hideCloudFromPicker": "Hide Cloud from the Instance Picker",
     "hideCloudFromPickerDescription": "For local-only users — hides the Cloud tile (and the Try Cloud CTA) from the Dashboard. Cloud remains fully functional; you can re-enable this any time.",
     "checkForUpdates": "Check for updates",
--- a/locales/en.json
+++ b/locales/en.json
@@ -407,7 +407,7 @@
     "appBehavior": "App Behavior",
     "reopenLastInstanceOnLaunch": "Auto-restart last instance on launch",
     "confirmBeforeClosingWindow": "Confirm before closing a window",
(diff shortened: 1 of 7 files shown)


deepme987:
> Closing a local install instance window now always shows the Close Window / Cancel confirm, including the last window. Confirming the last window tears it down and quits Comfy Desktop via window-all-closed.

we had multiple complaints for this - they don't like the extra friction; hence the toggle in Desktop Settings

Kosinkadink:
Cool, I'll just remove these modals from popping up in general in that case.

deepme987 (approved):

PR #998, opened 2026-06-09
fix: keep frozen ComfyUI installs frozen on Load Snapshot and Track Existing (#986)

### LINKED ISSUE #986: [Bug] Track existing / Load snapshot should NOT update ComfyUI automatically
defeats the whole "frozen comfy" purpose.

Kosinkadink:
## Summary

The "frozen comfy" promise — that Load Snapshot and Track Existing recreate a ComfyUI environment **as it was**, without silently rolling it forward — regressed on 2026-06-03. This PR restores that behavior end to end.

A user reported (#986) that **Load Snapshot** and **Track Existing / adopt** were auto-updating ComfyUI to the latest version, defeating the whole point of a frozen install.

## Root cause

Two changes on 2026-06-03 broke the frozen flows:

- **#832** reworked the standalone release dropdown to expose only the `stable` / `latest` channels. The snapshot-load path still tried to match the snapshot's per-version `releaseTag` against those channel ids, which can never match, so it silently fell back to `stable` and `buildInstallation` set `autoUpdateComfyUI: true` — auto-updating before the snapshot restore could re-pin.
- **#669** added a forced `updateComfyToStable()` checkout during adoption, rolling the user's existing ComfyUI source forward to latest stable on every Track Existing.

## Changes

**1. Snapshot-created installs stay frozen**
- New shared `frozenSnapshotInstallOverrides()` (in `snapshots/restore.ts`) forces `autoUpdateComfyUI: false` and mirrors the snapshot's `updateChannel` as the manual update preference. Applied in the create-from-snapshot handler and the standalone migration path.
- `LoadSnapshotModal` now preselects the release channel by matching the snapshot's `updateChannel` instead of an impossible per-version `releaseTag`.

**2. Load Snapshot reinstalls the exact standalone bundle**
- New `buildPinnedVariant()` resolves the snapshot's original R2 bundle tag (`comfyui.releaseTag`) from the vendor's `releases.json` history and installs that exact environment (matching the original bundled Python/torch), rather than the newest bundle for the channel.
- Falls back to the newest bundle when the tag has been pruned from R2.
- Extracted a shared `buildVariantOption()` helper so the install-wizard list and the snapshot-load path build variant cards identically.

**3. Track Existing / adopt no longer auto-updates ComfyUI**
- Removed the forced `updateComfyToStable()` checkout and its progress step; the user's existing checkout is preserved as-is. ComfyUI updates remain opt-in (`autoUpdateComfyUI: false`) and triggerable from the Update tab.
- Version resolution now falls back to the adopted source's own `comfyui_version.py`.

## Testing

- `pnpm typecheck`, `pnpm lint`, `pnpm build` — all pass.
- `pnpm test` — 1990 passing, including new tests for `frozenSnapshotInstallOverrides`, `buildPinnedVariant` (exact pin, install freezing, pruned-tag fallback, missing-vendor fallback), and the updated adoption tests asserting no auto-update.

Closes #986


Kosinkadink:
Out-of-scope follow-up work is tracked in #1006 (snapshot-install slowness: reuse the modal-resolved release/variant instead of re-fetching R2 on Install, fixing the >1 min freeze + selection drift; plus detectGPU memoization and a etchJSON timeout). This PR is scoped to keeping frozen installs frozen and is ready to merge.

deepme987 (approved):

PR #1016, opened 2026-06-09
Clean up unique browser partitions on delete + drive-aware data defaults

### LINKED ISSUE #1017: Disk usage: orphaned browser partitions on delete + data dirs always default to system drive
Two related disk/storage issues in the launcher (Desktop 2.0).

## 1. Unique browser partitions are orphaned forever on delete

Deleting an install only removes the install directory and the `installations.json` record. Installs with `browserPartition: 'unique'` each own a `persist:<id>` bucket under `%APPDATA%\Roaming\comfyui-desktop-2\Partitions\<id>` (cookies, IndexedDB, localStorage, service-worker cache, GPUCache). Because the id is unique per-install and never reused, every create-unique -> delete cycle strands one of these folders permanently. Nothing ever reclaims them.

(Copied installs are unaffected from a safety standpoint: each copy gets a fresh id and its own deep-copied partition, so cleanup of one never touches a sibling.)

## 2. Large data dirs always default to the system drive

When installing the app, the user can choose an install location (e.g. `D:`), but the multi-GB data directories still default to the system drive:

- install root -> `%USERPROFILE%\ComfyUI-Installs`
- shared models/input/output -> `%USERPROFILE%\ComfyUI-Shared`
- download cache -> `%APPDATA%\Roaming\comfyui-desktop-2\download-cache`

So picking `D:` for the app still fills up `C:` with installs, models, and the download cache. The defaults should follow the drive the user picked, and ideally stay grouped under a single parent folder on that drive rather than scattering folders at the drive root.

## Fix

Addressed in #1016:
- Clean up a unique install's browser partition on delete (bounded, best-effort, never blocks the delete).
- On Windows, when the app is installed to a non-system drive, default `installDir`, the shared models/input/output root, and the download cache to a single `<drive>\Comfy-Desktop` folder. Non-Windows / same-drive installs keep the existing home/userData defaults; the small Electron-managed `userData` state stays on the system drive.

Kosinkadink:
Closes #1017

## Summary

Two related disk/storage fixes for the launcher.

### 1. Browser partitions no longer leak on delete

Deleting an install previously only removed the install dir and the `installations.json` record. Unique-partition installs (`browserPartition: 'unique'`) each own a `persist:<id>` bucket under `userData/Partitions/<id>` (cookies, IndexedDB, localStorage, service-worker cache, GPUCache). The id is unique per-install and never reused, so every create-unique -> delete cycle stranded one of these folders forever. (Copies were already safe: each copy gets a fresh id and its own deep-copied partition.)

Cleanup now happens at two layers:

- **Inline, on delete** (`handleDelete`, both delete paths): the install record is removed first (so the user-visible delete is deterministic), then the partition is cleaned best-effort — clear the session (`clearStorageData`, raced against a 5s timeout so it can never hang the operation or hold its lock) and remove `Partitions/<id>` (`force` + transient-lock retries). This also fires when an install was created `unique` but later toggled to `shared` (the per-install dir is removed whenever it exists, not just when the current setting reads `unique`), so a toggled install's partition can't leak.
- **Startup orphan sweep** (`sweepOrphanPartitions`): on Windows the inline removal often loses a race — `clearStorageData` re-materializes a live, locked session dir, so `Partitions/<id>` survives the delete. At startup no install session is live yet, so any `Partitions/inst-*` whose id is not a current install is removed, reliably reclaiming both delete leftovers and crash leftovers.

Both layers only ever touch install-id-shaped (`inst-*`) dirs; they never touch `persist:shared` (`Partitions/shared`, the bucket all shared-partition installs collectively own) or any other session dir. Worst case degrades to one folder briefly left behind until the next launch — never a lock-up, never lost shared data.

### 2. Large data dirs default to a sensible, grouped location

Previously the multi-GB consumers defaulted to the system drive even when the app was installed elsewhere, and on the common system-drive install they landed in poor spots: the install root and shared models/input/output under the **home root**, and the download cache under **roaming AppData** (which bloats roaming/AD profiles on login sync). These are still only *defaults* - users can override each in Settings.

**Non-system-drive installs (e.g. app installed to `D:`)** now group everything under a single `<drive>\Comfy-Desktop` parent instead of scattering at the drive root:
- `<drive>\Comfy-Desktop\ComfyUI-Installs`
- `<drive>\Comfy-Desktop\ComfyUI-Shared` (models / input / output)
- `<drive>\Comfy-Desktop\ComfyUI-Cache\download-cache`

"Non-system-drive" is keyed on the app's own exe drive vs `process.env.SystemDrive` (anchored on the OS system drive, not the user profile — a redirected profile on another drive must not be misread as a redirected install).

**System-drive Windows installs** now distinguish new vs existing users:
- **New installs** group under `%LOCALAPPDATA%\Comfy-Desktop` (the standard per-user, non-roaming spot for large app data — and writable without admin, unlike a drive root):
  - `%LOCALAPPDATA%\Comfy-Desktop\ComfyUI-Installs`
  - `%LOCALAPPDATA%\Comfy-Desktop\ComfyUI-Shared` (models / input / output)
  - `%LOCALAPPDATA%\Comfy-Desktop\ComfyUI-Cache\download-cache`
- **Existing installs keep their original home-root + roaming-cache layout** — no data is moved or migrated.

New vs existing is classified from a home-root footprint (`~/ComfyUI-Installs` / `~/ComfyUI-Shared`) and then **pinned with a one-time marker** (`userData/data-location.json`) so a user's chosen location can never flip on a later launch (e.g. if the legacy folders appear externally). The marker is written once at startup, never at module import. `settings.json` existence is deliberately not used as the signal (a new user creates it on first run).

**Dead-volume fallback:** `installDir` / `cacheDir` reset to the built-in default only when their whole volume/drive root is gone (e.g. a reinstall onto a different drive, or a removed disk), so the app never strands installs/cache on an inaccessible path. A custom location on a live volume that simply hasn't been created yet is preserved (installs/cache are created on demand) — only the volume root is checked, not the leaf.

Non-Windows behavior and the small Electron-managed `userData` state are unchanged (the latter intentionally stays on the system drive).

## Changes

- `src/main/lib/ipc/shared.ts` - add `deleteBrowserPartition` (timeout-bounded session clear + dir removal; also handles installs toggled unique -> shared) and `sweepOrphanPartitions` (startup reclaim of orphaned `Partitions/inst-*`).
- `src/main/lib/ipc/sessionActions/delete.ts` - call `deleteBrowserPartition` on both delete paths, after record removal.
- `src/main/lib/ipc/index.ts` - run `sweepOrphanPartitions` at startup against the current install-id set (after the empty/broken-install sweep).
- `src/main/lib/paths.ts` - add `selectedInstallDrive` (anchored on `SystemDrive`) / `defaultDataRoot` / `defaultDownloadCacheDir`; system-drive Windows classifier (`%LOCALAPPDATA%\Comfy-Desktop` for new installs, home for existing) with a pinned marker via `persistWinDataRootChoice`; `builtinDefaultInstallDir` uses the data root.
- `src/main/index.ts` - call `persistWinDataRootChoice()` once at startup.
- `src/main/settings.ts` - default `installDir`, shared root, and download cache to the data root; add the dead-volume fallback for `installDir` / `cacheDir`.

## Tests

- `delete.integration.test.ts` - unique partition removed + session cleared; shared partition untouched; install toggled unique -> shared still has its partition cleaned; delete still completes when `clearStorageData` rejects; cleanup works when the install dir is already gone; `sweepOrphanPartitions` removes only orphaned `inst-*` dirs and never `shared`.
- `paths.test.ts` - drive redirect to `<drive>\Comfy-Desktop` on win32 non-home drive; system-drive new install -> `%LOCALAPPDATA%\Comfy-Desktop`; existing install (home footprint) -> home root + roaming cache; marker overrides later footprint changes in both directions; `persistWinDataRootChoice` writes the classified mode once and never overwrites; redirected-profile/system-drive anchoring regression; non-Windows unchanged.
- `settings.test.ts` - dead-volume `installDir`/`cacheDir` fall back to default; a not-yet-created custom dir on a live volume is preserved.
- Full suite green: typecheck, lint, test (2326), integration (24), build.

> Verified end-to-end on real Windows: existing-user safety (no data moved on update), partition cleanup on delete via the startup sweep, dead-volume vs live-volume fallback, new-user `%LOCALAPPDATA%\Comfy-Desktop` layout, and non-system-drive `<drive>\Comfy-Desktop` redirect.


Kosinkadink:
Follow up - see if we can get the electron-manager userData moved as well.

deepme987 (approved):

PR #1022, opened 2026-06-09
feat: opt-in auto-launch instance on Desktop startup

deepme987:
Adds an opt-in setting for picking an instance that Comfy Desktop launches automatically on startup.

## What
New entry in Global Settings → General → App Behavior: **Auto-launch on startup** (`select` field). Options:
- **None (show dashboard)** — default. Today's behavior.
- **Last used instance** — opens the install with the largest `lastLaunchedAt`.
- One option per installation (any source — local, cloud, remote).

## Why off by default
Per request — opt-in. Most users want the dashboard. A few have a single install they always use and want it to just open.

## Wiring
- `settings.ts`: `autoLaunchOnStartup?: string`. Default-means-unset so `set('autoLaunchOnStartup', 'none')` deletes the key (no accumulating no-op writes). Sentinels `AUTO_LAUNCH_NONE` / `AUTO_LAUNCH_LAST`.
- `installations.ts`: `resolveAutoLaunchInstall(value)` — null for null/empty/'none', `getRecent()` for 'last', `get(id)` for an id (null when stale).
- `registerSettingsHandlers.ts`: `buildSettingsSections` accepts an optional `installs` list and emits the dropdown field next to `reopenLastInstanceOnLaunch`.
- `titlePopup.ts`: fast-open path opens the popup with the basic none/last snapshot instantly; the existing rebroadcast (already gated on stars-fetch) then hydrates the per-install options once `installations.list()` resolves.
- `main/index.ts`: `openStartupSurface` reads the setting first. When explicitly set, it overrides the implicit `reopenLastInstanceOnLaunch` restore — same restore plumbing, just with the user-picked id. Default users (no setting) still take the synchronous path; the await only happens for opt-in users.

## Edge cases
- Stale id (install was deleted) → silently falls through to the default boot. No error toast — this is a convenience.
- First-use not completed → skipped (matches the existing reopen-last gate).
- All other entry points (deep-link, file-association, OAuth callback, second-instance) continue to route through their own handlers; `openStartupSurface` is only called on the standard boot path.

## Tests
53 tests pass across `settings.test.ts` and `installations.test.ts`, including get/set/default roundtrips for the new key and all branches of `resolveAutoLaunchInstall`.


Kosinkadink:
Since the autolaunch option as of v1.0.15 selects last used instance, we should have that be the default value that users can then switch from.

Kosinkadink:
Or maybe it would be fine to have it set to 'dashboard' by default, users can always toggle it to restore old behavior. Not sure which experience people want more as default, but we can resolve that later.

On a separate note, since our settings are so important, we should probably have a more visible way to access them from the dashboard, like a big gear icon somewhere or something. We want to reduce confusion to get to stuff users will likely want to see.

deepme987:
pulled this idea directly from discord -- https://discord.com/channels/1218270712402415686/1512196069809852596/1513754995252658226

I do like the option. As for my stance, I think the default should be dashboard, and only opt-in to launch X instance by default. Default on for launch feels counter-intuitive.

Kosinkadink:
yep dashboard should be fine, i find it annoying that it boots up my instance unless i toggle that feature off

Kosinkadink (approved):

PR #1038, opened 2026-06-10
fix(git): fall back to system git when bundled pygit2 can't authenticate (#1036)

### LINKED ISSUE #1036: Bug: On Mac, failure to update
Error message:

```
Update process failed with exit code 1.

[PRE_UPDATE_HEAD] 64b8457f55cd7fb54ca7a956d9c73b505e903e0c
Creating backup branch: backup_branch_2026-06-10_04_04_13
[BACKUP_BRANCH] backup_branch_2026-06-10_04_04_13
Uncommitted changes saved to backup branch.
Fetching from origin…
[ERROR] Failed to fetch from origin: authentication required but no callback set
Check your internet connection and try again.
```

The same error prevented a successful migration (failed when cloning git into the old ComfyUI directory that legacy desktop had).

Additional information: this user has a system git that is set up to use ssh.

Kosinkadink:
## Summary

Fixes #1036 - on macOS (and any locked-down setup) Desktop's update/migration failed with:

```
[ERROR] Failed to fetch from origin: authentication required but no callback set
```

### Root cause
The bundled pygit2/libgit2 honors the user's **global** git config but ships **HTTPS-only (no SSH transport)**. A corporate-style global `insteadOf` rewrite (`https://github.com/` ? `git@github.com:`) makes every pygit2 fetch/clone demand SSH credentials it cannot provide, so it fails with `authentication required but no callback set`. This broke in-place update, migration/clone, and (because the Launcher was forcing `CM_USE_PYGIT2=1`) ComfyUI-Manager v4.

### Approach: prefer the tooling the user already has
Anyone who has configured an `insteadOf` rewrite (or a git proxy) necessarily has **system git** installed. Rather than mutating user/repo git state, we fall back to system git on authentication-class pygit2 failures.

- **Per-op system-git fallback** - `gitClone`, `gitCheckoutCommit`, `gitFetchAndCheckout`, and `update_comfyui.py`'s fetch try pygit2 first and, on an auth-class failure, retry via **system git** when available. System git honors the user's full config (proxy, `insteadOf`, ssh keys, credential helpers).
- **`COMFY_FORCE_PYGIT2=1`** developer override - disables that per-op fallback and forces Manager's pygit2 backend, so the pygit2 path stays exercised during the beta.
- **Manager no longer force-set to `CM_USE_PYGIT2`** by default - it now prefers system git when present and falls back to bundled pygit2 only when system git is absent (or when forced).
- **No on-disk rewrites** - removed the persistent SSH?HTTPS `origin` remote rewrites from the update path (risky if interrupted by power loss/crash). The in-memory libgit2 config-search-path blanking and `http.proxy` preservation remain for the pygit2-only path (system git absent).

### Why not just blank the global config?
That was the earlier idea but it (a) only helped the pygit2 path and (b) felt risky. The system-git fallback is non-destructive and lets locked-down users succeed using the exact tooling they already configured.

### Paired with
Comfy-Org/ComfyUI-Manager#2972 (slim fallback-path hardening on `manager-v4`).

### Validation
`typecheck`, `lint`, `build`, and the full test suite (2059 tests, incl. 11 new fallback tests) pass; Python files compile.


Kosinkadink:
Follow-up commit: **preserve corporate `http.proxy`**.

The bundled pygit2 ships with HTTPS but **no SSH transport** (`pygit2.features & Feature.SSH == False`), so a user's `insteadOf` https->ssh rewrite can never succeed regardless of credentials - forcing anonymous HTTPS is the only viable path. The one corporate setting we'd otherwise lose by blanking the config is the proxy, so we now snapshot `http.proxy` from the global config **before** blanking the search path and pass it explicitly (`proxy=`) to every fetch/clone/list_heads call.

Verified end-to-end against pygit2 1.19.2: proxy is read correctly, `insteadOf` is still neutralized, and fetch/clone with the proxy value succeed.

Known limitation: `http.sslVerify=false` (corporate MITM with an untrusted CA) is still dropped. That's a deliberate choice - auto-disabling TLS verification is security-sensitive. Can add it (snapshot + `certificate_check` callback) if needed.

Kosinkadink:
## Updated approach: system-git fallback instead of on-disk config/remote rewrites

Following review, this PR moves away from persistently mutating user/repo git state and toward a **runtime fallback to system git** on authentication-class pygit2 failures.

### Root cause (issue #1036)
The bundled pygit2/libgit2 honors the user's **global** git config but ships **HTTPS-only (no SSH transport)**. A corporate-style global `insteadOf` rewrite (`https://github.com/` ? `git@github.com:`) makes pygit2 fetch/clone fail with `authentication required but no callback set`. This broke update, migration/clone, and Manager v4 (which Launcher was forcing onto pygit2).

### What changed
- **System-git fallback**: `gitClone`, `gitCheckoutCommit`, `gitFetchAndCheckout` (and `update_comfyui.py`'s fetch) now try pygit2 first and, on an auth-class failure, retry via **system git** when one is available. System git honors the user's full config (proxy, `insteadOf`, ssh keys, credential helpers) - and anyone running a git proxy / SSH rewrite invariably has system git installed.
- **`COMFY_FORCE_PYGIT2=1`** developer override: keeps every op on pygit2 and forces Manager's pygit2 backend, so the pygit2 path stays exercised during the beta/dev phase.
- **Manager no longer force-set to `CM_USE_PYGIT2`** by default: it now prefers system git (GitPython) when present and falls back to bundled pygit2 only when system git is absent (or when forced).
- **No more on-disk rewrites**: removed the persistent SSH?HTTPS `origin` remote rewrites from the update path (avoids risk if interrupted by power loss/crash). The in-memory libgit2 config-search-path blanking and `http.proxy` preservation remain for the pygit2-only path.

### Why not just blank the global config?
That was the earlier idea but it (a) only helped the pygit2 path and (b) felt risky. The system-git fallback is non-destructive and lets locked-down users succeed using the exact tooling they already configured.

Validation: typecheck, lint, build, and full test suite (2059 tests, incl. 11 new fallback tests) pass.

MaanilVerma (approved):

deepme987 (approved):

PR #1050, opened 2026-06-10
fix(update): surface update button when picking a stable tag on an up-to-date install

Kosinkadink:
## Problem

In the per-instance **Update** tab, the stable channel shows a `ComfyUI Version` dropdown regardless of whether an update is available. But the actionable buttons (`Update Now` / `Copy & Update`) only come from server-built `data.actions`, which are only attached when `card.data.updateAvailable && hasGit`.

So when an install is **already up to date on the latest stable tag**, the user can pick an *older* stable tag from the dropdown but **no button appears** — there's no way to act on the selection (e.g. to downgrade).

## Fix

Synthesize an `update-comfyui` action in `ChannelPicker.vue` when:
- the drafted channel is `stable`,
- a valid `vX.Y.Z` tag is picked that differs from the installed tag, and
- no server-provided `update-comfyui` action already exists (the up-to-date case).

The backend (`handleUpdateComfyUI` -> `runComfyUIUpdate` with `--tag vX.Y.Z`) already fully supports upgrade/downgrade to a specific tag via `targetTag`, which `attachTargetTagIfNeeded` adds on click. Only the known-installed-tag case is handled, so we never offer an update we can't pin (e.g. installs ahead of their base tag rendered as `v0.24.1 +5`).

The manual `Check for update` button is hidden once the synthesized update action is shown, to avoid clutter.

## Tests

Adds `ChannelPicker.test.ts` covering: no button when picked == installed, button appears + emits `targetTag` when a different tag is picked, and no synthesis when the installed tag isn't a clean `vX.Y.Z`.

All typecheck / lint / 2109 tests / build pass.

### DIFF
--- a/src/renderer/src/views/comfyUISettings/ChannelPicker.test.ts
+++ b/src/renderer/src/views/comfyUISettings/ChannelPicker.test.ts
@@ -1,0 +1,109 @@
+import { flushPromises, mount } from '@vue/test-utils'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { createI18n } from 'vue-i18n'
+import type { ActionDef, DetailField } from '../../types/ipc'
+import { TID } from '../../../../shared/testIds'
+import BaseSelect from '../../components/ui/BaseSelect.vue'
+import ChannelPicker from './ChannelPicker.vue'
+
+const i18n = createI18n({
+  legacy: false,
+  locale: 'en',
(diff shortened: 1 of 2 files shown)


Kosinkadink:
Superseded by #1051, which removes the ComfyUI version selector from the update screen entirely for this release (version selection stays in the new-install wizard). Closing.

PR #1051, opened 2026-06-10
fix(update): remove ComfyUI version selector from the update screen

Kosinkadink:
## Why

The per-instance **Update** tab exposed a `ComfyUI Version` dropdown for the stable channel. Choosing a specific tag there was confusing:

- After updating, the dropdown kept its **stale selection** instead of resetting to the version that will actually be installed.
- You could press **Update** with the dropdown set to a version you were *already on* (the modal would say `v0.22.0` while you're already on `v0.22.0`).

For this release we want version selection to live **only in the new-install wizard**, not the update screen.

## Change

Remove the version picker (and its `targetTag` wiring / `attachTargetTagIfNeeded`) from `ChannelPicker.vue`. The update screen now just switches channel (stable/latest) and updates to that channel head.

The new-install version picker is a **separate code path** (`InstallWizardModal.vue` / `QuickInstallModal.vue` via the `release` field) and is untouched. Backend `targetTag` plumbing is left intact (harmless, validated) in case the picker is reintroduced.

## Supersedes

Replaces #1050 (which fixed a symptom of the now-removed picker). #1050 is closed.

## Verification

typecheck, lint, 2106 tests, build — all pass.

### DIFF
--- a/src/renderer/src/views/comfyUISettings/ChannelPicker.vue
+++ b/src/renderer/src/views/comfyUISettings/ChannelPicker.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { computed, onBeforeUnmount, onMounted, reactive, ref, watch } from 'vue'
+import { computed, onBeforeUnmount, reactive, ref, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { Loader2 } from 'lucide-vue-next'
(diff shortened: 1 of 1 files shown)


deepme987:
can we comment this out instead?

deepme987 (approved):
approving, but in favor of commenting that deleting ; easier to fix later

PR #1058, opened 2026-06-11
fix: honor per-install model paths for downloads + overhaul Storage model-paths UI (#1045)

### LINKED ISSUE #1045: When shared models are turned off, ComfyUI opeorply only looks at hte right directories, but the desktop app does not - downloads 'succeed' instantly, and point to the model files in shared directories.

Kosinkadink:
## Summary

Fixes #1045. Two related things:

1. **Backend correctness** — when an install turns **shared models off**, ComfyUI searches only that install's own model directories, but the launcher's download manager always used the **global shared library**. Downloads "succeeded" instantly (the file existed in a shared dir ComfyUI no longer sees) and pointed at paths the running ComfyUI couldn't find. This PR makes the launcher agree with ComfyUI on **the full set of directories an install searches**, and uses that for downloads.

2. **Storage tab UX** — the previously-invisible `extra_model_paths.yaml` is now surfaced, and the model/input/output lists were reworked so shared vs per-instance vs forced directories are clear at a glance.

## What ComfyUI actually searches (now mirrored)

```
1. <comfyDir>/models/<type>            built-in default
2. <comfyDir>/extra_model_paths.yaml   auto-loaded, user-authored
3. --extra-model-paths-config          launcher-generated
     = global modelsDirs   (useSharedModels = true)
     = per-install modelDirs (useSharedModels = false)
```

Previously the launcher's downloader only knew about the global shared dir.

## Backend changes

**Download manager (`comfyDownloadManager.ts`)**
- Resolves the initiating install (via the comfy webview's `webContents`) and uses **its** model settings.
- **Destination** = the install's UI-visible download target (shared first dir, or per-install promoted/own dir) — never the global shared dir blindly.
- **Existence check** probes only directories the install's ComfyUI actually searches, including arbitrarily-mapped per-type dirs from its own `extra_model_paths.yaml` (e.g. `loras: somedir/myname`). An instant "completed" now fires only when ComfyUI would truly find the file, and only for a real regular file (not a directory that merely shares the name).
- Temp dir lives on the **same volume** as the destination so the final rename stays atomic (avoids cross-device `EXDEV`).
- Hardened URL-keyed dedupe: if another install is already fetching the same URL into a different dir, fail closed rather than report a completion ComfyUI can't see.
- Falls back to the global shared dir when a download can't be attributed to a running install (e.g. panel-initiated retries) — preserving today's behavior.

**Model path resolution (`models.ts`)**
- New `resolveExtraModelPaths()` mirrors ComfyUI's `utils/extra_config.py`: `~`/env expansion, `base_path` resolved relative to the YAML's own dir, per-type subpath joining (arbitrary + multiple subpaths), absolute-subpath passthrough, legacy folder aliases (`clip → text_encoders`, `unet → diffusion_models`), `is_default`, `normpath`.
- New `resolveInstallModelSearchPaths()` aggregates built-in + launcher + extra-yaml dirs into one source of truth.
- `t2i_adapter` is now registered under `controlnet` (matching ComfyUI's built-in secondary dirs) so those models stay visible.
- The `extra_model_paths.yaml` parser moves here (its natural home) and is re-exported from `desktopAdopt.ts` for back-compat. `TEMP_DIR_NAME` is shared between the download manager and the folder filter.

## Storage tab UX

**`extra_model_paths.yaml` is now visible (read-only)**
- The install's yaml is shown as a **single row** in the same bordered models list. Clicking it opens a detail modal that lists every section, each with its `base_path` and resolved per-type directories.
- Modal supports: open base folder / per-type dir, **reveal the yaml in its folder** (not open-in-default-app), a **refresh** button to re-check on-disk status, and missing directories shown as **red path text** (no separate badge).
- The launcher never edits the user's file.

**Clearer directory provenance**
- Shared directories (global model dirs, and shared Input/Output when the toggle is on) now show a distinct **linked-folder glyph** (`FolderSymlink`) instead of relying on pills alone. Per-instance/private dirs get no badge.
- The forced built-in install-own models dir keeps the `Instance` pill; the yaml row uses a plain file glyph and a `YAML` pill.
- The "Primary" pill is renamed **`Downloads`** and the action **`Use for Model Downloads`**, since that directory's only role is where ComfyUI saves newly downloaded models. This also removes the old "Primary vs default" ambiguity.
- Dropped the large shared-models-off warning block (all folders are now shown directly) and tidied the modal's `default` pill styling for consistency.

## Design notes

- **The download target stays a launcher-managed root.** A sparse `extra_model_paths.yaml` base_path (e.g. one that only declares `checkpoints`) is intentionally **not** selectable as the global download target — undeclared types would have no ComfyUI-visible destination, recreating this very bug. Custom-yaml dirs are existence-check + display only.
- **`is_default` reconciliation.** ComfyUI resolves multiple `is_default` by load order (last-insert-wins), and our generated YAML loads last, so the UI-selected download dir remains authoritative.

## Tests
- `resolveExtraModelPaths`: arbitrary/relative/absolute/multiple subpaths, legacy aliases, `is_default`, missing/malformed → `[]`, `custom_nodes` excluded.
- `resolveInstallModelSearchPaths`: shared-on, shared-off (install-own primary), promoted external primary, extra-yaml inclusion.
- `buildExistenceCandidates`: no-context fallback, all model roots, arbitrarily-mapped extra dirs, legacy-alias matching, nested-directory remainder.
- `StoragePane`: yaml file rendered as one row, multi-section modal, reveal-in-folder, per-dir open, red missing dirs, refresh, shared badge present on shared model + I/O rows and absent on private/instance/yaml rows, renamed download action.
- en/zh locale parity maintained.

`pnpm typecheck`, `lint`, `test` (2160 passed), and `build` all green.

## Follow-ups (out of scope)
- Launch-time auto-relaunch folder discovery (`preLaunchExtras`) does not yet consider custom-yaml dirs.
- Per-install download temp dirs aren't covered by the existing global temp-dir cleanup (only matters on crash; temps are renamed away on success).
- GUI-editable / globally-shared `extra_model_paths` sets with per-install visibility selection (discussed; larger follow-up).


Kosinkadink:
Follow-up filed for the URL-keyed download contract limitation noted here (fail-closed guard): #1063. Decoupling download identity from URL to allow concurrent same-URL/different-destination downloads will also require coordinated changes in ComfyUI_frontend.

deepme987 (approved):

Issue #1065, opened 2026-06-11
Automatic desktop update (applied on shutdown) currently might not play nice with Windows shutdowns. 


Kosinkadink:
User report: "My Comfy Desktop was forcibly upgraded to ComfyUI Desktop 2. Since then, every time I shut down and restart Windows, the installation somehow becomes corrupted. I have to reinstall ComfyUI Desktop every single time, and this keeps happening in an endless loop. It’s extremely frustrating."

Theory: Windows shutdown closes app (while it may be running). Update install is immediately triggered, but interrupted by OS trying to shutdown, leaving app in a broken state.

Kosinkadink:
https://x.com/WaiMaksim/status/2065075972022452252?s=20

Kosinkadink:
v1.0.20 now uses "install on startup" for automatic desktop updates on Windows, which should not resolve the issue.

PR #1066, opened 2026-06-11
fix(update): gate startup-install behind config flag; guard install on session-end (#1065)

Kosinkadink:
## Problem

Issue #1065: after the forced upgrade to Desktop 2, some users hit an endless reinstall loop — Windows shutdown/restart corrupts the install every time.

**Root cause (the classic Electron+Windows+NSIS bug, also seen in Cursor and others — see [electron-builder#7807](https://github.com/electron-userland/electron-builder/issues/7807), [#7294](https://github.com/electron-userland/electron-builder/issues/7294)):**

1. An update downloads → electron-updater arms install on `app.on('quit')`.
2. User quits the app normally → the quit handler spawns the NSIS installer **detached** (`detached:true`, `unref()`); the app exits.
3. User shuts down / restarts Windows shortly after.
4. The OS force-kills the still-running detached installer **mid-write** → the install dir is left half-written → app broken → reinstall → re-stage → loop.

ToDesktop drives the same `electron-updater` singleton on Windows, so `autoInstallOnAppQuit` is the relevant lever.

### Option B — default (ON)
- Keep electron-updater's install-on-quit (normal quits still install).
- On `query-session-end` / `session-end`, call `suppressInstallOnQuit()` → set `autoInstallOnAppQuit = false`. The quit handler re-reads the flag at quit time, so an install won't be spawned if the OS is already tearing down.
- `installUpdate()` (the explicit "Restart to update" pill, app still alive) also bails when the session is ending.

### Option C — gated (OFF by default)
- Local config flag only: set `"installUpdatesOnStartup": true` in `settings.json` (not an env var, not remote yet).
- When enabled: disable install-on-quit entirely and apply the staged update on the **next launch** (bounded check + "Updating…" splash, per-version loop-breaker, backstop recovery, version-pinned to the staged marker).

## ⚠️ Honest scope note

Investigation (with the Electron docs + upstream issues) shows **Option B is defense-in-depth, not a full fix**: `before-quit`/`will-quit`/`quit` do **not** fire on a Windows shutdown of a running app, so the quit handler doesn't spawn the installer in that case anyway; and the primary documented vector (normal quit → detached installer → then shutdown) happens *after* our windows are gone, so `session-end` can't fire in time. **Option C is the real fix** (it decouples install from shutdown and preserves automatic delivery). Option B is shipped as cheap insurance while Option C is canaried.

## Telemetry
- `comfy.desktop.app_update.startup_install` — startup install triggered
- `comfy.desktop.app_update.startup_install_skipped` — `reason`: `loop_breaker` | `session_ending` | `not_ready`
- `comfy.desktop.app_update.startup_install_backstop_recovered` — startup install didn't quit; recovered into the normal UI

## Testing
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run build` — pass
- `pnpm run test` — 2142 pass (28 in `updater.test.ts` covering both modes, the config-flag gate, session-end suppression, loop-breaker, and skip telemetry)
- **End-to-end Option C validation requires 2 releases** (a build with this code + flag on, and a newer build to update to). Recommend a ToDesktop staging/preview channel: install A (flag on) → it stages B → relaunch → B installs at startup → verify B runs with no corruption.

## Follow-ups
- Canary Option C on controlled machines via the config flag; watch `startup_install*` telemetry.
- Consider a remote flag later for a population-level canary (deferred).


Kosinkadink:
## Code review (self-review)

Reviewed the implementation for correctness, races, and lockout risk. Findings and how they were addressed:

### Fixed before merge
1. **Splash backstop could strand the user (lockout risk).** The recovery timer originally keyed off `isUpdateInstallQuit()` — but that reflects quit *intent* (`quitReason` is set before `restartAndInstall`). If `restartAndInstall` returns without actually quitting (e.g. the staged installer is gone), the flag stays true forever and the user is stuck on the "Updating…" splash. Now keyed off a real `before-quit` listener, and on recovery we `clearQuitReason()` so the next genuine quit still runs its cleanup (otherwise the `before-quit` cleanup is skipped because `isQuitInProgress()` would stay true).
2. **Ready-state ordering race.** `applyPendingUpdateOnStartup` assumed `runCheck` resolving implied the `'ready'` state was set. The `'ready'` transition happens in the `update-downloaded` handler, which can fire after the check promise settles. Now subscribes via `onUpdateStateChanged` and waits for `'ready'` within the timeout budget.

### Considered, accepted as-is
- **`whenReady` delay:** awaiting the startup check delays subsequent wiring by ≤5s only when an update is staged (gated by `hasPendingStartupUpdate()`); the normal UI hasn't opened yet, and the `before-quit` lifecycle handler is registered before the await. Acceptable.
- **First auto-check (2s) overlapping the startup check:** the updater wrapper dedups concurrent checks via a shared in-flight promise, so no harmful duplication.
- **Stale pending marker:** the realistic stale cases self-heal (the check re-downloads or finds us up to date); not auto-clearing on a non-ready result avoids abandoning a staged update while offline (the two are indistinguishable through the wrapper).

### Validation
typecheck ✅ · lint ✅ · test ✅ (2137; +8 new) · build ✅

Kosinkadink:
Updated direction per review: rather than making staged-on-startup (Option C) the default, this now ships **Option B by default** (suppress install during OS shutdown) and keeps **Option C behind a local off-by-default flag** (`installUpdatesOnStartup` / `COMFY_STARTUP_UPDATE_INSTALL=1`).

Rationale: auto-update is critical for delivering fixes, so we de-risk the channel first with the small, easy-to-test change (download an older build, restart Windows, confirm no corruption) and canary the more invasive startup-install path separately before any wider/remote rollout.

The previously-fixed oracle findings are retained: the startup backstop keys off a real `before-quit` (not quit intent), and the startup path waits for the `ready` state transition rather than just the `runCheck` resolution.

deepme987 (approved):

PR #1067, opened 2026-06-11
fix(update): install staged updates at startup, not on quit (#1065)

### LINKED ISSUE #1065: Automatic desktop update (applied on shutdown) currently might not play nice with Windows shutdowns.

Kosinkadink:
## Summary

Fixes #1065 — Desktop installs becoming corrupted on every Windows shutdown, forcing endless reinstalls.

### Root cause
The app runs its own updater UX (download in background, show a "Desktop Update Ready" pill, install on explicit restart) and disables ToDesktop's loop with `todesktop.init({ autoUpdater: false })`. But it never disabled **electron-updater's** `autoInstallOnAppQuit`, which defaults to `true`. After any background download, electron-updater registers an `app.once('quit', … install())` handler that **silently spawns the NSIS installer whenever the app quits**.

During a Windows shutdown the OS force-kills that detached installer mid-write, leaving a half-installed/corrupted app. Reinstalling restores an older build that re-downloads on next launch and re-arms the same trap → endless loop.

### Fix — install at startup, never on quit (decouples install from shutdown)
- **Disable `autoInstallOnAppQuit`** on the electron-updater singleton (the same instance the ToDesktop runtime drives). Nothing installs on quit/shutdown anymore.
- **Apply on next launch instead.** When a download completes we persist a `pendingDownloadedUpdateVersion` marker. On the next launch a bounded (≤5 s), splash-gated check re-validates the cached installer (no re-download, since electron-updater persists it across restarts) and applies it, then the installer relaunches the app.
- **Silent loop-breaker.** We record the attempted version; if an install doesn't "take" (app still on the old version), we back off silently on the next launch rather than looping. The "Desktop Update Ready" pill remains for manual retry — no nag.
- **`session-end` guard.** A `BrowserWindow` `session-end` flag makes `installUpdate()` a no-op while the OS is shutting down, so even an explicit restart-to-update can't spawn an installer mid-shutdown.
- **Boot safety-net.** If a startup install doesn't actually quit the app (e.g. the staged installer vanished), we fall back to the normal surface so the user is never stranded on the splash.

The explicit pill path (`installUpdate` / `restartAndInstall`) is otherwise unchanged, and auto-check + auto-download are untouched.

### User-visible behavior
- Updates still apply automatically — at the next launch instead of on close.
- At most one "it restarted to update" blip; the very next launch always lets the user in.
- A failed auto-install is silent (no error popups); the pill stays available for a manual retry which surfaces the existing error modal only if the user clicks it.

## Testing
- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- `pnpm run test` ✅ (2137 tests; added 8 covering the new startup-install + session-end + loop-breaker behavior)
- `pnpm run build` ✅

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -1015,6 +1015,8 @@
     "instanceRunningReplace": "Close & Launch",
     "modelFolderRelaunchTitle": "Restarting ComfyUI",
     "modelFolderRelaunchDesc": "New model folders were detected from custom nodes. Restarting so ComfyUI can use them right away.",
+    "updateInstallTitle": "Installing update",
+    "updateInstallDesc": "Applying the latest ComfyUI Desktop update. The app will restart automatically.",
     "launchSplashTitle": "Starting ComfyUI",
     "launchSplashDesc": "Launching your ComfyUI instance…",
     "activity": {
--- a/locales/zh.json
+++ b/locales/zh.json
@@ -1015,6 +1015,8 @@
     "instanceRunningReplace": "关闭并启动",
     "modelFolderRelaunchTitle": "正在重启 ComfyUI",
     "modelFolderRelaunchDesc": "检测到来自自定义节点的新模型文件夹。正在重启，以便 ComfyUI 立即使用它们。",
+    "updateInstallTitle": "正在安装更新",
+    "updateInstallDesc": "正在应用最新的 ComfyUI Desktop 更新。应用将自动重启。",
     "launchSplashTitle": "正在启动 ComfyUI",
     "launchSplashDesc": "正在启动你的 ComfyUI 实例…",
     "activity": {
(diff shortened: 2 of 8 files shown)


Kosinkadink:
Duplicate of #1066 (same branch, created in the same second by a double-submitted request). Closing this one; tracking the change in #1066.

PR #1073, opened 2026-06-12
feat(deeplink): open comfy:// links to cloud.comfy.org in the app

deepme987:
PR **A of 2** for "open cloud.comfy.org links in Desktop" (companion: a cloud-web banner in ComfyUI_frontend that fires the link).

## What
Registers the **`comfy://`** URL scheme and opens deep links into the Comfy Cloud host window:
`comfy://open?path=/workflows/123` → loads `https://cloud.comfy.org/workflows/123` (focus if running, launch if not). Also accepts `?url=<encoded cloud url>`.

## How
- `app.setAsDefaultProtocolClient('comfy')` + `mac.extendInfo.CFBundleURLTypes` in `todesktop.json`.
- Cross-platform receive: macOS `open-url` (buffers a link that arrives before `whenReady`); Windows/Linux from `process.argv` at cold start and from the existing `second-instance` handler.
- New pure module `src/main/lib/deepLink.ts` → `resolveDeepLink(raw): string | null`.

## Security
The raw OS string is untrusted. `resolveDeepLink` accepts only the `comfy:` scheme and **allowlists the resolved origin to `https://cloud.comfy.org`** — rejecting other schemes/origins, smuggled absolute URLs, protocol-relative `//evil.com`, and backslash tricks, with a final origin re-validation. It never returns a non-cloud URL. Fully unit-tested (22 cases).

## Verify
typecheck + lint clean; `deepLink.test.ts` 22/22; full unit suite 2151/2151. **Note:** OS-level scheme *registration* can only be confirmed on a packaged/ToDesktop build (`open "comfy://open?path=/x"`); the parsing/routing/allowlist is what's unit-tested here.


deepme987:
## Companion PRs
- Cloud-web nudge (the "Open in Comfy Desktop" banner that fires these `comfy://` links): Comfy-Org/ComfyUI_frontend#12808

This PR handles the scheme + routing; #12808 is the web side that produces the links. Contract: `comfy://open?url=<encoded https://cloud.comfy.org/...>`.

PR #1079, opened 2026-06-12
fix(update): scope startup-install to Windows, polish update splash, add gated installer-UI flag (#1065)

Kosinkadink:
## Scope

Follow-up to **#1066** (already merged into `main`), which shipped the core fix for #1065 — installing staged Desktop updates at **startup** instead of on quit (Option C, config-gated) plus the **session-end** install-on-quit guard (Option B). This PR is the delta on top of that merge.

> Background on the root cause (Windows + NSIS install-on-quit corruption on shutdown) lives in #1066; not repeated here.

## What this PR changes

**1. Scope startup-install (Option C) to Windows only**
`isStartupInstallEnabled()` now returns `false` off Windows. The corruption it addresses is specific to electron-updater's NSIS install-on-quit; macOS (Squirrel.Mac / ShipIt) and Linux don't have that failure mode, so applying updates at startup there would only add risk to a working update channel. (This scoping wasn't part of the squashed #1066.)

**2. Hold the "Updating…" splash for a readable minimum**
The startup-install splash flashed for a fraction of a second because the bounded readiness check usually resolves instantly off a cached installer. A `STARTUP_INSTALL_MIN_SPLASH_MS` (2s) floor — measured from when the splash is shown, so the check's own elapsed time counts toward it — keeps it readable before the app quits to install. Re-checks `isSessionEnding()` after the hold.

**3. Gated NSIS installer progress UI (`showInstallerUI`)**
New hidden, Windows-only, default-off setting. When enabled, explicit installs (`installUpdate()`) run with `isSilent: false`, so the user sees the installer's **native progress window** during the actual file copy — which our Electron splash can't cover (the copy happens after the app quits). Verified safe for our assisted (`oneClick: false`) installer: on update (`--updated`), electron-builder's `skipPageIfUpdated` skips the welcome/license/directory pages and our existing `customFinishPage` auto-launches + aborts the finish page, so the user sees only a progress window — **no clicks, no NSIS changes**. The install-on-quit path stays silent regardless (electron-updater hardcodes silent on quit).

**4. Fix session-end latch (from CodeRabbit review)**
`query-session-end` (WM_QUERYENDSESSION) is **cancellable**, but the handler latched `setSessionEnding()` on it with no reset path — a canceled shutdown would permanently no-op `installUpdate()` (the "Restart to update" pill) for the rest of the session. Now `query-session-end` only calls `suppressInstallOnQuit()` (reversible, re-armed next launch); `setSessionEnding()` is latched only on the finalized, non-cancelable `session-end`.

**5. Test cleanups (from CodeRabbit review)**
- Advance fake timers just past the 5000ms bounded-check timeout (`5100`) to avoid event-loop boundary flake.
- Extract a shared `findEmitCalls(event)` helper for the repeated telemetry-call lookups.

## How to enable the gated bits (local config, Windows)
`%APPDATA%\Comfy Desktop\settings.json`:
```json
{ "installUpdatesOnStartup": true, "showInstallerUI": true }
```
Neither is remote yet — both are flipped by hand to canary.

## Testing
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run build` — pass
- `pnpm run test` — **2189 pass** (`updater.test.ts` covers the Windows-only gate, splash-minimum hold, the silent-vs-installer-UI toggle, session-end suppression, loop-breaker, and skip telemetry)
- **End-to-end validation requires 2 releases** (a build with this code + the flags on, and a newer build to update to): install A → it stages B → relaunch → B installs at startup with the splash + native progress UI → verify B runs with no corruption.

## Diff
4 files, +180/−34: `src/main/index.ts`, `src/main/lib/updater.ts`, `src/main/lib/updater.test.ts`, `src/main/settings.ts`.


Kosinkadink:
@coderabbitai review

deepme987 (approved):

PR #1084, opened 2026-06-12
Feat/starter templates

MaanilVerma:
## Summary

Adds a **starter-template picker** to the standalone install flow: pick a workflow (Image / Video / Audio / 3D) at install time and land on a **ready-to-run canvas** — the template auto-opens on first launch *and* its required models were pre-downloaded in the background — instead of a blank canvas with "missing model" errors. Picking nothing (**Skip & Install**) preserves today's blank-canvas behaviour exactly. The picker is **gated behind a PostHog experiment** (`treatment` shows it, `control` keeps the legacy flow), with funnel telemetry on each step.

## Changes

**What**

- `bundledTemplates.ts` — curated one-per-modality picks (`flux_schnell`, `text_to_video_wan`, `audio_stable_audio_example`, `3d_hunyuan3d_image_to_model`); metadata copied verbatim from the live template index. Adds pure `shouldWarnVram()` and `buildTemplateDeeplink()` helpers.
- `TemplatePickerStep.vue` — compact selectable-row picker (thumbnail + title + `modality · ~size · VRAM ~x`, description expands in the selected row), full keyboard nav (Arrow/Home/End move+select; Enter/Space via native button), live disk-block + VRAM-warning alerts, modality-glyph fallback on thumbnail load error. Thumbnails are **bundled in-repo** (`'self'`, CSP-safe) since a fresh user has no local template package.
- `InstallWizardModal.vue` — dedicated `template` step after Configure; **Skip & Install** vs **Install**; `skipTemplatePickerStep` setting + "Don't show again" (only when ≥1 local install). The step is **A/B-gated** (`treatment` only) and **auto-skipped** when free disk can't fit even the cheapest template. The wizard resets to Configure when the source changes.
- `installHelpers.ts` — one shared disk decision: `minTemplateModelBytes` (cheapest of N) → `templateDiskRequiredBytes` (×1.1 headroom) → `isTemplateDiskBlocked` → `checkTemplateDiskOrBlock`. Drives the skip-the-step gate, the picker alert, the Install guard, and the save-time hard gate from one source so they can't drift.
- `templateModels.ts` — resolves required models dynamically from the template's workflow JSON (site-packages first, GitHub-raw fallback), scans top-level/node-level/subgraph `models[]`, whitelists hosts (HTTPS-only), sanitizes paths (no traversal), dedupes.
- `templateDownloadTask.ts` + `templateDownloadCore.ts` — background download starting at install-begin; three-layer split (stateful task → pure unit-tested core → display reader). Bounded-concurrency pool (cap 3), 2× per-file retry, Windows MAX_PATH guard, gated-repo (401/403) messaging, surfaced disk + per-file errors.
- `launch.ts` + `launchPhases.ts` — model download shown as the **last** launch step; `serverUp`-gated reader keeps the bar honest through real phases; **launch gate** holds the ComfyUI reveal until the download settles, with a footer **"Skip & open ComfyUI"** and, on failure, a "retry in-app" line + 3·2·1 countdown before opening.
- `attach.ts` — first-launch `?template=<id>&source=default` deeplink (zero frontend change; `useTemplateUrlLoader` opens it), one-shot cleared after.
- `comfyDownloadManager.ts` — Skip hands the resume-capable task off to the title-bar downloads tray (no restart); the mirror is read-only. Template rows now also surface in the **"All Downloads" modal** (live progress + dismiss + clear-all), via the existing `model-download-progress` channel. Finished mirror rows are purged on dismiss / clear-finished so they don't linger.
- `index.ts` + `titlePopup.ts` — `before-quit` confirm when template downloads are still in flight; the downloads tray **auto-opens ~2.5 s after the ComfyUI reveal** when a template download is still running, so the user notices it.

**Breaking**

- `None.` No frontend changes (deeplink only). New `InstallationRecord` fields (`bundledTemplateId`, `pendingTemplateOpen`, `downloadTemplateModels`) are all optional → existing installs load unchanged. IPC channels, specs, and telemetry are preserved (the All-Downloads integration reuses `model-download-progress`; no new channels).

## Review Focus

- **Launch gate** (`launch.ts` `waitForTemplateDownloadGate` + the `serverUp` reader) is the crux: it must show the download as the last step and hold the reveal without falsely marking earlier phases done, then redirect on done/skip/error. The settle condition is one pure primitive (`awaitTemplateDownloadSettled` → `done | error | cancelled | skipped | aborted | absent`).
- **Shared disk decision** — the skip-the-step gate, the picker alert, the Install guard, and the save-time gate all call one `isTemplateDiskBlocked()` so they can't drift. Storage-too-small now **skips the picker** when nothing fits; otherwise the selected pick is blocked and **clicking the (dimmed-but-clickable) Install shakes the error**, mirroring the first-use consent nudge — no dead-end disabled button.
- **A/B gating** — `treatment` shows the picker, `control` keeps the legacy flow; fails closed to `control` on a slow/failed flag fetch; exposure + variant recorded once per session.
- **Cross-OS** — disk (`statfs`), VRAM (`nvidia-smi` / Apple `os.totalmem()` / `systeminformation` fallback, never false-warns), `path.join` throughout, Windows-only MAX_PATH. Verified handled on all three OSes.
- **Scope boundaries** — `InstallWizardModal.vue` / `ProgressModal.vue` are already oversized; this PR adds to them rather than extracting composables. That refactor is a deliberate **follow-up**, not in scope here.

## Testing

Unit + component coverage by design: the branchy logic is pure and Electron-free, so it's pinned without a window; the picker is component-tested; only the live launch flow (real ComfyUI server + multi-GB download) is manual, since the e2e harness never boots a real server. `pnpm typecheck` (node + web + e2e + integration), `pnpm lint`, and `npx vitest run` (**2,286 tests**) all green.

- `templateDownloadCore.test.ts` — cumulative math, `runPool` concurrency/abort, `withRetry` budget + fatal-cancel, `truncateForMaxPath`, `describeDownloadFailure` (401/403 vs generic, no false-match), tray-entry mapping.
- `templateDownloadGate.test.ts` — `awaitTemplateDownloadSettled` resolves done / error (incl. disk pre-flight) / cancelled / skipped / aborted / absent; clears the stale skip flag on settle.
- `templateModels.test.ts` — `sanitizeModelPath` (rejects traversal) + `resolveTemplateModels` URL/host (HTTPS-only) guards.
- `bundledTemplates.test.ts` — `shouldWarnVram` branches; `buildTemplateDeeplink` round-trip + malformed-URL fallback.
- `index.test.ts` — `buildInstallation`: Skip (no template) builds no download; a real pick sets the id + one-shot + download flag.
- `installations.test.ts` — `clearPendingTemplateOpen` fires once, no-ops on a legacy record with no template fields, returns false when the install is gone.
- `comfyDownloadManager.test.ts` — tray-mirror cleanup on dismiss/clear-finished; mirror rows appear in `getAllDownloads()` + fan out a `model-download-progress` broadcast per row.
- `TemplatePickerStep.test.ts` — disk-block / VRAM alerts, Recommended tag, select + keyboard nav, Enter/Space native activation, thumbnail `@error` → glyph, `nudgeDiskError()` shakes when blocked / no-ops when not.
- `installHelpers.test.ts` — `templateDiskRequiredBytes` headroom, `isTemplateDiskBlocked`, `minTemplateModelBytes` (none / smallest / ignores-zero) feeding the skip-the-picker gate.

**Manual verification (`pnpm dev`) — every edge case:**

| # | Scenario | Steps | Expected |
|---|---|---|---|
| 1 | Template picker appears | New Install → Standalone → Configure → Continue | Template step: 4 rows (Image "Recommended" + pre-selected, Video, Audio, 3D); thumbnail + title + `modality · ~size · VRAM ~x` |
| 2 | Thumbnails | Look at all rows | Real previews (bundled, offline-safe); no broken-image glyph, no title bleed |
| 3 | Select + keyboard | Click rows; ↑/↓/Home/End; Enter/Space | Yellow ring + check; description expands; keyboard moves *and* selects |
| 4 | Install a template | Pick Image → Install | "Downloading template models" is the **last** launch step; ComfyUI opens with the template on canvas |
| 5 | No stepper jump | Watch the loader bar | Advances smoothly through real phases; model step activates only at the end |
| 6 | Skip & Install | Picker → Skip & Install | No template, no download (`[templateDownload]` absent), blank canvas |
| 7 | Slow → gate hold → Skip → tray | Heavy template; watch launch | Holds on the last step (server up, download running) + **"Skip & open ComfyUI"**; click → opens; tray **auto-opens ~2.5 s in** and shows the download continuing (not restarted); the row also appears in **"All Downloads"** |
| 8 | VRAM warning | GPU below a pick's recommendation (or lower `recommendedVramBytes`) | Amber "may run slowly"; **Install stays enabled**; silent on AMD/Intel/unknown VRAM |
| 9 | Disk block + wiggle | Heavy template on a small disk (free disk between the cheapest and this pick) | Red disk error; **Install reads disabled (dimmed) but is clickable** — clicking it **shakes the error** (no install); a pick that fits clears it; Skip & Install always works |
| 9b | Picker auto-skipped (no fit) | Volume too small for **even the cheapest** template; Configure → Continue | Template step **skipped entirely** — Continue installs with no template (no picker, no error) |
| 10 | Don't-show-again | With ≥1 install → wizard → template step | Checkbox appears (absent for first-ever user); tick + Install → next New Install skips the step |
| 11 | Quit mid-download | Heavy install, ⌘Q while downloading | Confirm dialog ("Quit Anyway / Keep Downloading", default Keep) |
| 12 | Download error → countdown | Go offline mid-download, or a gated repo | Substatus red + bold + X; gate shows "failed — retry in-app" + 3·2·1 → ComfyUI opens automatically |
| 13 | Gated repo (401/403) | Point a model at a license-gated HF repo | Clear "requires a login or license — open in ComfyUI to sign in" log line; non-fatal, counts toward retry |
| 14 | Source change resets step | Reach template step → Back → switch to a non-standalone source | Returns to Configure; picker not shown for the new source |
| 15 | A/B gate | Toggle the PostHog flag (or `VITE_FORCE_TEMPLATE_PICKER=1`) | `treatment` shows the picker; `control` skips straight to install; exposure recorded once |
| 16 | Resize | Wizard narrow → wide | Rows reflow cleanly; no overflow/overlap; footer reachable |
| 17 | Cross-OS smoke | Run on macOS / Windows / Linux | Disk pre-check, VRAM detect, and model paths behave per-OS |


deepme987:
Wired the starter-template picker into a PostHog experiment + funnel telemetry in `bf7024eb` so we can ship at 50% and read activation-funnel deltas.

## Experiment

- Flag key: `desktop-starter-templates-picker`
- Variants: `control` (legacy single-screen Configure) vs `treatment` (picker step shown).
- Lookup uses the cache-first `telemetryGetExperimentFlag` plumbing (the same path `FirstUseTakeover` uses for `desktop-fork-default`). Boot-time fetch lands on disk; the running modal reads from cache, so the wizard never blocks on the network.
- Exposure event (`comfy.desktop.experiment.exposed`) is fired once per session main-side.
- Variant is also pinned as a person property (`experiment_desktop-starter-templates-picker = control|treatment`) so any downstream activation event (boot success, first workflow exec) can be sliced by arm without joining back through exposure.
- Defaults to `control` on flag miss / network failure → legacy flow ships when the experiment can't be resolved.

The `treatment` arm still respects the user's `skipTemplatePickerStep` opt-out and the existing standalone-source gate, so an opted-out returning user in treatment still won't see the step.

## Funnel events (all `comfy.desktop.template.*`)

| Event | When | Props |
|---|---|---|
| `picker_shown` | `step` transitions to `template` | `template_count`, `has_local_install`, `default_template_id`, `variant` |
| `selected` | user picks a real (non-`None`) template, dedup'd on value change | `template_id`, `size_bucket`, `variant` |
| `install_confirmed` | "Install" clicked from picker | `template_id`, `size_bucket`, `has_models`, `dont_show_again`, `variant` |
| `skipped` | "Skip & Install" clicked | `had_template_selected`, `candidate_template_id`, `dont_show_again`, `variant` |

These compose with the existing `comfy.desktop.template.download.skipped` (ProgressModal) so we have shown → selected → install_confirmed → download_skipped/completed → first_workflow_exec.

## PostHog setup (before turning the experiment on)

1. Create a multivariate feature flag in **Prod Comfy (Homepage, Cloud, Hub)** (project 204330):
   - Key: `desktop-starter-templates-picker`
   - Variants: `control` (50%), `treatment` (50%)
   - Rollout condition: `app_name = comfyui-desktop` (or `id_class is set`) so it only targets desktop installs, not website / cloud users on the same project.
2. Create the experiment in PostHog Experiments using that flag, with primary metric = `comfy.desktop.template.install_confirmed` (north-star: pick + install a starter template) and a secondary metric on `comfy.desktop.comfyui.boot_success` / first workflow execution.
3. The renderer reads the flag at modal open; cached value applies on the next install, refreshed value lands on disk for the boot after that, per the `experiments.ts` contract (one-boot-of-lag is intentional).

## Scope still open (not in this commit)

- Main-side download lifecycle (`template.download_started / .download_completed / .download_failed`) and `template.auto_opened` from `attach.ts`. Happy to add in a follow-up; kept this commit focused on the install-time funnel so the experiment can ship.
- "Don't show again" toggle event is intentionally folded into `install_confirmed` / `skipped` rather than a standalone event (the value is already on both terminal events).

Pre-commit ran `typecheck` + `lint` + the 15 `TemplatePickerStep` unit tests all green.

deepme987:
PostHog side is fully wired (project 204330 / Prod Comfy). Nothing left for the team to click before merge — flag is in **draft** so it doesn't fire until the experiment is launched from the UI.

## Links

- **Experiment:** https://us.posthog.com/project/204330/experiments/376772 — `desktop-starter-templates-picker` (draft, 50/50, 100% rollout)
- **Feature flag:** https://us.posthog.com/project/204330/feature_flags/714811 — multivariate `control` / `treatment`, scoped to `id_class is_set` (desktop installs only — won't match website / cloud-only users in the same project)
- **Comparison dashboard:** https://us.posthog.com/project/204330/dashboard/1706793 — 8 insights + header tile

## Metrics on the experiment

| Slot | Metric | Type |
|---|---|---|
| Primary | exposed → `comfy.desktop.execution.first_completed` | funnel |
| Secondary | `picker_shown` → `selected` → `install_confirmed` | funnel |
| Secondary | `install.method.selected` → `comfyui.boot_started` | funnel |
| Secondary | Workflow executions per user (`execution.completed`, total) | mean |
| Secondary | Picker skip rate (`template.skipped`, total) | mean — goal: decrease |
| Secondary | `boot_started` → `first_completed` | funnel |

Exposure event is set to `comfy.desktop.experiment.exposed` filtered to `experiment_key = desktop-starter-templates-picker`, so the denominator is "users who opened the install wizard while enrolled," not "every desktop boot." Tighter exposure window than `$feature_flag_called`.

## Dashboard tiles

1. **Activation funnel** — exposed → first execution success, broken down by variant (primary north-star)
2. **First-execution conversion rate** — same numerator/denominator, rendered as a side-by-side bar with `B/A*100`
3. **Picker funnel** — shown → selected → install_confirmed (treatment-only; control fires zero by construction)
4. **Picker engagement timeline** — shown / install / skipped daily DAU
5. **Template selection distribution** — install_confirmed broken down by `template_id`, bar chart
6. **Install completion funnel** — method.selected (standalone) → boot_started → first_completed, by variant
7. **Model-download skip rate** — `template.download.skipped / template.install_confirmed * 100`
8. **Workflow executions per user** — weekly_active broken down by variant (retention check)
9. **Exposure population (sanity)** — daily exposed DAU broken down by `variant`, to catch flag-targeting drift

Test accounts filtered out everywhere.

## Launch flow when ready

1. Merge PR #1084 → events ship.
2. Wait for `comfy.desktop.experiment.exposed` to start landing in PostHog (open the wizard once on a build to verify).
3. Hit **Launch** on the experiment page (https://us.posthog.com/project/204330/experiments/376772).
4. The dashboard starts populating as users come through the wizard.

The flag stays **draft** until launched, so no risk of accidentally serving variants before merge.

docs/starter-templates-handoff.md
```diff
@@ -0,0 +1,175 @@
+# Starter Templates — Engineering Reference
```
deepme987: I believe this is an LLM artifact that we don't need in code

src/main/lib/experiments.ts
```diff
@@ -112,20 +130,9 @@ export function initExperiments(opts: {
...
-      //
-      // Empty result is also ignored on disk: ambiguous (timeout vs
-      // legitimately no flags configured), and overwriting with empty
-      // would roll every cached variant back to fallback on next boot.
-      if (Object.keys(flags).length > 0) {
-        writeCache(flags)
-      }
+      if (Object.keys(flags).length === 0) return
+      writeCache(flags)
+      backfillSessionCache(flags)
```
deepme987, line 134: verify if this is the right way; and if this is the right way, ensure the cloud flag uses the same approach too. Right now, I believe cloud status uses a different flag

deepme987 (changes requested):

src/main/lib/experiments.ts
MaanilVerma: This backfillSessionCache is intentionally different from how cloud merges, and that difference is the whole point. Side by side:


```
// cloudCapacity.ts — overwrites in-memory with the fresh fetch, always
.then((value) => {
  if (typeof value === 'string' && VALID.has(value)) cached = value
})

// experiments.ts — fills ONLY keys absent at boot, never overwrites
function backfillSessionCache(flags) {
  for (const [key, value] of Object.entries(flags)) {
    if (!(key in cached)) cached[key] = value   // <- skip already-assigned keys
  }
}
```

**Why they differ — kill-switch vs experiment:**

Cloud capacity is a live kill-switch: if capacity changed, we want the freshest value reflected now. Overwriting is correct.
An experiment variant must NOT flip mid-session — that corrupts the experiment population and changes the UI under the user. So we keep the boot value for keys we already have, and only back-fill keys that were absent (the first-boot case). That's what the if (!(key in cached)) guard buys us.
So same async-init plumbing, but the merge rule is deliberately different per the data type. The back-fill's doc comment spells this out. Sound to me — but happy to talk through if you'd want experiments to behave like cloud (it shouldn't, IMO, for variant stability).

The async-await accessor I added in this PR (getFlagAsync) is the same as cloud, though — that part's now aligned. 

@deepme987

deepme987:
image
can we avoid downloading it WHILE it's downloading Comfy?

MaanilVerma:
Closing this as it has been superseded by the merged https://github.com/Comfy-Org/Comfy-Desktop/issues/1135 containing these changes.

PR #1088, opened 2026-06-13
feat(updater): startup update splash countdown + skip NSIS install-mode page on update

Kosinkadink:
## What

Two related startup-update improvements, bundled so they can be tested together in one build.

### 1. 5-second countdown on the startup update splash

When a previously-downloaded Desktop update is applied on startup (Windows-only Option C, gated behind the `installUpdatesOnStartup` config flag), the purple "Updating ComfyUI Desktop" splash now:

- Explains a new version was downloaded and is about to install.
- Shows a localized **"Restarting in {n}s…"** line that ticks down from 5.
- Switches to **"Restarting now…"** when it reaches zero, right as the install fires.

Previously the splash could flash for a fraction of a second before the app quit, feeling like a glitch. This gives the user a few seconds to read what's happening before the restart.

### 2. Skip the NSIS install-mode page on non-silent update (cherry-picked from #1087)

A non-silent update (electron-updater passes `--updated` but no install-mode flag) would otherwise stop on the install-mode page — the one pre-progress page electron-builder does **not** auto-skip on update — forcing the user to confirm before the progress bar appears. `customInstallMode` now re-uses the mode `initMultiUser` already detected to force-skip that page on update (ambiguous cases still show it).

## Changes

- `relaunchPage.ts` — `showSplashPage` accepts an optional `countdownSeconds`; renders a countdown line with an inline ticker.
- `updateSplash.ts` — new `UPDATE_INSTALL_COUNTDOWN_SECONDS = 5`; passes it to the splash; splash height bumped to fit the extra line.
- `updater.ts` — `STARTUP_INSTALL_MIN_SPLASH_MS` raised `2000 → 5000` so the splash stays up for the full countdown (kept in sync with the countdown constant).
- `locales/en.json`, `locales/zh.json` — updated title/desc + new `updateInstallCountdown` / `updateInstallStartingNow` strings.
- `updater.test.ts` — updated the min-splash-hold timing test for 5000ms.
- `scripts/installer.nsh` — skip the install-mode page on non-silent update.

## Scope / safety

- Windows-only, startup-install path only.
- Countdown still gated behind the existing `installUpdatesOnStartup` flag — no behavior change unless that's enabled.
- Worst case if the inline ticker didn't run: the splash simply shows title + desc without the number (graceful) — the update still proceeds normally.

## Validation

`pnpm run typecheck`, `pnpm run lint`, `pnpm run build`, `pnpm run test` (2189 tests) all green.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -1023,8 +1023,10 @@
     "instanceRunningReplace": "Close & Launch",
     "modelFolderRelaunchTitle": "Restarting ComfyUI",
     "modelFolderRelaunchDesc": "New model folders were detected from custom nodes. Restarting so ComfyUI can use them right away.",
-    "updateInstallTitle": "Installing update",
-    "updateInstallDesc": "Applying the latest ComfyUI Desktop update. The app will restart automatically.",
+    "updateInstallTitle": "Updating ComfyUI Desktop",
+    "updateInstallDesc": "A new version was downloaded and is about to install. The app will close briefly and reopen automatically.",
+    "updateInstallCountdown": "Restarting in {seconds}s…",
+    "updateInstallStartingNow": "Restarting now…",
     "launchSplashTitle": "Starting ComfyUI",
     "launchSplashDesc": "Launching your ComfyUI instance…",
     "activity": {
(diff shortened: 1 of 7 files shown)


deepme987 (approved):
ooo I like this

PR #1102, opened 2026-06-13
Feat/titlebar theme match and version pill

### LINKED ISSUE #1060: Can't see version of ComfyUI + Update pill at the same time in dashboard
Image

MaanilVerma:
## Summary

Match the desktop title bar to ComfyUI's in-page theme while inside an instance (dark/light, Win/Linux/macOS), and stop the dashboard tile from hiding an install's version when an update is available. Two independent fixes, one commit each.

## Changes

**What**

- `applyComfyTheme` (`src/main/host/attach.ts`) — drive the Vue header and the OS window-controls overlay from ComfyUI's reported `--comfy-menu-bg` in one call, so the strip behind min/max/close stays seamless with the bar (the divergence fixed in #647). `symbolColor` is luminance-derived (`readableSymbolColor`) so the glyphs stay legible on any theme. Instance-only — the install-less chooser keeps `--titlebar-bg`.
- `readableSymbolColor` (`src/main/lib/theme.ts`) — new helper: parses `#rgb`/`#rrggbb`/`rgb()`, returns `#333` on light bgs / `#ddd` on dark by perceived luminance; dark-safe fallback on parse failure.
- `isLight` (`src/renderer/src/comfyTitleBar/useTitleBarIdentity.ts`) — un-stubbed from `false` to the real luminance test, activating the existing `.is-light` chrome.
- `TitleBarApp.vue` — bind `--titlebar-bg-active` / `--titlebar-icon` inline from the reported theme; extend `.is-light` to the resting/hover/open states it never covered (install pill, downloads tray, icon buttons) so nothing washes out or lifts to invisible yellow on a light bar.
- `ChooserInstallTile.vue` + `chooser-tiles.css` — render the version pill independently of the action pill (was a `v-if`/`v-else-if` chain), right-align the update/migrate action (`.chooser-tile-pill-action`), and carry the target version in the label (`Update v0.25.0`) via `chooser.updatePillVersion`, sourced from `statusTag.version`.
- `src/types/ipc.ts`, `locales/en.json`, `locales/zh.json` — add `statusTag.version` to the renderer `Installation` type and the new `updatePillVersion` key.

**Breaking**

None. The theme path only runs for attached instances (`applyComfyTheme` is bound inside `attachInstall`); the chooser/dashboard still reverts via `applyChooserHostTheme`. The `desktop2-theme-report` IPC channel, overlay-at-creation defaults, and the migrate/update action emits are unchanged.

## Review Focus

- The crux is that both title-bar surfaces are painted from the same `theme.bg` in one call — that's what keeps the min/max/close strip seamless (re #647). Confirm no path updates the Vue header theme for an instance without the matching overlay repaint.
- `.is-light` was already authored but gated behind a hardcoded `isLight = false`; this enables it and fills the resting/hover/open gaps. Worth a glance on a custom (non-standard) ComfyUI theme.
- Scope is title-bar theming + the dashboard version/update pill only. macOS has no app-controlled overlay (traffic lights are system-drawn), so it themes the header only — intentional, not a gap.

## Testing

Unit/typecheck-only by design: the title-bar binding and pill split are presentational, and the existing `TitleBarApp` suite already exercises the theme-changed and pill-render paths. No new user-visible IPC behavior to E2E.

- `src/renderer/src/comfyTitleBar/TitleBarApp.test.ts` — 59 tests pass (theme binding + pill render unaffected).
- `pnpm typecheck` (node/web/e2e/integration), `pnpm lint` — clean.
- Manual: launch an instance, flip ComfyUI dark⇄light → header + Win/Linux min/max/close follow live; return to dashboard → reverts to purple. Force an instance with an available update → version pill and `Update v…` pill both visible.

Screen Recording

https://github.com/user-attachments/assets/dae8c7e0-e2af-4b55-ab14-7c1076fb72d2

closes #1060

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -37,6 +37,7 @@
     "filterCloud": "Cloud",
     "filterRemote": "Remote",
     "updatePill": "Update",
+    "updatePillVersion": "Update {version}",
     "migratePill": "Migrate",
     "openInstall": "Open",
     "manageInstall": "Manage",
--- a/locales/zh.json
+++ b/locales/zh.json
@@ -37,6 +37,7 @@
     "filterCloud": "云端",
     "filterRemote": "远程",
     "updatePill": "更新",
+    "updatePillVersion": "更新到 {version}",
     "migratePill": "迁移",
     "openInstall": "打开",
     "manageInstall": "管理",
--- a/src/main/host/attach.ts
+++ b/src/main/host/attach.ts
@@ -5,7 +5,7 @@ import { getModelDownloadContentScript } from '../lib/comfyContentScript'
 import { getComfyTerminalContentScript } from '../lib/comfyTerminalContentScript'
 import { closeInstallPopouts } from '../lib/popoutWindows'
 import { _operationAborts, sourceMap } from '../lib/ipc/shared'
-import { TITLEBAR_BG } from '../lib/theme'
+import { readableSymbolColor } from '../lib/theme'
 import * as mainTelemetry from '../lib/telemetry'
 import { refreshCloudUserTier } from '../lib/userTier'
 import { noteCloudEntered } from '../lib/cloudEntry'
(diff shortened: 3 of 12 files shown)


Kosinkadink:
The changes seem fine overall, but on Latest on Github track, the look a bit off (text is shortened on Source + Version pills)
image

Kosinkadink:
That can be resolved in a different PR, but this one looks good. I will approve and merge.

Kosinkadink (approved):

MaanilVerma:
@Kosinkadink thoughts what should I do, let pills go into second row or increase width of cards ?

Kosinkadink:
Hmm, I feel like second row would be the cleanest

PR #1123, opened 2026-06-16
feat(downloads): serve "View All Downloads" as a title-bar popup

### LINKED ISSUE #1103: View All Downloads Opening Delay + Irrelavant Border
Image

- Click View All Downloads for the very first time it takes bit of time opening after click 
- THere is irrelavant border around the whole screen of yellow color.

MaanilVerma:
## Summary

Serve the full **"View All Downloads"** view as a centred title-bar `WebContentsView` popup — the same surface `menu` / `instance-picker` / `global-settings` already use — instead of a DOM modal mounted in the panel renderer via the `downloads-v2` overlay mode. It now opens instantly with a real blurred backdrop, and the long-standing kind-allowlist bug in the preload validators is fixed.

## Changes

**What**

- `src/types/ipc.ts` — add `POPUP_KIND` + `PopupTheme` so the popup config discriminant has one source across main, preload, and renderer instead of literal strings re-typed in each file.
- `src/main/popups/titlePopup.ts` — add the `downloads-full` popup kind (backdrop + centred card), reusing the tray's existing `tray-state-changed` download feed (no new data pipeline). Generalise the global-settings sizing into `computeCenteredCardBounds` + `kindIsCentered`, shared by the open path and the resize refit. Route `comfy-titlepopup:open-downloads-modal` to a new `openDownloadsFullForHost`.
- `src/preload/comfyTitlePopupPreload.ts` — **the core bug:** both `isPopupConfig` and the `onWillShow` handler gated on a hardcoded kind allowlist and silently dropped any kind they didn't enumerate, so `downloads-full` never reached the renderer (it stayed on the tray `downloads` view). Key both off `POPUP_KIND`.
- `src/renderer/.../DownloadsFullView.vue` — port the `DownloadsModal` design verbatim (filter chips, status badges, progress bars, footer counts, clear-finished), rewired from the panel store / `window.api` to the popup prop + `__comfyTitlePopup` bridge, with the popup card replacing `BaseModal` chrome. Drop the per-row entrance animation whose lingering `transform` layer broke pointer hit-testing (dead `:hover` / cursor) in the transparent popup view.
- `src/renderer/.../TitlePopupApp.vue` — render `DownloadsFullView` for `kind === 'downloads-full'` and give it the shared centred-card chrome.
- `src/main/host/*` + `src/renderer/src/panel/*` — remove the now-dead `downloads-v2` body mode (`BodyMode`, `ComfyPanelKey`, `PanelKey`, the `PanelApp` mount, the host-window overlay branch) and the old `DownloadsModal.vue`. `feedback` stays the lone overlay mode.

**Breaking**

- None. The `comfy-titlepopup:open-downloads-modal` trigger IPC, the bridge method name, and the tray `downloads` popup are unchanged; the full view is the same design served from a different surface.

## Review Focus

- The preload validator fix is the root cause — without it the popup opens but the renderer never flips off the tray kind. Worth confirming no other handler carries a stale kind allowlist.
- `computeCenteredCardBounds` / `kindIsCentered` now back both `global-settings` and `downloads-full`; check the shared sizing + centring + refit paths.
- Scope is the Downloads full view only. The dropped row entrance animation is intentional (it was the hit-testing culprit); `BaseModal` is deliberately untouched.

## Testing

E2E is the right level here — this is a renderer/IPC surface swap with no new business logic — so the new specs drive the real tray → full-popup hand-off rather than re-unit-testing the ported view. `pnpm typecheck` (node/web/e2e/integration), `pnpm lint`, and `pnpm test` (2205 unit tests) all pass; each of the 6 commits passes the pre-commit gauntlet independently. Verified manually that "View All Downloads" opens the centred popup with live rows, working hover/cursor, filters, and clear-finished.

- `e2e/downloads-shelf.test.ts` — "View All Downloads" reopens the same view as `.dlm-panel` with every seeded row, filter chips, and footer; and the empty state when nothing is seeded.
- `src/main/host/{panelView,registry}.test.ts` — panel-key fixtures retargeted at `feedback`.
- `src/renderer/src/panel/PanelApp.test.ts` — dropped the dead `DownloadsModal` mock.

https://github.com/user-attachments/assets/fd2f612b-f5de-478e-ba0b-957bf0da8f17

closes #1103


Kosinkadink:
nice, taking a look

Kosinkadink (approved):

PR #1128, opened 2026-06-16
docs: add linux to supported platforms badge

Nynxz:
https://github.com/Comfy-Org/Comfy-Desktop/commit/b82c0276b2d9aab0b452ccdcc77c09fc3b85b91d removed Linux from the supported operating system badge. This PR adds it back to remove mixed signals and keeps it removed from the installer badge - where I assume the actual support is dropped


deepme987:
It's not the docs, we had a change that disabled the linux builds.
They should be re-added in the next release [<12hrs]. dl.comfy.org should point you to it!

PR #1129, opened 2026-06-16
fix(titlebar): sync menu zoom reset state

Nynxz:
## Problem

Resetting zoom from the title menu reset the ComfyUI webContents directly, but programmatic `setZoomLevel(0)` does not trigger the same zoom update path as user zoom changes. That could leave the titlebar zoom pill visible/stale after using **Reset Zoom** from the menu.

The existing dropdown e2e also asserted a dashboard/chooser Reset Zoom path by forcing the internal dummy `comfyView` zoom level. That is not a supported user-visible dashboard zoom path.

**Broken**

https://github.com/user-attachments/assets/36eed195-aac5-4bb4-a46b-08d209090bd1

## Changes

- Route title-menu **Reset Zoom** through the same per-install reset closure used by the titlebar zoom pill.
- Share zoom-reset telemetry payload generation across titlebar, menu, and keyboard shortcut reset paths, with a typed `source`.
- Keep **Reset Zoom** out of the dashboard/chooser menu because dashboard zoom is not currently supported user behavior.
- Update dropdown e2e coverage to assert dashboard Reset Zoom remains hidden, even if the internal dummy `comfyView` has stale zoom state.
- Extract shared zoom-level-to-percent conversion into a helper used by both telemetry and menu labeling.
- Add unit coverage for install-backed Reset Zoom menu visibility and menu reset dispatch.

**Working**

https://github.com/user-attachments/assets/d61e4905-ed81-4d13-8482-a4109863af77

## Testing

- `pnpm exec vitest run src/main/popups/titlePopup.test.ts`
- `pnpm exec playwright test e2e/dropdowns.test.ts --project=linux`
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` (`150` files / `2221` tests passed)

Manual verification:
- Recorded before/after video showing title-menu Reset Zoom clearing the titlebar zoom pill after the fix.

### DIFF
--- a/e2e/dropdowns.test.ts
+++ b/e2e/dropdowns.test.ts
@@ -3,8 +3,8 @@
  * targeted assertions on top of the existing dropdown smoke tests in
  * `chooser.test.ts`:
  *
- * 1. The Reset Zoom item only appears when the comfyView is at a
- *    non-default zoom level (gated branch in `buildTitlePopupMenuItems`).
+ * 1. The Reset Zoom item is absent on the chooser dashboard, even if the
+ *    internal dummy comfyView carries a non-default zoom level.
  * 2. The popup webContents doesn't accumulate listeners across opens
  *    (regression net for `EmbeddedPopupView` lifecycle drift).
  * 3. Showing the title-bar tooltip and then opening the menu hides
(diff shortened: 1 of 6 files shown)


deepme987:
superseeded by https://github.com/Comfy-Org/Comfy-Desktop/pull/1130; closing this

deepme987:
Thanks for pointing it out!

Issue #1135, opened 2026-06-17
Starter templates: nice-to-have follow-ups (model prefetch, modality-tabbed picker)


MaanilVerma:
## Summary

Two non-blocking follow-ups for the starter-templates-at-install feature. Neither gates the deploy — capturing them so they don't get lost.

## 1. Pre-download the template's assets/models

Explore prefetching a template's required models during install so the user lands on a ready-to-run canvas instead of waiting on first run.

- Worth confirming current state: model pre-download already landed (background download during install + `template-models` launch-phase progress; see the `feat(templates)` / template-downloads-tray commits). This item is to **verify it covers the asset cases we care about** and close any gaps, not necessarily net-new work.
- Open question: are there template assets *beyond* the workflow's declared models (e.g. example inputs, custom-node deps) that aren't prefetched today? If so, scope whether they can ride the same pipeline.

## 2. Categorize the picker by modality (Pablo's idea)

Restructure the template picker around output modality as top-level tabs:

- **Image | Video | 3D | Audio** as main tabs
- 2–3 curated templates per tab

This reads better than a flat list and matches how users think about what they want to make. We already have the nice **"Getting Started" UI from March** — move the picker UI in that direction (reuse rather than rebuild).

## Notes

- **Not blocking for the deploy** — both are nice-to-haves.
- Builds on the shipped starter-templates work (install-time picker, `bundledTemplateId`, background model download).

Slack - https://comfy-organization.slack.com/archives/C0B9BCUAGVD/p1781649233890769?thread_ts=1781648929.798099&cid=C0B9BCUAGVD

https://github.com/user-attachments/assets/11c76670-2bf6-4497-b377-6ef7aff3f6ed

PR #1140, opened 2026-06-17
fix(templates): place starter-template LoadImage input on disk at install

MaanilVerma:
## Summary

Starter templates with an image input opened with a **Missing Input** — the workflow pointed at a sample image that was never actually on disk. We now ship that image with the app and drop it into the install's `input/` folder so the template runs out of the box.

## Changes

**What**
- The template's `LoadImage` node references an image by filename, but nothing ever placed that file on disk — not ComfyUI core, the frontend, or the templates package. So the node opened broken.
- Templates can now declare their input image(s) (`inputAssets`), we bundle those images in the repo, and copy them into the right `input/` folder at install time — the same idea as how we already pre-download a template's models.
- The copy is best-effort: if anything goes wrong it skips the image and leaves the model download alone. ComfyUI's missing-input prompt stays the safety net.

**Breaking**
- None. No IPC, telemetry, or model-download changes; templates without an input image are untouched.

## Review Focus

- **`resolveInputDir` in `templateInputAssets.ts`** is the one thing to scrutinize — it has to point at the same `input/` folder ComfyUI reads at launch (shared vs per-install). It mirrors the resolution in `actions.ts`; a mismatch would copy to a dir ComfyUI ignores.
- Scope is the input-image gap only. Only TripoSplat needs an image today; a future image-input template means committing its image + listing it.

## Testing

Unit-only by design — a main-process file copy with no UI change. Tests cover the copy, idempotent re-runs, the never-throw contract, and all three input-dir resolution branches against a real temp dir.

- `templateInputAssets.test.ts` — 10 tests.
- `pnpm typecheck` ✓ · `pnpm lint` ✓ · `pnpm test` → **2340 passed**.
- Manual before merge: fresh TripoSplat install → confirm the image lands in `input/` and the node renders.

## Screen Recording

https://github.com/user-attachments/assets/cd2fafb2-bd11-4018-a2bf-aed8103eff4f

related to #1135


deepme987 (changes requested):
Let’s discuss this further. Not a huge fan of adding another source for us to maintain

MaanilVerma:
Totally fair concern,  I dug into this before changing anything, and the answer's a bit non-obvious.

The natural fix would be "don't vendor it, just read the image from the installed templates package." 

**Problem**: the pip package doesn't ship input images. The installed comfyui_workflow_templates packages contain only the workflow JSONs + .webp previews — I checked the dist-info RECORD and there's zero /input/ content. The sample images live only in the GitHub source repo, which isn't what pip installs. That omission is actually *why* the LoadImage node opens as a Missing Input in the first place.

So there's no upstream on-disk copy to point at. The bytes have to come from one of two places:
  1. our repo (vendored — what the PR does), or
  2. a network fetch from GitHub at install.

I went with 
(1): one  PNG, because it's the only option that's offline-safe and works on first boot. 
(2) removes the maintained file but adds a network call for a static asset, which we'd deliberately avoided elsewhere.

The real cost of vendoring is drift — if upstream renames the image, our copy goes stale. I can make that safe: a tiny test asserting our bundled filename matches the LoadImage reference in the template JSON, so an upstream rename turns CI red instead of silently breaking users. The runtime path is already best-effort (a missing/mismatched image just skips → ComfyUI's missing-input prompt; never crashes the install).

Longer term the right home is upstream — the templates package should ship its input/ assets (or core/frontend should copy them on template open). Happy to file that issue. Until then we're patching a gap they left.

So: keep the one vendored file + add the drift guard, or switch to the network fetch — your call. Leaning vendored-with-guard.

@deepme987

PR #1143, opened 2026-06-17
feat: signal enable_telemetry to ComfyUI on opted-in standalone installs

### LINKED ISSUE #1142: ComfyUI needs support for a enable-telemetry feature flag, akin to show-signin-button
Default is off; this is needed to signal frontend to enable telemetry.

Kosinkadink:
Passes `--feature-flag enable_telemetry=true` to the spawned ComfyUI **only when** the install is `standalone` **and** the user opted into telemetry (`telemetryEnabled === true`). Nothing is passed otherwise (default off), and portable / git-manual installs never receive it. Injection stays gated by the running ComfyUI's `--list-feature-flags` registry, so older versions that don't know the flag won't get it.

`desktopFeatureFlags` is now a pure `(inst, telemetryEnabled)` helper with colocated unit tests covering each gating case.

Requires the matching ComfyUI registry entry (Comfy-Org/ComfyUI).

- Fixes #1142

### DIFF
--- a/src/main/lib/ipc/sessionActions/launch.ts
+++ b/src/main/lib/ipc/sessionActions/launch.ts
@@ -20,7 +20,7 @@ import {
   getComfyFeatureFlagRegistry,
   _broadcastToRenderer,
 } from '../shared'
-import type { ChildProcess, LaunchCmd } from '../shared'
+import type { ChildProcess, InstallationRecord, LaunchCmd } from '../shared'
 import { displayLaunchUrl } from '../../cloudUrl'
 import type { ModelPathsOptions } from '../../models'
 import type { ActionContext, ActionResult } from './types'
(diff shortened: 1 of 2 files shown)


src/main/lib/ipc/sessionActions/launch.ts
```diff
@@ -240,7 +251,10 @@ export async function handleLaunch({ event, installationId, inst: instArg, actio
         const desktopFlagArgs: string[] = []
         if (schema.knownFlags.has('feature-flag') && schema.knownFlags.has('list-feature-flags')) {
           const registry = await getComfyFeatureFlagRegistry(launchCmd.cmd, mainPyAbs, launchCmd.cwd, installationId, version)
-          for (const [key, value] of Object.entries(DESKTOP_FEATURE_FLAGS)) {
+          const flagEntries = Object.entries(
+            desktopFeatureFlags(inst, settings.get('telemetryEnabled') === true)
```
deepme987, line 255: I would default false instead of true.

deepme987 (approved):

PR #1158, opened 2026-06-22
fix(updater): gate auto-install on the autoInstallUpdates toggle

deepme987:
## Problem

Turning off **Automatically install Desktop updates** (`autoInstallUpdates`) did not stop updates. The toggle only swapped the update pill's presentation (silent `ready` vs an `available` prompt); it never gated the updater, so a discovered update still installed on the next restart. The two paths:

- **Windows (startup-install is the default):** `evaluateStartupInstall()` applied a staged update at the next launch without consulting the toggle.
- **macOS / Linux:** electron-updater's `autoInstallOnAppQuit` stayed at its default (`true`), so a staged update installed on quit.

Cleanest repro: with the toggle **on**, let an update download in the background; turn the toggle **off**; restart. The staged update still installed.

## Fix

Gate both auto-install paths on `isAutoInstallEnabled()`:

- `evaluateStartupInstall()` now returns a new `auto_install_off` skip when the toggle is off, so a staged update is left in place instead of being applied at launch (also suppresses the "Updating…" splash).
- `register()`'s opted-out path and `notifyAutoUpdateChanged()` set `autoInstallOnAppQuit` from the preference via a new `syncInstallOnQuitToPreference()` helper. The quit handler re-reads the flag at quit time, so toggling off also disarms an update that was downloaded while the toggle was on.

The always-on check loop and background download are unchanged: an update still stages so the `ready` pill offers a one-click install. It just no longer installs automatically when the toggle is off.

## Test plan

`src/main/lib/updater.test.ts` (added):
- `register()` disables install-on-quit when auto-install is off (opted-out path)
- toggling auto-install off after `register()` disarms install-on-quit, and back on re-arms it (stale-download case)
- startup install is skipped when auto-install is off, even with a pending staged update

All 36 updater tests pass; full `typecheck` (node/web/e2e/integration) and `lint` clean.

### DIFF
--- a/src/main/lib/updater.test.ts
+++ b/src/main/lib/updater.test.ts
@@ -380,9 +380,50 @@ describe('startup update install + session-end guard (issue #1065)', () => {
     settingsStore['installUpdatesOnStartup'] = false
     const updater = await import('./updater')
     updater.register()
-    // Opted out: electron-updater's install-on-quit stays armed; it's only
-    // suppressed when the OS session ends (see suppressInstallOnQuit).
+    // Opted out with auto-install on (the default): install-on-quit stays armed;
+    // it's only suppressed when the OS session ends (see suppressInstallOnQuit).
     expect(electronUpdaterMock.autoInstallOnAppQuit).toBe(true)
+  })
+
+  it('register() disables install-on-quit when auto-install is off (opted out of startup install)', async () => {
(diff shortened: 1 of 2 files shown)


Kosinkadink:
solved by https://github.com/Comfy-Org/Comfy-Desktop/pull/1127

PR #1196, opened 2026-06-26
feat(terminal): enable frontend terminal + logs injection on Portable + Git installs

### LINKED ISSUE #1126: Enable frontend terminal + logs injection on Portable + Git Manual installs
The same type of frontend is shipped there, so should be safe to enable.

Kosinkadink:
## Summary

Closes #1126.

The stopgap **Terminal** bottom-panel tab injection (and the Terminal/Logs pop-out buttons) was gated to `standalone` installs only. Portable and Git (manual) installs ship **the same served ComfyUI frontend** and already expose everything the injection needs:

- the `window.__comfyDesktop2.Terminal` / `.Logs` bridges (`comfyPreload.js` is attached to every install-backed comfyView),
- a working per-install PTY - both `portableSource` and `gitSource` define `getTerminalEnv` and the desktop Settings *Console* tab already uses it,
- the per-install logs broadcast (`appendLog` is keyed by `installationId`, not source type),
- the `supports_terminal` feature flag (injected for all installs).

So enabling the injection there is safe, and the existing JS dedupe guard prevents a duplicate when a newer ComfyUI ships the native `command-terminal` tab.

## Changes

- `attach.ts`: broaden the inject gate from `sourceId === 'standalone'` to a `TERMINAL_INJECTION_SOURCE_IDS` allowlist (`standalone`, `portable`, `git`). Legacy v1 `desktop` (external launch) and remote/cloud installs remain excluded.
- Refresh the now-stale "standalone only" doc comments in `attach.ts` and `comfyTerminalContentScript.ts`.

## Testing

- `pnpm run typecheck` ?
- `pnpm run lint` ?
- `pnpm run build` ?
- `pnpm run test` ? (176 files / 2644 tests)

### DIFF
--- a/src/main/host/attach.ts
+++ b/src/main/host/attach.ts
@@ -29,6 +29,12 @@ import type { ComfyWindowEntry } from './registry'
 
 const APP_VERSION = getAppVersion()
 
+/** Local managed source types that back the served frontend's terminal tab
+ *  with a per-install shell (a defined `getTerminalEnv` or the standalone
+ *  default). These get the stopgap Terminal-tab injection; remote/cloud and
+ *  external (legacy v1 desktop) installs do not. */
+const TERMINAL_INJECTION_SOURCE_IDS = new Set(['standalone', 'portable', 'git'])
+
 /** Entry point that triggered a zoom reset, tagged as `source` on the
  *  `comfy.desktop.zoom.reset` telemetry event. `titlebar` (the zoom pill)
  *  and `menu` (the title menu's "Reset Zoom") flow through the per-install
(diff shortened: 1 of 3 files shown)


Kosinkadink:
### Code review (self-review + oracle pass)

**Verdict: ship.** The core change is a minimal gate broadening with all dependencies already present for portable/git; one real gap surfaced during review and was fixed in 13b171a2.

**Verified safe**
- `installation.sourceId` is typed `string` (not an enum), so the `Set<string>` allowlist is correct; unknown source ids just fail `.has(...)`.
- Portable + git installs reach `onDomReady` with `comfyPreload.js` attached and `isLocal=true` (fresh windows + in-place partition rebuilds both set the preload). Portable is Windows-only, which matches its `python_embeded` layout.
- Legacy v1 `desktop` (external launch) returns before host attach and isn't in the set; remote/cloud are `isLocal=false`. Both stay excluded.
- `getTerminalEnv` returning `{}` / cwd-only degrades to a plain shell (no broken standalone-env / pip alias) via the `defaultTerminalEnv` bypass in `terminal.ts`.
- No other terminal/logs path was gated on `sourceId === 'standalone'` - console visibility is `hasConsole`/category based and the terminal/logs IPC is source-agnostic.

**Fixed during review (13b171a2)**
- `alreadyHasTerminalTab` only inspected `app.extensions[].bottomPanelTabs` + a hypothetical registry, but the frontend registers the native `command-terminal` tab **directly into the bottom-panel store** (`app.extensionManager.bottomPanel.panels.terminal.tabs` - the same shape `hasLogsTab` reads). A frontend shipping the native tab could therefore get a duplicate Terminal tab. Extracted a shared `bottomPanelHasTab(app, id)` helper and now check the store for `command-terminal` too. Pre-existing latent issue for standalone; broadening to portable/git widened the exposure, so fixed here.

**Checks:** `typecheck` ? `lint` ? `build` ? `test` ? (176 files / 2645 tests, +1 new dedupe test).

MaanilVerma (approved):

Kosinkadink:
@coderabbitai review

Kosinkadink:
@coderabbitai full review

PR #1197, opened 2026-06-26
Strip ANSI color codes from plain-text ComfyUI logs

### LINKED ISSUE #1024: ComfyUI now has colored logs; we do not suppor them properly and print out the magic color characters instead

Kosinkadink:
## Problem

Fixes #1024.

ComfyUI core recently added a `ColoredFormatter` (`app/logger.py`) that wraps **every** log line in ANSI escape codes. It does this unconditionally - there is no `isatty()`, `NO_COLOR`, or `FORCE_COLOR` check - so even when ComfyUI's stdout is a pipe (always, under the launcher) it emits color codes.

The launcher's interactive terminal (xterm.js, via the `terminal-output`/node-pty path) renders those fine. But the **plain-text** log surfaces render the raw escape sequences as literal `magic` characters:

- the progress modal (`ProgressModal.vue`)
- the finished/crashed-state logs (`BrandFinishedSurface.vue`)
- the pop-out logs window

A `stripAnsi` helper already existed and was already applied to the on-disk log file, the launch progress tracker, and the stdout/stderr parsers - it just wasn't applied to the live user-facing stream.

## Fix

- **`makeSendOutput` (`shared.ts`)** - strip ANSI at this single choke point, which feeds both the live `comfy-output` IPC and the logs ring buffer. This covers every plain-text consumer across all ops (launch/copy/migrate) in one place. The xterm.js terminal is untouched.
- **Crash stderr tail (`launch.ts`)** - strip ANSI from the buffer that backs the crashed-state UI message and the telemetry `last_stderr`/`boot` payloads.
- **Dedup** - the main launch path had an inline `sendOutput` duplicating `makeSendOutput` (without the strip or the destroyed-sender guard); collapsed onto `makeSendOutput`.
- **Test** - added a regression test locking in ComfyUI's exact `ColoredFormatter` output shape.

## Notes

- Rendering colors (ansi?html) was considered and rejected: these views are plain-text `{{ }}` bindings, it adds an XSS surface, and stripping is the pattern the codebase already uses for the on-disk log.
- `stripAnsi` is regex-per-chunk; an escape split across two stdout chunks could slip through, but this matches the existing accepted behavior everywhere else (log file, taps) - no new risk.

## Verification

`pnpm run typecheck`, `lint`, `build`, and `test` (2645 tests) all pass.

### DIFF
--- a/src/main/lib/ipc/sessionActions/launch.ts
+++ b/src/main/lib/ipc/sessionActions/launch.ts
@@ -461,7 +461,9 @@ export async function handleLaunch({ event, installationId, inst: instArg, actio
     })
     proc.stderr?.on('data', (chunk: Buffer) => {
       const text = chunk.toString('utf-8')
-      stderrBuf += text
+      // Strip ANSI before buffering: this tail feeds the crashed-state UI and
+      // telemetry, both of which want clean text rather than raw color codes.
+      stderrBuf += stripAnsi(text)
       if (stderrBuf.length > 8192) stderrBuf = stderrBuf.slice(-4096)
       writeLog(logStream, text)
       sendOutput(text)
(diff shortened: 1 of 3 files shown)


Kosinkadink:
Addressed in e8b9f54: hoisted a single `stripAnsi(text)` into `const clean` in the stderr handler, reused by both the tail buffer and `tracker.ingest` so they can't drift and we skip the redundant regex pass.

MaanilVerma (approved):

PR #1203, opened 2026-06-26
Make Add Existing Install detect standalone/portable in nested ComfyUI folder (#1118)

### LINKED ISSUE #1118: Make Add Existing Install more robust
Users adding an existing Standalone might point at the ComfyUI folder within the standalone install's dir; currently, we detect this as a 'git manual' instead of their standalone. We should look up one directory in this case for the standalone structure (which we have a marker for too I think).

We should make sure we do the same for Portable if we don't currently do that.

Kosinkadink:
Fixes #1118.

## Problem

**Add Existing Install** probes only the exact directory the user picks. Standalone and portable installs nest ComfyUI inside a root:

```
Standalone <root>/                 Portable <root>/
├── standalone-env/   (marker)     ├── python_embeded/   (marker)
├── manifest.json     (marker)     └── ComfyUI/
└── ComfyUI/                           ├── main.py
    ├── main.py                        └── .git/  (sometimes)
    └── .git/  (git false-positive)
```

When a user points at the inner `ComfyUI/` folder, the standalone/portable markers live one level up, so those probes miss and the inner `.git` makes it match the **git** source — it's reported as a *git manual* install.

## Fix

- **Standalone** (`standalone/install.ts`): new `findStandaloneRoot` resolves upward (picked dir or its parent), runs all checks against the root, and returns `installPath: root`.
- **Portable** (`portable.ts`): `findPortableRoot` now also checks one level **up** (in addition to self + one level down); the probe returns `installPath: root`.
- **Renderer** (`TrackModal.vue` + `ProbeResult` type): `handleSave` records the probe-resolved `installPath` when present, falling back to the picked path. This matters because standalone runtime paths resolve off `installPath` (e.g. `installPath/standalone-env/python.exe`), so it must be the real root.

Because `standalone` and `portable` come before `git` in the sources array, they now win as the default detected type; git remains selectable as a fallback.

**Existing tracked installations are untouched** — this only changes how new probes report and what installPath they record.

## Tests
- `install.probe.test.ts` / `portable.probe.test.ts`: detect root, detect when pointed at nested `ComfyUI/` (and, for portable, the parent), correct `installPath`, null for unrelated dirs.
- `TrackModal.test.ts`: save records the probe-resolved root; falls back to picked path when absent.

`typecheck` / `lint` / `build` / `test` (2673 tests) all pass.

### DIFF
--- a/src/main/sources/common/nestedRoot.ts
+++ b/src/main/sources/common/nestedRoot.ts
@@ -1,0 +1,21 @@
+import path from 'path'
+
+/**
+ * Resolve the install root when the user pointed at the nested `ComfyUI/` folder
+ * inside a managed install (standalone or portable). Returns the parent only
+ * when the picked directory is named `ComfyUI` and the parent satisfies
+ * `parentHasMarker` (the source's own layout check). Returns null otherwise.
+ *
+ * Shared by the standalone and portable probes so the ascend rule stays
+ * consistent across them.
+ */
+export function resolveNestedComfyUIParent(
+  dirPath: string,
+  parentHasMarker: (parent: string) => boolean,
+): string | null {
+  const parent = path.dirname(dirPath)
+  if (parent !== dirPath && path.basename(dirPath).toLowerCase() === 'comfyui' && parentHasMarker(parent)) {
+    return parent
+  }
+  return null
+}
--- a/src/main/sources/portable.ts
+++ b/src/main/sources/portable.ts
@@ -41,9 +41,6 @@ interface GitHubAsset {
 
 function findPortableRoot(installPath: string): string | null {
   if (fs.existsSync(path.join(installPath, 'python_embeded'))) return installPath
-  // User pointed at the nested `ComfyUI/` folder — the root is one level up.
-  const parent = path.dirname(installPath)
-  if (parent !== installPath && fs.existsSync(path.join(parent, 'python_embeded'))) return parent
   let entries: fs.Dirent[]
   try {
     entries = fs.readdirSync(installPath, { withFileTypes: true })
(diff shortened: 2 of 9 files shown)


Kosinkadink:
@coderabbitai review

Kosinkadink:
@coderabbitai full review

Kosinkadink:
@coderabbitai full review

src/main/sources/portable.ts
```diff
@@ -57,6 +57,25 @@ function findPortableRoot(installPath: string): string | null {
...
+  const parent = path.dirname(dirPath)
+  if (
+    parent !== dirPath &&
+    path.basename(dirPath).toLowerCase() === 'comfyui' &&
+    fs.existsSync(path.join(parent, 'python_embeded'))
+  ) {
+    return parent
+  }
+  return null
+}
```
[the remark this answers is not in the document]
Kosinkadink: Done in a964bb1 — extracted the shared `resolveNestedComfyUIParent(dirPath, parentHasMarker)` helper into `src/main/sources/common/nestedRoot.ts` and reused it in both `findStandaloneRoot` and `findPortableProbeRoot`, with a focused unit test.

src/main/sources/common/nestedRoot.ts
christian-byrne, line 1: Why not an infinitely nested comfyui folder as well?

christian-byrne (approved):
For identifying the standalone root, don't we have a comfy version env file that is more authoritative and would feel better?

Kosinkadink:
Tested locally, works

src/main/sources/common/nestedRoot.ts
Kosinkadink: we dont actually have a comfy version env file off rip - we create it in desktop before launch such that we can track that its a 'standalone' install in desktop. So gotta use structure.

PR #1207, opened 2026-06-26
fix(templates): show starter-template downloads in the downloads drawer

### LINKED ISSUE #1173: The downloads through the desktop-side templates are completely hidden
They should show up in the downloads drawer. Otherwise extremely confusing if you dont have fast internet (you wouldn't even know they have started downloading).

Kosinkadink:
## Problem

Fixes #1173.

Starter-template model downloads were **completely hidden** from the downloads drawer/tray. With slow internet you'd have no indication a multi-GB download had even started.

## Root cause

Template downloads run in a **separate main-process download engine** (`lib/download.ts` — resume-capable via `.dl-meta`, bounded-concurrency pool, idle-timeout watchdog, R2 mirror fallback). This separation is justified: the downloads start at *install time*, before any ComfyUI `BrowserWindow`/session exists, so they cannot use the renderer-driven Electron `DownloadItem` pipeline that feeds the drawer.

The drawer state itself, however, is plain main-process data — and the read-only bridge that reflects a template download into it (`mirrorTemplateDownloadToTray`) was only wired to the **Skip** action. So during install and any non-skip launch, nothing showed.

## Fix

Start the existing, idempotent tray mirror at **download start** instead of only on Skip. The download is now reflected into the drawer for its whole lifetime; the launch-stepper phase and the Skip flow are unchanged (the later Skip call becomes a no-op).

One-line behavioral change in `startTemplateDownload` + a regression test.

## Verification

- `pnpm typecheck` ✅  `pnpm lint` ✅  `pnpm build` ✅
- `pnpm test` — 2664 passed (incl. new `#1173` mirror test)

### DIFF
--- a/src/main/lib/comfyDownloadManager.ts
+++ b/src/main/lib/comfyDownloadManager.ts
@@ -155,6 +155,17 @@ const templateTrayMirrorByInstall = new Map<string, Map<string, DownloadProgress
  */
 export function setTemplateTrayMirror(installationId: string, entries: DownloadProgress[]): void {
   const prev = templateTrayMirrorByInstall.get(installationId)
+  // Empty set clears this install's rows. Don't keep (or create) an empty
+  // bucket: while a download is still resolving it has no files yet, and a
+  // lingering empty bucket would both leak and emit `tray-state-changed` on
+  // every poll with nothing to show.
+  if (entries.length === 0) {
+    if (!prev) return
+    for (const url of prev.keys()) createdAtByUrl.delete(url)
+    templateTrayMirrorByInstall.delete(installationId)
+    downloadEvents.emit('tray-state-changed')
+    return
+  }
   const nextUrls = new Set(entries.map((e) => e.url))
   if (prev) {
     for (const url of prev.keys()) {
--- a/src/main/lib/ipc/registerInstallationHandlers.ts
+++ b/src/main/lib/ipc/registerInstallationHandlers.ts
@@ -45,6 +45,7 @@ import {
   startTemplateDownload,
   abortTemplateDownload,
   requestSkipTemplateDownload,
+  stopTemplateTrayMirror,
 } from '../../sources/standalone/templateDownloadTask'
 import { recordIpcInvocation } from '../e2eOverrides'
 
@@ -339,8 +340,10 @@ export function registerInstallationHandlers(): void {
       } catch (err) {
         _operationAborts.delete(installationId)
         // Install failed or was cancelled — tear down the background template
-        // download too (the models would have nowhere to land).
+        // download too (the models would have nowhere to land) and drop its
+        // tray rows, since no comfy window will ever attach to clean them up.
         abortTemplateDownload(installationId)
+        stopTemplateTrayMirror(installationId)
         if (abort.signal.aborted) {
           if (isComfyUpdate) {
             mainTelemetry.emit('comfy.desktop.comfyui.update.applied', {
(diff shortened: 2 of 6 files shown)


Kosinkadink:
## Code review (self-review, GPT-5.5 oracle assisted)

The core fix is sound: starting the existing idempotent `mirrorTemplateDownloadToTray` at download start is the right boundary (separate *engine*, shared *display*). The review surfaced lifecycle edge cases that only become reachable once the mirror runs for the whole lifetime instead of just after Skip. All addressed in 72296430:

| # | Severity | Finding | Fix |
|---|----------|---------|-----|
| 1 | should-fix | A cancelled/errored task left unfinished files mapped to `downloading` → counted **active forever** by `getDownloadsTrayState` (e.g. install abort, disk-space pre-flight error). | `templateStateToTrayEntries` now maps unfinished files to the task's terminal status (cancelled/error). |
| 2 | should-fix | Install failure/cancel only called `abortTemplateDownload`; mirror rows leaked when no comfy window ever attached to run `stopTemplateTrayMirror`. | Call `stopTemplateTrayMirror` in the install catch path. |
| 3 | should-fix | `setTemplateTrayMirror([])` during the resolving phase (no files yet) created/kept an empty bucket and churned `tray-state-changed` every 500 ms. | Empty entry set is now a true clear/no-op. |
| 4 | should-fix (concurrency) | Synthetic url `template-model://<dir>/<name>` collides across concurrent installs of the same template (global `createdAtByUrl` + renderer store keyed by url). | Install-scope the url prefix. |

### Verified safe / out of scope
- `requestSkipTemplateDownload` remains a true no-op once the start-time mirror is active; the skip gate still releases correctly.
- Synthetic urls don't collide with real HTTP downloads.
- Broadcasting progress for template downloads is not a permissions concern (user opted into template model downloads).
- **Known pre-existing UX gap (not changed here):** the drawer's cancel/retry controls are shown for `template-model://` rows but route to `cancelModelDownload`/`retryDownload`, which return `false` for mirror rows (no-op). This predates this PR (the Skip path already created such rows). Worth a follow-up to either hide controls or wire cancel → `abortTemplateDownload`.
- Resolving / input-asset phase is still rowless (acceptable for a model-download-focused fix).

### Checks
`pnpm typecheck` ✅  `pnpm lint` ✅  `pnpm build` ✅  `pnpm test` → 2665 passed (added: terminal-aware mapper test + start-time mirror test).

Kosinkadink:
@coderabbitai review

MaanilVerma:
Are you sure this is an issue ? If you download something from starter templates we already open the download section + show the downloads there as well. 
@Kosinkadink 


https://github.com/user-attachments/assets/16e1e3e5-bfca-4975-bf9f-34be97218f2b

christian-byrne (approved):
LGTM

Kosinkadink:
@coderabbitai review

Kosinkadink:
> Are you sure this is an issue ? If you download something from starter templates we already open the download section + show the downloads there as well. @Kosinkadink
> 
>  Screen.Recording.2026-06-26.at.9.59.59.PM.mov

The downloads dont show in the drawer until you actually get into the ComfyUI instance - the downloads happen throughout the whole installation process up to that point. I have gigabit internet so the downloads finish before the screen transitions to ComfyUI actually being booted where the downloads drawer actually picks it up. This PR fixes that. 
image

MaanilVerma:
Oh, okay, understood! Yeah, when we originally implemented this, we decided to only show model download progress in the logs. If the download finished before entering the instance, we decided keeping it just in the logs was fine. But you're right, it's definitely nice to have it in the downloads tab before transitioning to the instance. @Kosinkadink

PR #1243, opened 2026-07-10
Reduce desktop update polling cadence

benceruleanlu:
## What changed

- preserve the update check shortly after startup and all manual check paths
- replace 10-minute polling with a recursive 6-hour cadence plus ±30 minutes of jitter
- check after system resume when the last update check is at least 6 hours old
- rearm the background timer after any check and avoid overlapping background checks
- add focused deterministic tests for startup timing, jitter bounds, cadence, manual rearming, and resume behavior

## Why

A continuously running client currently checks for Desktop updates 144 times per day. This reduces background traffic to roughly four checks per day while retaining prompt startup/manual checks and preventing clients from synchronizing.

The existing auto-install opt-out behavior is unchanged.

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` (188 files, 2,770 passed, 1 skipped)

### DIFF
--- a/src/main/lib/updater.test.ts
+++ b/src/main/lib/updater.test.ts
@@ -6,6 +6,7 @@ let mockAppImage: string | undefined
 let mockIsPackaged = true
 let mockExePath = '/opt/Comfy Desktop/comfyui-desktop-2'
 let mockAppVersion = '1.0.0'
+let powerMonitorListeners: Record<string, Array<() => void>> = {}
 
 vi.mock('electron', () => ({
   app: {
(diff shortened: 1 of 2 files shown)


Kosinkadink:
Does this solve some issue that was reported? Currently, within 48 hours of an update releasing, ~90% of active users are on the latest version of desktop. I would wager one reason for this is that the 144 checks a day create a lot of opportunities for the update prompt to pop up while users are using the app.

benceruleanlu:
Sure.

PR #1249, opened 2026-07-14
Complete Desktop auth and PostHog identity stitching (GTM-93)

deepme987:
## Summary

This is now the single Desktop PR for GTM-93. It restores the reverted login-code implementation from #1222, keeps the useful hardening from the original #1249, and incorporates the complete Desktop identity-stitching work from #1272.

After this lands, the intended journey is one PostHog person:

`comfy.org anonymous visit -> Desktop download/install/open -> Firebase sign-in`

## What is included

- Cloud desktop-login-code authentication with a Desktop-held PKCE verifier and one-time Firebase custom-token exchange.
- Session injection back into the exact local, preview, or Cloud origin that initiated sign-in, independently from the production Cloud login/backend origin.
- The `Telemetry.reportFirebaseAuthState` preload bridge and declarative signed-in/signed-out consensus across live Cloud views; active hosted views bind only after restored renderer consensus, while local/legacy views retain a main-verified fallback.
- Website anonymous-ID adoption from the installer carrier on fresh Windows per-user installs, while keeping `installation_id` as a stable property rather than a PostHog distinct ID.
- Consent-gated capture and identity transitions, logout/account-switch safety, conflicted multi-window isolation, and durable anonymous-ID rotation.
- A durable pending-merge queue that records W/D -> Firebase UID before rotating the reusable anonymous ID, replays unacknowledged PostHog identifies after restart, and clears them only after the SDK reports the matching identify in a successful delivery batch.
- Focused hardening: fail-closed origin checks before refresh-token injection, strict server timing validation, bounded 408/429/5xx polling retries, sanitized failure telemetry, and retry-safe telemetry volume accounting.
- Preservation of current `main`, including #1275's `openModelAccessPage` preload bridge.

## Accepted boundaries

The Cloud contract is S256-only, so this PR verifies the verifier/challenge wiring but does not send an unmodeled `code_challenge_method` field. It also does not add number matching or manual code confirmation: the short-lived code is auto-redeemed by the signed-in browser session.

The installer carries the raw website PostHog ID. Copying the same carrier-named installer can therefore replay that analytics identity across fresh installations. This limitation is explicitly accepted for this rollout; replacing it with a single-use router redemption token is optional follow-up work, not a blocker for this PR.

## Rollout

The frontend auth-state bridge and router redemption path are prerequisites for the RC integration test. Once this PR is reviewed and merged, cut a Desktop RC and validate the full packaged flow before stable promotion. This PR supersedes #1272.

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` — 201 files passed; 3,046 tests passed, 1 skipped
- Focused auth, consensus, carrier, and telemetry suite — 180 tests passed

Closes GTM-93 after packaged RC and PostHog verification, not on merge alone.


benceruleanlu:
Since I made this unreviewable (sorry Deep) I opened a stack of PRs that keeps changes somewhat more reviewable. I would consider this PR superseded.

christian-byrne:
Superseded — the login-code flow landed via #1303 (v1.0.31), with the identity/consensus half in #1306 and follow-ups in #1353, #1356 and #1359. `desktopLoginCode/` and `reportFirebaseAuthState` are both on `main` today, and GTM-93 / GTM-277 / GTM-278 are all closed.

Per the author's own note above ("I would consider this PR superseded"), this can be closed so the remaining auth work is just the open drafts #1357 and #1358.

PR #1281, opened 2026-07-18
fix(adopt): restore Legacy Desktop settings from workspace

Kosinkadink:
## Summary

- read Legacy Desktop's `comfy.settings.json` from the canonical `<basePath>/user/default` workspace path and back up that same file
- preserve CPU installs by deriving `--cpu` from the persisted `selectedDevice` when the settings entry is absent
- repair records already adopted with generated default args before startup auto-launch, while preserving launch and path settings edited in Desktop 2.0

## Root cause

Legacy Desktop stores frontend and launch settings under its ComfyUI workspace. The adoption flow instead looked for `comfy.settings.json` in the Electron config directory, so it silently migrated an empty settings object. For CPU installations this dropped `--cpu`, then launched a CPU-only Torch environment through the CUDA path.

## Safety

- canonical workspace settings take precedence; a decoy config-directory file is ignored
- explicit `cpu: false` is preserved
- already-adopted records are changed only when their launch args still exactly match the generated defaults from the broken migration
- v2-edited input/output paths and launch args are not overwritten
- malformed or transiently unreadable settings are retried on the next startup

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` (188 files, 2812 tests)

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -2148,7 +2148,7 @@ if (app.isPackaged && !app.requestSingleInstanceLock()) {
     registerDownloadHandlers()
     registerAssetDownloadHandlers({ findInstallationIdForWindow })
     cleanupTempDownloads()
-    ipc.register({
+    await ipc.register({
       onLaunch,
       onStop,
       onComfyExited,
--- a/src/main/lib/desktopAdopt.test.ts
+++ b/src/main/lib/desktopAdopt.test.ts
@@ -859,7 +859,13 @@ describe('adoptDesktopInstall', () => {
   it('does not replace launch args edited after adoption', async () => {
     const legacy = buildFakeLegacy({
       configFiles: {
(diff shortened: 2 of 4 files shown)


MaanilVerma (approved):

deepme987 (approved):

Kosinkadink:
Follow-up from the affected user's supplied configuration files: NVIDIA was selected and `Comfy.Server.LaunchArgs` was empty, so this specific CUDA crash was not caused by losing or restoring a legacy `--cpu` setting. The settings-path fix in this PR remains valid independently. The incident's immediate mismatch was an adopted CPU-only Torch build with an NVIDIA selection; follow-up fix: https://github.com/Comfy-Org/Comfy-Desktop/pull/1287

PR #1289, opened 2026-07-22
feat: share GPU tier classifier via IPC

benceruleanlu:
## Summary

- move `deriveGpuTier` from renderer-only telemetry code into `src/shared`
- add `gpu_tier` and `gpu_vram_gb` to the `get-system-info` IPC response alongside the canonical `gpu_model`
- consume the main-process values in renderer telemetry instead of re-deriving them
- move the tier boundary coverage into shared tests

## Why

Main-process and renderer surfaces need one authoritative GPU tier classifier. Returning the derived tier, selected GPU model, and VRAM through `get-system-info` prevents surfaces from duplicating the PostHog cohort logic and drifting over time.

## Impact

GPU-aware surfaces can consume `gpu_tier`, `gpu_model`, and `gpu_vram_gb` directly from the existing system-info API. Existing telemetry retains the same tier vocabulary and boundaries.

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` — 189 files passed; 2,814 tests passed, 1 skipped

Linear: [DESK2-140](https://linear.app/comfyorg/issue/DESK2-140/d-25-shared-gpu-tier-classifier-ipc-plumbing)

PR #1291, opened 2026-07-22
feat: recommend Cloud for low-GPU first use

benceruleanlu:
## Summary

- show a compact “Recommended for your hardware” line on the first-use Cloud card only for `sub_low` and `cpu_only` tiers
- read the shared classifier through `get-system-info` when the takeover opens, while preserving the existing Express-install GPU hint
- reuse one system-info request across same-instance `open()` resets instead of repeating the full OS/CPU/GPU scan
- fail closed when hardware IPC fails, and suppress the recommendation unless Cloud capacity is `normal`
- leave neither Cloud nor Local selected for users who see the GPU recommendation, while preserving an explicit choice made before detection finishes
- keep the first enabled choice keyboard-reachable when neither radio is selected
- emit `comfy.desktop.first_use.gpu_reco_shown` with hardware/capacity context and extend `fork_chosen` with `gpu_tier` and `reco_shown`
- preserve the existing fork defaults for users outside the recommendation cohort and add English/Chinese copy
- prevent delayed hardware responses from recording exposure after the user leaves the start step or commits a choice

## Telemetry consent

Recommendation rendering is independent of telemetry consent. Only analytics emission is consent-gated: already-consented users emit immediately, first-time users emit after accepting telemetry, and opt-outs emit nothing.

## Stack

This is stacked on #1289 (DESK2-140), which adds the shared GPU classifier and `get-system-info` fields. Review this PR against its current base branch for the focused FE-1348 diff.

The design/copy ticket DES-548 is still Todo, so the recommendation text is intentionally conservative and contains no model-specific hardware claim.

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` — 189 files passed; 2,831 tests passed, 1 skipped

Screenshot 2026-07-21 at 9 29 45 PM

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -606,6 +606,7 @@
     "pickLead": "Choose the setup that fits how you work. You can switch later.",
     "cloudTagline": "Instant setup. Full power.",
     "cloudDesc": "Best for **most users** who want to work from anywhere with models verified for commercial license.",
+    "gpuRecommendation": "Recommended for your hardware",
     "localLabel": "Local",
     "localTagline": "Full control. Your hardware.",
     "localDesc": "Best for **power users** who want to manage versions, custom nodes, and offline workflows.",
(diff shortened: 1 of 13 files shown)


deepme987:
unassigning @Kosinkadink ; I can take this one

benceruleanlu:
Waiting for design, related https://github.com/Comfy-Org/Comfy-Desktop/pull/1299

benceruleanlu:
1299 is final version

PR #1301, opened 2026-07-24
feat(comfybuilder): stage distribution models before launch

james00012:
## TL;DR

A distribution archive carries only code and the environment, never model weights, so an installed distribution had none of the models it declares. This adds a model-staging phase to the comfybuilder install: after the archive extracts and before ComfyUI launches, download each declared model and place it at `<installPath>/ComfyUI/models/<type>/<filename>`, the way comfy-deploy provisions a volume before boot.

Stacked on `james/builder-ui` (#1297). The backend endpoint that will serve the model list (cloud #5528) is not deployed yet, so the manifest is mocked behind a seam for now.

## Intent / Why

The archive is `venv/` + `ComfyUI/` + `lock.txt`: code and environment only. A distribution that declares models produced an install with none of them, so a workflow needing those weights would fail at run time. comfy-deploy already stages a version's models onto a volume before starting ComfyUI; this brings the same guarantee to Desktop.

## Decision

- **Placement targets the install's own model root.** Models go to `<installPath>/ComfyUI/models/<type>/<filename>`. That root is always on ComfyUI's model search path, so a staged model is found whether or not the user shares a global model library. This keeps a distribution self-contained (a reproducible unit) and needs no change to the install record's shared-models setting. Verified on Windows: ComfyUI's own `folder_paths.get_filename_list('vae_approx')` returns the staged file.
- **Integrity mirrors archive install.** A model whose manifest carries a sha256 is verified byte for byte and a mismatch fails the install; a model with no hash installs unverified (a public source that supplied none). Each model downloads to a `.partial` sibling and is renamed into place only after it verifies, so an interrupted install never leaves a truncated model looking complete, and a re-run is idempotent (a present, verified model is skipped; a present-but-wrong one is re-fetched).
- **Path safety.** `type` and `filename` are each validated as a single safe segment (no separators, no `..`, no absolute/drive markers) before any download, so a manifest cannot traverse out of the model tree.
- **The manifest source is a seam, not a hardcode.** `resolveModelManifest` picks, in order: an env-injected manifest (`COMFY_BUILDER_MODELS_MANIFEST`, the test/e2e seam), the real endpoint under an opt-in flag (`COMFY_BUILDER_MODELS_LIVE=1`, which resolves the version id and calls `ComfyBuilderClient.fetchModelManifest`), else a static per-distribution mock. The client method already targets the real `/v1/distribution-versions/{id}/manifest` shape, so when cloud #5528 deploys, flipping the default is the only change and the mock is deleted.

## Illustration

```
install():
  1. installArtifact   download archive -> verify sha -> extract  (venv/ + ComfyUI/)
  2. resolveModelManifest(distributionId, version)   -> { models[], policies }
  3. stageModels:  for each model
        download -> <dest>.partial
        verify sha256 (when present)
        rename -> <installPath>/ComfyUI/models/<type>/<filename>
  -> launch  (ComfyUI finds the staged models on its search path)
```
The stepper gains a third phase, `Download models`, after `Download` and `Extract`.

## Implementation Details

- `src/main/comfybuilder/models.ts`: `stageModels` (download + verify + place, `.partial` rename, idempotent skip, path safety) + `installModelsRoot`.
- `src/main/comfybuilder/modelManifest.ts`: `resolveModelManifest` (override / live / mock) + the quarantined static mock, keyed by the two staging distribution ids, using small public models so a real install genuinely downloads.
- `src/main/comfybuilder/client.ts`: `fetchModelManifest(versionId)` GETting `/v1/distribution-versions/{id}/manifest`.
- `src/main/comfybuilder/types.ts`: `ModelDescriptor`, `ModelManifest`, `ModelPolicy`, `StageProgress`.
- `src/main/comfybuilder/install.ts`: export `sha256File` + `normalizeSha256` (reused by staging; archive path now shares the normalizer).
- `src/main/sources/comfybuilder/index.ts`: `install()` runs staging after the archive; `installSteps` gains the `models` phase.
- `src/main/lib/e2eHooks.ts`: an E2E-gated `stageDistributionModels` hook that runs the real staging isolated from the archive/auth path.
- `locales/en.json` / `zh.json`: the `Download models` step label.

## Validation

- Unit: `stageModels` (place, checksum-mismatch, path-safety x4, no-hash, idempotent skip, re-fetch-on-mismatch, progress, abort), `resolveModelManifest` (mock / inline+file override / live / no-match), `client.fetchModelManifest` wire shape. Full suite green (207 files, 2951 tests).
- E2E (`e2e/comfybuilder-models.test.ts`, `@windows @macos`): drives the REAL staging code in the real Electron app against a local HTTP server, asserting placement, checksum failure, and empty-manifest. Proven a real regression test by breaking placement and confirming it fails. Green on macOS and on a Windows VM.
- Real-remote on a Windows VM: staged the actual `taesd_decoder.pth` from GitHub via the static mock into a real distribution install; on-disk sha256 matched exactly, and ComfyUI's `folder_paths.get_filename_list('vae_approx')` returned `['taesd_decoder.pth']`, proving the model is discoverable on ComfyUI's search path.
- `typecheck` (4 projects) and `lint` clean.

## Follow-ups (not this PR)

- When cloud #5528 deploys, make `LIVE` the default and delete the static mock.
- The install record does not carry the version id (only the number); the LIVE path re-derives it via `listVersions`. Persisting `versionId` on the record would drop that extra call.

### DIFF
--- a/e2e/comfybuilder-launch.test.ts
+++ b/e2e/comfybuilder-launch.test.ts
@@ -9,9 +9,10 @@
  *
  *   1. an installed distribution tile launches (run-action `launch` plus the
  *      progress takeover) and never mounts the wizard;
- *   2. a not-yet-installed distribution exposes a DISABLED launch action
- *      carrying the not-ready message, and clicking its tile explains itself
- *      instead of bouncing into the wizard.
+ *   2. clicking a not-ready one explains itself instead of bouncing there.
+ *
+ * The per-status action contract itself lives in the plugin unit test, which
+ * pins it without needing a seeded record.
  *
  * The seeded venv python exits immediately, so the launch is attempted and
  * then fails at the boot wait. Attempted-vs-never-attempted is the whole
@@ -31,9 +32,6 @@ import { byTestId, TID } from './support/testIds'
 /** Both steps of the new-install wizard: the surface a missing launch action
  *  used to bounce the user into. */
 const NEW_INSTALL_WIZARD = '.config-shell, .template-shell'
-
-/** Same tile-name selector `clickInstallTile` matches against. */
-const TILE_NAME_SELECTOR = '.chooser-tile:not(.chooser-tile-new):not(.chooser-tile-cloud) .chooser-tile-name'
 
 interface DistributionCase {
   id: string
@@ -58,12 +56,6 @@ const FAILED: DistributionCase = {
   name: 'desktop-4target-stg-v0192',
   status: 'failed',
   installPath: '',
-}
-
-interface ListActionShape {
-  id: string
-  enabled: boolean
-  disabledMessage?: string
 }
 
 let ctx: AppContext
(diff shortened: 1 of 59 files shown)


james00012:
I have read and agree to the Contributor License Agreement

james00012:
Folded into the layer PRs per the functionality/UI split: the model library (stageModels, manifest resolver, client) is now in the comfybuilder functionality PR #1295, and the model wiring (source plugin install(), stepper label, progress pacing) is in the UI PR #1297. No code lost: the #1297 branch tree is byte-identical to this branch's tested tree.

PR #1304, opened 2026-07-24
feat(telemetry): add anonymous identity core

benceruleanlu:
Split C of #1249.

## Summary

- Separate the installation property, anonymous Desktop UUID, and Firebase UID; durably replay pending anonymous-to-user merges.
- Keep pre-auth PostHog events anonymous and prevent person-profile creation until Firebase binding.
- Make telemetry volume accounting retry-safe and discard deferred data after explicit consent denial.
- Remove legacy device-ID aliasing and `desktop_device_id` URL propagation.
- Keep deferred person-property enrichment process-local; hardware and settings values are recomputed on the next launch.

## Scope

This PR is based directly on `main` after #1303. It intentionally does not persist deferred person-property enrichment or include the conflict-epoch state consumed by #1306.

The old download-token bridge remains unchanged here and is removed by #1305. Land C and D before the next release rather than treating C as independently releasable.

## Stack

| Split | PR | Base | Release gate |
|---|---|---|---|
| A | [#1302 — download-path extraction](https://github.com/Comfy-Org/Comfy-Desktop/pull/1302) | `main` | Merged |
| B | [#1303 — desktop login-code auth](https://github.com/Comfy-Org/Comfy-Desktop/pull/1303) | `main` | Merged |
| C | [#1304 — anonymous identity core](https://github.com/Comfy-Org/Comfy-Desktop/pull/1304) | `main` | Land with D before release |
| D | [#1305 — installer website-ID carrier](https://github.com/Comfy-Org/Comfy-Desktop/pull/1305) | C | Packaged Windows RC |
| E | [#1306 — Firebase auth consensus](https://github.com/Comfy-Org/Comfy-Desktop/pull/1306) | D | Packaged RC |

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` — 3,122 passed, 1 skipped
- Focused identity and telemetry validation — 130 passed
- Independent final-diff review: no concrete reachable-production findings

Final scope: 17 files, `+1,397/-1,181`.

PR #1303, opened 2026-07-24
feat(auth): add desktop login-code flow

benceruleanlu:
Split B of #1249.

## Summary

- Add the PKCE desktop login-code client, polling, custom-token exchange, and origin validation.
- Restructure the Firebase bridge into explicit flow-state/control/shared modules.
- Harden refresh-token injection against frame replacement, navigation, and competing sign-in sessions.
- Keep the existing main-process telemetry binding seam; split E replaces that one call with consensus.

This slice is behaviorally independent of the identity-core work in C, even though the GitHub stack is linear for clean diffs.

## Stack

| Split | PR | Base | Release gate |
|---|---|---|---|
| A | [#1302 — download-path extraction](https://github.com/Comfy-Org/Comfy-Desktop/pull/1302) | `main` | Normal |
| B | [#1303 — desktop login-code auth](https://github.com/Comfy-Org/Comfy-Desktop/pull/1303) | A | Packaged RC |
| C | [#1304 — anonymous identity core](https://github.com/Comfy-Org/Comfy-Desktop/pull/1304) | B | Normal |
| D | [#1305 — installer website-ID carrier](https://github.com/Comfy-Org/Comfy-Desktop/pull/1305) | C | Packaged Windows RC |
| E | [#1306 — Firebase auth consensus](https://github.com/Comfy-Org/Comfy-Desktop/pull/1306) | D | Packaged RC |

## Architecture
image

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` — 2,948 passed, 1 skipped
- Independent security review: no remaining actionable findings

## RC gate

Validate packaged login, fallback, cancellation/re-entry, origin navigation, and frame/session replacement before rollout.

PR #1302, opened 2026-07-24
refactor: extract model download path helpers

benceruleanlu:
Split A of #1249.

## Summary

- Extract model download-path and filename helpers from `comfyDownloadManager.ts`.
- Preserve compatibility through re-exports.
- Update the template and standalone imports with no behavior change.

## Stack

| Split | PR | Base | Release gate |
|---|---|---|---|
| A | [#1302 — download-path extraction](https://github.com/Comfy-Org/Comfy-Desktop/pull/1302) | `main` | Normal |
| B | [#1303 — desktop login-code auth](https://github.com/Comfy-Org/Comfy-Desktop/pull/1303) | A | Packaged RC |
| C | [#1304 — anonymous identity core](https://github.com/Comfy-Org/Comfy-Desktop/pull/1304) | B | Normal |
| D | [#1305 — installer website-ID carrier](https://github.com/Comfy-Org/Comfy-Desktop/pull/1305) | C | Packaged Windows RC |
| E | [#1306 — Firebase auth consensus](https://github.com/Comfy-Org/Comfy-Desktop/pull/1306) | D | Packaged RC |

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` — 2,853 passed, 1 skipped

This mechanical slice can use the normal release path.

PR #1306, opened 2026-07-24
feat(telemetry): add Firebase auth consensus

benceruleanlu:
Split E of #1249.

## Summary

- Add frame-bound Firebase auth consensus across trusted Cloud, Desktop-verified loopback, and fail-closed remote fallback reporters.
- Replace renderer bind/unbind IPC with declarative auth-state reporting.
- Persist verified loopback authorization and monitor local Firebase IndexedDB from preload.
- Route the desktop login-code seam through the process-wide consensus and publish bridge types 0.1.3.
- Fail closed across identity replacement, logout, navigation/frame races, detached hosts, persistence failures, and canceled loads.

This final slice depends functionally on B and C; D is included because the GitHub stack is linear.

## Stack

| Split | PR | Base | Release gate |
|---|---|---|---|
| A | [#1302 — download-path extraction](https://github.com/Comfy-Org/Comfy-Desktop/pull/1302) | `main` | Normal |
| B | [#1303 — desktop login-code auth](https://github.com/Comfy-Org/Comfy-Desktop/pull/1303) | A | Packaged RC |
| C | [#1304 — anonymous identity core](https://github.com/Comfy-Org/Comfy-Desktop/pull/1304) | B | Normal |
| D | [#1305 — installer website-ID carrier](https://github.com/Comfy-Org/Comfy-Desktop/pull/1305) | C | Packaged Windows RC |
| E | [#1306 — Firebase auth consensus](https://github.com/Comfy-Org/Comfy-Desktop/pull/1306) | D | Packaged RC |

## Validation

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test` — 3,078 passed, 1 skipped
- Focused identity/IPC/preload suite — 211 passed
- Independent security/lifecycle review: no remaining actionable findings

## RC gate

Validate packaged multi-window login/logout/account-switch behavior, Cloud and local/remote installs, navigation/frame replacement, and identity recovery before rollout.

### DIFF
--- a/src/main/lib/firebaseAuthIdentity.ts
+++ b/src/main/lib/firebaseAuthIdentity.ts
@@ -1,0 +1,636 @@
+import type { WebContents } from 'electron'
+import type { ComfyDesktop2FirebaseAuthState } from '../../types/comfyDesktopBridge'
+import * as mainTelemetry from './telemetry'
+import { isIllegalPostHogDistinctId, normalizeOpaqueIdentifier } from './opaqueIdentifier'
+import { isTrustedCloudUrl } from './trustedCloudUrl'
+import {
+  clearVerifiedLocalFirebaseUser,
+  isLoopbackOrigin,
+  persistVerifiedLocalFirebaseUser,
+  readVerifiedLocalFirebaseUser
+} from './verifiedLocalFirebaseAuth'
+
+interface Reporter {
+  eligible: boolean
+  active: boolean
+  localReportingAuthorized: boolean
+  localExpectedUserId: string | null
+  awaitingCommittedFrame: boolean
+  mainFrameNavigationsInFlight: number
+  provisionalFailureTerminals: Map<string, number>
+  committedFrame: FirebaseAuthFrameIdentity | null
+  recoverableState: ReporterFrameState | null
+  committedCandidate: ReporterFrameState | null
+  state: ComfyDesktop2FirebaseAuthState
+  onDidStartNavigation: (
+    details: Electron.Event<Electron.WebContentsDidStartNavigationEventParams>
+  ) => void
+  onDidFrameNavigate: (
+    event: Electron.Event,
+    url: string,
+    httpResponseCode: number,
+    httpStatusText: string,
+    isMainFrame: boolean,
+    frameProcessId: number,
+    frameRoutingId: number
+  ) => void
+  onDidFailProvisionalLoad: (
+    event: Electron.Event,
+    errorCode: number,
+    errorDescription: string,
+    validatedURL: string,
+    isMainFrame: boolean,
+    frameProcessId: number,
+    frameRoutingId: number
+  ) => void
+  onDidFailLoad: (
+    event: Electron.Event,
+    errorCode: number,
+    errorDescription: string,
+    validatedURL: string,
+    isMainFrame: boolean,
+    frameProcessId: number,
+    frameRoutingId: number
+  ) => void
+  onDestroyed: () => void
+}
+
+export interface FirebaseAuthFrameIdentity {
+  processId: number
+  routingId: number
+}
+
+interface ReporterFrameState {
+  frame: FirebaseAuthFrameIdentity
+  active: boolean
+  state: ComfyDesktop2FirebaseAuthState
+}
+
+interface MainVerifiedAuthState {
+  userId: string
+  origin: string
+  properties: Record<string, mainTelemetry.TelemetryValue>
+  propertiesApplied: boolean
+  reportedState?: ComfyDesktop2FirebaseAuthState
+  rendererMayReaffirm: boolean
+}
+
+const reporters = new Map<WebContents, Reporter>()
+const mainVerifiedStates = new Map<WebContents, MainVerifiedAuthState>()
+let requestedUserId: string | null = null
+let anonymousEpochIsUnmergeable = false
+let persistedEpochStateLoaded = false
+let epochTaintIsDurable = true
+
+function originOf(url: string): string | null {
+  try {
+    return new URL(url).origin
+  } catch {
+    return null
+  }
+}
+
+function localExpectedUserIdForUrl(url: string): string | null {
+  const origin = originOf(url)
+  return origin && isLoopbackOrigin(origin) ? readVerifiedLocalFirebaseUser(origin) : null
+}
+
+function refreshReporterAuthScope(reporter: Reporter, url: string): boolean {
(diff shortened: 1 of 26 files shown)


src/main/lib/trustedCloudUrl.ts
```diff
@@ -0,0 +1,23 @@
+const TRUSTED_CLOUD_HOSTS = new Set([
+  'cloud.comfy.org',
+  'stagingcloud.comfy.org',
+  'testcloud.comfy.org'
+])
+
+export function isTrustedCloudUrl(value: string): boolean {
```
deepme987, line 7: Solid trust boundary. https-only + rejecting username/password/port closes the credential-in-URL and proxy-port tricks, and the `*.testingcloud.comfy.org` check is correctly anchored (leading-dot `endsWith` + `!== 'testingcloud.comfy.org'` apex exclusion), so `eviltestingcloud.comfy.org` and `x.testingcloud.comfy.org.attacker.com` both fail. This is the single line the whole consensus's trust rests on — worth keeping it exact-match + anchored in any follow-up.

src/main/lib/ipc/registerTelemetryHandlers.ts
```diff
@@ -25,6 +28,26 @@ function asString(value: unknown): string | null {
   return typeof value === 'string' && value.length > 0 ? value : null
 }
 
+function firebaseAuthReporterFrame(event: IpcMainEvent): WebFrameMain | null {
```
deepme987, line 31: This is the key anti-spoof gate: auth-state reports are accepted **only from the sender's main frame** (`senderFrame.processId/routingId ≡ mainFrame`), so a malicious embedded sub-iframe can't report an identity. Combined with the payload validation (`asFirebaseAuthState`) and the reporter-authorization check downstream, an untrusted frame's report is dropped before it can reach consensus. Nicely layered.

src/main/lib/verifiedLocalFirebaseAuth.ts
```diff
@@ -0,0 +1,94 @@
...
+import { writeFileSafe } from './safe-file'
+
+const VERIFIED_LOCAL_FIREBASE_AUTH_FILE = 'verified-local-firebase-auth.json'
+const MAX_VERIFIED_LOCAL_ORIGINS = 32
+
+function verifiedLocalFirebaseAuthPath(): string {
+  return path.join(configDir(), VERIFIED_LOCAL_FIREBASE_AUTH_FILE)
+}
+
+function normalizeLoopbackOrigin(value: unknown): string | null {
```
deepme987, line 15: Good airtight scoping: because this rejects any non-loopback origin (and requires `url.origin === value`, no path/query smuggling), an external/remote origin can **never** be persisted as a verified-local binding. So the "Desktop-verified loopback" trust path is genuinely loopback-only — the exact property that keeps a remote page from claiming a local identity.

src/main/lib/firebaseAuthIdentity.ts
```diff
@@ -0,0 +1,630 @@
...
+}
+
+function clearUnmergeableEpoch(): boolean {
+  if (!anonymousEpochIsUnmergeable) return true
+  requestAnonymousIdentity()
+  if (!epochTaintIsDurable) {
+    epochTaintIsDurable = mainTelemetry.markAnonymousEpochUnmergeable()
+    if (!epochTaintIsDurable) return false
+  }
+  if (!mainTelemetry.discardUnmergeableAnonymousEpoch()) return false
```
deepme987, line 160: The privacy-critical fail-closed step. When an identity conflict can't be tainted durably, discarding the anonymous epoch here means a conflicted anonymous ID is never merged to the wrong user and can't be resurrected untainted after a restart. This is the right call over trying to keep a possibly-conflicted epoch alive — ties cleanly into #1304's taint/discard primitives.

deepme987 (commented):
**Agent Review** — focused on the trust core (the "can an untrusted frame spoof a trusted identity?" question). Excellent, security-critical work; the answer is **no** at every layer. Nothing blocks merge from my read.

**What I verified:**
- **Trust classification** — `isTrustedCloudUrl` is https-only, rejects username/password and non-empty port, exact-host allowlist + a correctly-anchored `*.testingcloud.comfy.org` check (leading-dot `endsWith` + apex exclusion, so `eviltestingcloud.comfy.org` and `…comfy.org.attacker.com` both fail). _(inline)_
- **Main is the trust authority** — renderers *report* auth state; main *decides* (authorized only if trusted-cloud URL **or** Desktop-verified loopback; else ignored, fail-closed).
- **IPC frame-binding** — the `firebaseAuthState` handler accepts a report **only from the sender's main frame** (`senderFrame` ≡ `mainFrame`), so a sub-iframe can't report; payload is validated (status enum + normalized, non-illegal `userId`) before authorization is even applied. _(inline)_
- **Verified-loopback is genuinely loopback-only** — `normalizeLoopbackOrigin` rejects any non-loopback origin, so an external origin can never get a verified-local binding; the preload monitor runs only on loopback pages and reads the actual Firebase IndexedDB (Desktop verifies persistence, not a renderer claim; >1 user → `pending`, never guesses). _(inline)_
- **Fail-closed on every race** — navigation/frame swap (`isSameFrame`/`settleFailedNavigation`), detached hosts (`isDestroyed` filter), logout (revoke), and the privacy-critical one: an unmergeable identity conflict taints and, if the taint can't persist, **durably discards the anonymous epoch** so a conflicted ID is never wrong-merged nor resurrected untainted across a restart. _(inline)_

**Bonus (not auth):** the telemetry IPC is well-hardened — `client`/`deployment` axes are host-authoritative (stripped/resolved from the sender, never trusted from payload), and values are clamped with the larger JSON ceiling gated by an explicit key allowlist (not a `_json` suffix), so a renderer can't bypass the clamp by renaming a field.

**Minor (non-blocking):** `*.testingcloud.comfy.org` preview subdomains count as "trusted cloud" for auth reporting. They're first-party ephemeral preview envs and the report only affects telemetry identity (not privilege), so it's reasonable — just flagging it's a broad-ish trust grant.

**Scope:** I deep-read the trust core (`trustedCloudUrl`, the authorization model + fail-closed structure in `firebaseAuthIdentity.ts`, `verifiedLocalFirebaseAuth.ts`, the IPC binding, the preload monitor) but not the full 630-line `reconcile()` winner resolution line-by-line, nor the ~1,100 lines of tests. The PR cites an independent security/lifecycle review + 3,078 tests.

deepme987 (approved):

PR #1307, opened 2026-07-24
feat: require explicit first-use choice on poor hardware

benceruleanlu:
## Summary

- use the shared `getSystemInfo().gpu_tier` result on the first-use Local/Cloud choice
- leave both choices unselected for `sub_low` and `cpu_only` hardware until the user explicitly selects one
- preserve explicit choices across late experiment and system-info resolution while retaining the capacity-disabled Local safeguard and initial legacy Desktop Local default
- lock Continue before awaiting system info so late poor-tier resolution safely returns to an unselected picker and repeated activation cannot duplicate telemetry or routing
- keep a no-default picker keyboard-accessible and prevent keyboard navigation from selecting Cloud while capacity is disabled

This is intentionally behavior-only. It does not add the pending Cloud recommendation design, recommendation copy, new telemetry, or any Why Cloud changes.

## Base

Rebased onto `main` at `25d77cb3` after #1289 merged.

## Validation

- `mise exec -- pnpm exec vitest run src/renderer/src/views/FirstUseTakeover.test.ts` — 39 passed

### DIFF
--- a/src/renderer/src/views/FirstUseTakeover.test.ts
+++ b/src/renderer/src/views/FirstUseTakeover.test.ts
@@ -34,8 +34,13 @@ vi.mock('../components/ModalShell.vue', () => ({
 }))
 vi.mock('../components/ChoiceCard.vue', () => ({
   default: {
+    props: {
+      selectable: { type: Boolean, default: false },
+      selected: { type: Boolean, default: false },
+    },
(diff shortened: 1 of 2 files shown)


deepme987 (approved):

benceruleanlu:
Incorporated into 1299

PR #1313, opened 2026-07-24
feat(desktop): builder distribution UI (account chip, distribution tiles, ChooserView)

james00012:
## TL;DR

The renderer/UX half of the builder distribution feature, split out of the interaction PR so it can be iterated on independently. Adds the account chip, distribution tiles, and the ChooserView surface that let a user sign in and install/launch distributions.

Stacked on the interaction PR (james/builder-ui), which provides the main-process session, IPC, and the install/launch plugin. This PR is renderer-only: Vue components, the pinia auth store, progress pacing, and locale strings. It calls the main process exclusively over the existing `window.api.comfybuilder.*` bridge and never touches the install/launch pipeline.

## Intent / Why

The interaction PR is fully functional headlessly (sign in, list, install, launch over IPC) but has no visual surface. This PR adds that surface, and keeping it separate lets the UX iterate without touching core logic or re-running the functionality review.

## Decision

- **Renderer-only.** Every file is under `src/renderer/` or `locales/`. No main-process, IPC-handler, or plugin change lives here.
- **Backed by the global auth store.** A single pinia `authStore` holds sign-in status, workspaces, and distributions, self-hydrating from the persisted session and subscribing to `onAuthChanged`, so any surface (chip, tiles) reacts in lockstep.
- **Extends the existing grid.** ChooserView already renders install tiles; this adds the account chip and distribution tiles into it, and dedupes a distribution against an install that already backs it.

## Illustration

```
ChooserView (existing grid)
  + DevPlatformAccountChip   -> window.api.comfybuilder.signIn / listWorkspaces / switchWorkspace
  + DevPlatformDistributionCard[]  <- authStore.distributions
       on click -> installInstance -> (interaction PR does the work)
```

## Implementation Details

- `stores/authStore.ts` - sign-in status, workspaces, distributions; revision-guarded fetches.
- `views/devplatform/` - DevPlatformAccountChip (sign-in + workspace switcher), DevPlatformAvatar, DevPlatformDistributionCard, tiles.css.
- `views/ChooserView.vue` - mounts the chip, renders distribution tiles, dedupes against installs.
- `lib/progressWeights.ts` - pacing entry for the distribution install steps.
- `locales/en.json`, `zh.json` - the account/workspace/step strings.

## Validation

Full stacked build is green (this is byte-identical to the pre-split #1297 tree). typecheck (4 projects), lint, unit (2952), and the distribution e2e all pass on the stacked result.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1314, opened 2026-07-24
feat(chooser): two-line install tiles — drop the launch-recency row

shrimbly:
First of four small chunks porting the chooser card design work onto the builder-ux stack. Targets `james/builder-ux` (#1313) so the UX iteration stays off the functionality review.

## What

Install tiles lose the third line — the `"Launched 3h ago"` / `"Not launched yet"` recency row. The meta facts and the update/migrate pill now share a single row.

## Why

Recency is already encoded in the grid's ordering (tiles sort by `lastLaunchedAt` desc, never-launched last), so the line restates what position already says. Removing it shortens every tile and gives the install name more room to breathe before truncating.

## Scope

Deliberately touches no distribution code — `ChooserInstallTile.vue` and `chooser-tiles.css` are byte-identical between this branch's base and where the design work started, so this lands clean and can be reviewed without builder-stack context.

## One call for the reviewer

`useInstallList.lastLaunchedLabel` loses its only consumer here. I left it in place — it still has five passing tests, and deleting a tested composable export felt like a decision for you rather than a drive-by. Say the word and I'll remove it with its tests.

## Testing

`ChooserView.test.ts` + `useInstallList.test.ts` pass (43 tests). The recency-line test is rewritten to pin the new contract: no recency row, meta line retained.

### DIFF
--- a/e2e/comfybuilder-launch.test.ts
+++ b/e2e/comfybuilder-launch.test.ts
@@ -1,0 +1,163 @@
+/**
+ * E2E: launching a ComfyBuilder distribution from the chooser.
+ *
+ * The renderer resolves launch ONLY from `get-list-actions`
+ * (`performChooserLaunch`), so a source plugin without `getListActions`
+ * hands it an empty array, which reads as "this install cannot launch" and
+ * bounces the tile click into the new-install wizard. This file pins the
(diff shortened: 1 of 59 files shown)


shrimbly:

PR #1315, opened 2026-07-24
feat(chooser): kebab menu on distribution cards

shrimbly:
Second of four small chunks porting the chooser card design work onto the builder-ux stack. **Stacked on #1314** — review that one first; this PR's diff is against it.

## What

Distribution cards get the same always-present top-right kebab as install tiles, with an Install item wired through the existing `ContextMenu`.

## Why

The cards already share the grid, the box, the type scale and the `TransitionGroup` with install tiles. The corner was the one place they diverged: install tiles put a kebab there, distributions put a state pill and nothing else. Same position, different meaning, one grid.

Install is the only action a distribution supports today, so that's the only item. Blocked distributions keep the kebab and show Install **disabled** rather than opening an empty menu — an empty menu reads as a bug, and the blocking reason is already on the card's own tag.

## Notes

- Adds `devPlatform.distribution.menuInstall` to `en.json` / `zh.json`.
- The existing pills stay where they are. Reshuffling the card's top-right vocabulary is the next chunk's job, kept separate so this one is easy to revert.
- The `ContextMenu` test stub now declares its props, so tests can assert what the view handed the menu without rendering the real one.

## Testing

`ChooserView.test.ts` passes (21 tests). Two new: the kebab opens a menu whose Install item runs the same `installDistribution` flow as activating the card, and a blocked distribution keeps its kebab with Install disabled.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -1141,6 +1141,7 @@
     },
     "distribution": {
       "availablePill": "Available",
+      "menuInstall": "Install",
       "updatedLabel": "Updated",
       "installTileMeta": "Not installed yet",
       "emptyTitle": "No distributions published to this workspace yet.",
--- a/locales/zh.json
+++ b/locales/zh.json
@@ -1141,6 +1141,7 @@
     },
     "distribution": {
       "availablePill": "可安装",
+      "menuInstall": "安装",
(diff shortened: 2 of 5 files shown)


shrimbly:

PR #1316, opened 2026-07-24
feat(chooser): one footer grammar for distribution cards

shrimbly:
Third of four small chunks porting the chooser card design work onto the builder-ux stack. **Stacked on #1315 → #1314.**

## What

Distribution cards adopt the same footer grammar as install tiles: name, then **identity facts left, one status/action slot right**.

## Why

#1314 took install tiles to two lines, which left distribution cards as the only three-line tile in the grid. This closes that gap and settles what the two families put where.

Three consequences fall out:

1. **The version is labelled** — `Dist v2` rather than `2`. These cards share a grid with install tiles whose version *is* the ComfyUI version; a bare `2` beside a `0.3.20` invites misreading.
2. **Blocked state leaves the corner.** #1315 gave the corner to the kebab, so the blocked tag moves to the footer's right slot and the full reason moves to the tile's `title` — costs no space, still available on hover.
3. **The recency row goes** (`"Not installed yet"` / `"Updated 3h ago"`), along with the two locale keys that fed it.

The rule the slot follows: pills say what you *can do*, tags say what *is so*, and the two never render together. State never displaces the facts.

## What I did not port

Our prototype showed **two** labelled versions here — the bundled ComfyUI version alongside the distribution version. That reads `distribution.comfyuiVersion`, which turned out to exist only in the renderer mocks; no backend on either stack populates it. Wiring it up means touching the distribution row mapping in `src/main/devplatform/distributions.ts`, which is #1297's territory, so I left the slot open rather than shipping a field that always renders empty. The facts line takes a second fact without changes when it lands.

## Testing

`ChooserView.test.ts` passes (24 tests). Three new: the version renders labelled, a blocked distribution puts its reason in the footer slot with an empty kebab corner and a `title` explanation, and no card ever renders both an action pill and a state tag.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -1142,8 +1142,7 @@
     "distribution": {
       "availablePill": "Available",
       "menuInstall": "Install",
-      "updatedLabel": "Updated",
-      "installTileMeta": "Not installed yet",
+      "distVersion": "Dist v{version}",
       "emptyTitle": "No distributions published to this workspace yet.",
       "installFailed": "Could not start the install. Try again.",
       "states": {
--- a/locales/zh.json
+++ b/locales/zh.json
@@ -1142,8 +1142,7 @@
     "distribution": {
       "availablePill": "可安装",
       "menuInstall": "安装",
-      "updatedLabel": "更新于",
-      "installTileMeta": "尚未安装",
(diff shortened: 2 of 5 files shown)


shrimbly:

PR #1317, opened 2026-07-24
feat(chooser): name the action on installable distributions

shrimbly:
Last of four small chunks porting the chooser card design work onto the builder-ux stack. **Stacked on #1316 → #1315 → #1314.**

## What

An installable distribution's right slot changes from a quiet `Available` tag to the blue `Install` pill — the same pill Update already wears.

## Why

`Available` names a state on a card whose entire purpose is an action. Clicking anywhere on the card installs it, and Update uses a blue pill for precisely that gesture, so Install should look the same. Under #1316's grammar, pills say what you can do and tags say what is so — `Available` was sitting on the wrong side of that line.

The quiet tag stays for the cases you genuinely can't act on: `Installed`, and the three blocked reasons.

## The call this PR is really asking you to make

This is deliberately its own PR because it's a judgement call and I don't think it's clear-cut.

**Worth knowing:** `ChooserView.vue:141` filters out any distribution backed by an installation, so today *every* distribution card is `installable`. This PR therefore turns every one of them blue. The design it came from assumed installed and uninstalled distributions sat side by side, where the pill distinguished within the family — that's not the current model.

**The case against merging:** when a whole family shares one state, a high-contrast pill on all of them carries no information. Blue is otherwise reserved for "something needs your attention" (Update, Migrate) on install tiles, and being installable isn't that.

**The case for:** the pill isn't distinguishing distributions from each other, it's distinguishing "things you can add" from "things you have" in a mixed grid — which is the question a user actually has while scanning it.

I lean toward merging, but closing this PR is a perfectly good answer and costs nothing: #1316 stands on its own with `Available` intact.

## Also held back

The prototype rendered uninstalled distributions as **hollow** tiles (transparent, dashed border). That was a contrast device against installed siblings, so it depends on the same question — under the current dedupe every card would be hollow, which makes it decoration. Worth revisiting alongside whether an installed distribution keeps its distribution identity on the chooser.

## Testing

`ChooserView.test.ts` passes (25 tests). One new: an installable distribution's slot reads `Install`.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -1140,7 +1140,6 @@
       "switchLabel": "Workspace"
     },
     "distribution": {
-      "availablePill": "Available",
       "menuInstall": "Install",
       "distVersion": "Dist v{version}",
       "emptyTitle": "No distributions published to this workspace yet.",
--- a/locales/zh.json
+++ b/locales/zh.json
@@ -1140,7 +1140,6 @@
       "switchLabel": "工作区"
     },
     "distribution": {
-      "availablePill": "可安装",
       "menuInstall": "安装",
       "distVersion": "分发版 v{version}",
       "emptyTitle": "此工作区尚未发布任何分发版本。",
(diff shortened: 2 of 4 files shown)


shrimbly:

PR #1337, opened 2026-07-29
feat(chooser): read the version delta off an updatable tile

shrimbly:
Closes DES-568.

## Problem

When an update exists, the tile shows a bare **Update** pill and hides the target version in a tooltip. A tooltip is not readable — it needs a hover the user has no reason to attempt. The ticket asks for "you have 11.3, 12.0 is out" to read off the tile.

## Change

The meta row now carries the delta: `Dist v9 → Dist v12`. The version you have recedes; the version you'd get takes the emphasis slot. No update, no arrow.

Version stays metadata, not a selector — the name is still the headline, and the tile keeps its two-line geometry.

Distribution installs resolve the target from the workspace catalog and stay silent while it's cold, so a loading catalog never renders a false claim. Other sources read the target from their existing update tag.

The delta sits outside the ellipsised facts, so the source label truncates before the version pair does.

## Already done in earlier PRs

The rest of DES-568 shipped in #1319 — the labelled `Dist v7` meta line, and the size removal (the platform dropped storage estimation, so Desktop stopped rendering it).

## Testing

`pnpm run typecheck` (4 configs), `lint`, `test` (3033), `test:integration` (27) — green. New tests verified to fail against the unmodified source.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -1161,6 +1161,7 @@
     "distribution": {
       "menuInstall": "Install",
       "distVersion": "Dist v{version}",
+      "distVersionNext": "v{version}",
       "notInstalled": "Not installed",
       "emptyTitle": "No distributions published to this workspace yet.",
       "loadError": "Couldn't load distributions. Retry",
--- a/locales/zh.json
+++ b/locales/zh.json
@@ -1161,6 +1161,7 @@
     "distribution": {
       "menuInstall": "安装",
       "distVersion": "分发版 v{version}",
+      "distVersionNext": "v{version}",
       "notInstalled": "未安装",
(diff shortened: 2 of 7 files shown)


shrimbly:
Closing — DES-568 came from the same batched ticket-writing pass as DES-565, and its premise isn't settled.

The ticket argues in point 1 that the version is metadata a user shouldn't need (Jacob: names over versions), then asks in point 2 for the delta to read off the tile. This PR built point 2, which pushes the version toward prominence. That tension wants a product decision, not an implementation.

Points 1 and 3 of the ticket (labelled `Dist v7` meta line, size removed) already shipped in #1319 regardless.

Branch `willie/des-568-desktop-show-the-builder-distribution-version-on-the-tile` left in place.

PR #1338, opened 2026-07-29
feat(chooser): add "Deploy as Distribution" to the install context menu

shrimbly:
Closes DES-566.

## Problem

Everything built so far is one-directional — distributions come *down* from Builder onto the machine. Design partners work the other way: author locally, then publish.

## Change

A **Deploy as Distribution** action on the install context menu, alongside Manage / Update / Delete. It confirms, then opens Comfy Builder in the browser.

The confirm is doing real work, not ceremony. The snapshot import carries **custom nodes and python pins only** — models, workflows, startup args and your files are set up separately in Builder, and the import is approximate. The dialog says so before the user leaves, split into "carried over" and "set up separately" so the gap is legible rather than buried in prose.

It also names the actual path: launch the instance once, export a snapshot via Share, then import. Offered on a fresh install, the old wording would have sent the user to Share's empty state — nothing captures a snapshot until first boot.

No in-Desktop publish flow and no new auth. The hand-off is one `openExternal` to a UTM-tagged Builder URL, repointable at build time via `VITE_COMFY_BUILDER_WEB_URL`.

## Notes

- UTM params moved to `src/shared/utmParams.ts` so main and renderer can't drift apart on attribution.
- Builder web origin is asserted in exactly one place, so the URL is a single edit when staging/prod split.

## Testing

`pnpm run typecheck` (4 configs), `lint`, `test`, `test:integration` — green.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -47,6 +47,7 @@
     "menuMigrate": "Migrate to Standalone",
     "menuRestoreSnapshot": "Restore Snapshot",
     "menuRevealInFolder": "Open Folder",
+    "menuDeployDistribution": "Deploy as Distribution",
     "menuDelete": "Uninstall",
     "stoppedActionGatedReason": "Stop the running instance first.",
     "menuDismissError": "Dismiss Error",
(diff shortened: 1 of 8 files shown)


shrimbly:
Needs better planning, this was claude on automation

PR #1339, opened 2026-07-29
feat(e2e): screen-capture harness for the comfy builder flow

shrimbly:
Part of DES-565.

## Change

An opt-in Playwright harness that drives the built app and captures the Comfy Builder flow screens to `captures/` at a consistent viewport and deterministic file names. Regenerate with `pnpm run test:e2e:capture` when the UI changes, instead of hand-grabbing once.

## Not in CI

The capture spec runs under its own `capture` Playwright project, gated on a `@capture` tag. The CI matrix runs `--project=macos|windows|linux` only, so a normal run never collects it. `captures/` is gitignored.

## Coverage and its gaps

The harness captures what is reachable **signed out**: chooser identity corner, host title bar, workspace shelf tile and kebab, manage popup, not-ready alert, install progress and failure.

Screens behind sign-in are **not** captured — that needs auth, which this PR deliberately does not implement. A final assertion fails if a declared screen is neither captured nor explicitly explained, so the set can't silently rot into claiming coverage it doesn't have.

## Figma upload

Out of scope here. This PR produces the PNG set; getting it into Figma is the follow-up.

## Shared harness change

`launchLauncherApp` now always resolves `app.getPath('userData')` rather than only when seeding installations, and exposes it on the context. One extra eval per launch; existing specs are unaffected.

## Testing

`pnpm run typecheck` (4 configs), `lint`, `test`, `test:integration` — green. Capture project verified absent from the CI matrix by inspection.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ out/
 result
 test-results/
 playwright-report/
+# Screenshot set written by `pnpm run test:e2e:capture` (regenerate, don't commit)
+captures/
 resources/vc_redist.x64.exe
 resources/vc_redist.x64.exe.tmp
 resources/vc_redist_version.nsh
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ On Windows/macOS, set up a fresh clone and start the app in one command with `pn
 | `pnpm test`                      | Unit tests ([Vitest](https://vitest.dev/))                            |
 | `pnpm run test:integration`      | Integration suite                                                     |
 | `pnpm run test:e2e`              | End-to-end tests ([Playwright](https://playwright.dev/))              |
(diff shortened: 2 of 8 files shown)


shrimbly:
Closing — DES-565 was written for a different purpose than this harness serves.

For the record, so the ticket isn't misread later: this covered ~11 screens against the ~21 the ticket enumerates, and missed the entire first-use takeover group. The ticket also names a mock-session hook for capture (`comfy.devplatform.mockSession` / `VITE_DEVPLATFORM_MOCK`) that does not exist anywhere in the repo.

Branch `willie/des-565-desktop-capture-all-comfy-builder-flow-screens-into-figma` is left in place if any of it is worth salvaging.

PR #1357, opened 2026-07-31
fix(auth): reuse active desktop login attempt

benceruleanlu:
## Summary

- keep the active Desktop login-code attempt alive when Sign in is clicked again
- reopen or focus the exact browser URL for the existing code instead of minting a replacement, and restore the copy-link card in case it was dismissed
- let repeat callers join the original attempt without taking ownership of legacy fallback
- require an explicit `startOver` option to cancel and replace an in-flight attempt
- add a one-shot **Start over** button to the sign-in card (en/zh) that re-enters the popup orchestration with `startOver: true` — the user-facing escape hatch from a wedged attempt
- clear `activeFlow` when an attempt's outcome rejects, and keep the entry point `async`, so a crashed attempt degrades to the legacy bridge instead of turning every later click into a silent no-op

## Why

A repeat click previously aborted Desktop's poller while leaving the original browser tab and login code alive. The replacement attempt then polled a different code, which made multi-click users disproportionately likely to hit terminal exchange failures such as 404s.

The active attempt now owns its code, URL, poller, fallback decision, and completion until it finishes or an explicit start-over replaces it.

Join-by-default makes the Sign in button idempotent — Desktop cannot close the tab it already opened, so converging on the existing code is the only state both sides can agree on. But it also means a wedged attempt has no recovery path until the code expires (30 minutes since #1356). The banner's Start over button provides that recovery, and the rejection guard closes the one state where recovery would otherwise require an app restart.

## Validation

- `mise exec -- pnpm exec vitest run src/main/auth` — 139 tests passed across 12 files
- repository commit hook: `pnpm run typecheck` — passed
- repository commit hook: `pnpm run lint` — passed
- targeted Prettier checks for the touched files — passed
- `git diff --check origin/main...HEAD` — passed

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -790,6 +790,7 @@
       "copy": "Copy link",
       "copied": "Copied",
       "openAgain": "Open again",
+      "startOver": "Start over",
       "dismiss": "Dismiss"
     },
     "capacityDegraded": "Heavy usage",
--- a/locales/zh.json
+++ b/locales/zh.json
@@ -790,6 +790,7 @@
       "copy": "复制链接",
       "copied": "已复制",
       "openAgain": "再次打开",
+      "startOver": "重新开始",
       "dismiss": "忽略"
     },
     "capacityDegraded": "高负载",
--- a/src/main/auth/desktopLoginCode/index.test.ts
+++ b/src/main/auth/desktopLoginCode/index.test.ts
@@ -587,32 +587,57 @@ describe('signInViaDesktopLoginCode', () => {
     })
   })
 
-  it('aborts the prior poll loop on re-entry without reporting a failure', async () => {
+  it('reopens the same URL and keeps polling the same code on repeat clicks', async () => {
     h.settingsGet.mockReturnValue(true)
     h.createDesktopLoginCode.mockResolvedValue(GRANT)
-    h.exchangeDesktopLoginCode.mockResolvedValue({ status: 'pending' })
+    h.exchangeDesktopLoginCode
+      .mockResolvedValueOnce({ status: 'pending' })
+      .mockResolvedValueOnce({ status: 'complete', custom_token: 'custom-token-value' })
+    mockSignInChain({ uid: 'uid-1' })
     const mod = await loadOrchestrator()
+    const contents = fakeContents()
 
-    const first = mod.signInViaDesktopLoginCode(AUTH_URL, fakeContents(), {})
+    const first = mod.signInViaDesktopLoginCode(AUTH_URL, contents, {})
     await vi.advanceTimersByTimeAsync(3500)
     expect(h.exchangeDesktopLoginCode).toHaveBeenCalledTimes(1)
+    const openedUrl = h.openExternal.mock.calls[0]![0]
 
-    // Second click: the new attempt cancels the stale poll loop. Its own
-    // create fails fast so the test doesn't leave a flow in flight.
-    h.createDesktopLoginCode.mockRejectedValueOnce(new Error('create failed'))
-    const second = mod.signInViaDesktopLoginCode(AUTH_URL, fakeContents(), {})
+    const second = mod.signInViaDesktopLoginCode(AUTH_URL, contents, {})
 
+    expect(h.createDesktopLoginCode).toHaveBeenCalledTimes(1)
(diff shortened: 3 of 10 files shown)


christian-byrne:
Reviewed this against the real modules (probes driving the actual state machine, not just a diff read). **The reuse state machine is correct** — ownership transfer, abort propagation, `sessionInjection` claim/release ordering, banner-cleanup ownership and the crashed-attempt-degrades-to-fallback guard all hold under double-click, click-during-create, click-during-poll, click-after-expiry, click-after-crash and explicit start-over. No listener leak, no spurious `sign_in_failed` on start-over, i18n complete in both locales, CI green, no conflict with `main`.

Two cases still reproduce the failure this PR is named after. Both are one-liners in `desktopLoginCode/index.ts`:

**1. Repeat click during the code-create window is a total no-op (up to 8s of dead button).** The reuse branch is gated on `existing.loginUrl` (:111), which is only populated at :245 after `createDesktopLoginCode` resolves — budget `CREATE_CODE_TIMEOUT_MS = 8000`. Clicks 2 and 3 inside that window produce zero `openExternal`, zero card, zero telemetry. That is exactly the impatient-double-click case, and it is worse than `main`, where the second click eventually opened a browser. Note the existing test `joins code creation without giving a repeat caller ownership of fallback` asserts `expect(h.openExternal).not.toHaveBeenCalled()`, so this is currently locked in as intended and won't be caught later.

Suggestion: either add `pendingReopen` to `ActiveDesktopLoginCodeFlow`, set it when the join hits a null `loginUrl`, and honor it at :246; or render the card immediately in a "preparing…" state and let the grant fill the URL in — `buildCopyLinkBannerScript` already supports in-place update via `existing.__cclUrl = URL` (`copyLinkBanner.ts:107`).

**2. Repeat click after redemption reopens an already-consumed code.** `activeFlow.loginUrl` is never cleared once the exchange returns `complete`, so the join branch keeps reopening it through the whole post-redeem tail (`signInWithCustomToken` → `lookupAccount` → 500ms hold → `injectFirebaseSession`). A click there opens a second tab on a spent code: the browser shows a redemption error while Desktop simultaneously completes sign-in. Clearing it at the `while` exit (:287) makes the join fall through to plain join-and-wait, which is the right behavior once the code is spent:

```ts
if (activeFlow?.controller === controller) activeFlow.loginUrl = null
```

Three non-blocking notes:

- `startedOver` (`flowState.ts:94`) is a closure local per `showCopyLinkBanner` call, so every restart installs a fresh card with `startedOver = false`. Repeated sentinels → unbounded restarts (measured: 5 emits → 6 bridges, 6 `openExternal`, each a backend write on the non-fallback path). Needs first-party XSS to abuse and `OPEN_LINK_SENTINEL` already offers an unguarded channel next to it, so this is amplification rather than new privilege — but the comment at `flowState.ts:92` claims a protection the code doesn't provide. A module-scope budget or a minimum interval between accepted sentinels would close it.
- The join checks `existing.comfyContents.isDestroyed()` but never compares it to the caller's `comfyContents` (:105-110), so a click from a different live view would join and get the card injected into the wrong view. Only theoretical today (`createHostWindow.ts:1191` is the sole caller) — worth a one-line guard rather than relying on that invariant.
- Each Start over emits an extra `sign_in_started` with no terminal event for the abandoned attempt, so N start-overs give N+1 starteds against one `user_logged_in`. `main` behaves the same on re-click and this PR's join actually *removes* that inflation for the common double-click, so it's not a regression — but a `restarted: true` property would let the funnel subtract them.

Worth coordinating with #1358 before either lands: the two PRs touch nearly the same file set (`flowState.ts`, `copyLinkBanner.ts`, `firebaseBridge/index.ts`, `desktopLoginCode/index.ts`, both test files, `en.json`) and both add a Start over affordance. Landing this one first and rebasing #1358 on top avoids the conflict and incidentally fixes #1358's locale-coverage CI failure, since this PR already updates `zh.json`.

One data point for context: on production telemetry the login-code flow's event-level failure rate is now ~6%/day, down from ~36% on the legacy loopback bridge, and ~92% of the residual failures are expired-code 404s rather than mechanical errors. The remaining wins in this area are about not stranding users mid-flow, which is what this PR is fixing.

PR #1358, opened 2026-07-31
Make Desktop browser handoff observable

benceruleanlu:
Desktop previously fire-and-forgot `shell.openExternal()`, swallowed rejection, and exposed only a dismissible recovery banner. This change awaits and records OS acceptance or rejection, keeps a persistent sign-in panel with countdown/status, Open browser again, Copy link, Cancel, and Start over, and preserves recovery after expiry or failure.

The OS result is deliberately treated only as handoff acceptance; Comfy-Org/ComfyUI_frontend#14457 adds the code-free browser page-arrival event.

Validation: full Desktop typecheck and ESLint pre-commit gates; 73 focused auth tests.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -786,11 +786,18 @@
     "label": "Cloud",
     "desc": "Run workflows on managed GPUs.",
     "signInBanner": {
-      "message": "We opened your browser to sign in. Didn’t open?",
+      "title": "Finish signing in in your browser",
+      "opening": "Opening your browser…",
+      "waiting": "Waiting for you to finish signing in.",
+      "openFailed": "We couldn’t open your browser. Try again or copy the link.",
+      "expired": "This sign-in link expired. Start over to try again.",
+      "failed": "Sign-in didn’t complete. Start over to try again.",
+      "remaining": "{time} remaining",
       "copy": "Copy link",
       "copied": "Copied",
-      "openAgain": "Open again",
-      "dismiss": "Dismiss"
+      "openAgain": "Open browser again",
+      "cancel": "Cancel",
+      "startOver": "Start over"
     },
     "capacityDegraded": "Heavy usage",
     "capacityDegradedHint": "ComfyUI Cloud is under heavy load right now — you may see slower response times. Please try again later for the best experience.",
--- a/src/main/auth/desktopLoginCode/index.test.ts
+++ b/src/main/auth/desktopLoginCode/index.test.ts
@@ -15,6 +15,7 @@ const h = vi.hoisted(() => ({
   bucketError: vi.fn(() => 'bucketed'),
   bindSignedInUser: vi.fn(),
   showCopyLinkBanner: vi.fn(),
+  updateSignInPanelStatus: vi.fn(),
   runBannerCleanup: vi.fn(),
   closeActiveBridge: vi.fn(),
   settingsGet: vi.fn(),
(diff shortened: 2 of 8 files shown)


christian-byrne:
Reviewed this together with Comfy-Org/ComfyUI_frontend#14457, since the two only pay off as a pair.

**The headline: after both land, you still can't compute a Desktop sign-in success rate.** The chain ends up as `sign_in_started {provider,flow}` → `browser_handoff {…,result,trigger}` → **[break]** → `app:desktop_login_code_captured {}` → `app:user_logged_in` (still deferred) | `sign_in_failed` (still direct).

**1. No join key anywhere.** The arrival event ships with no properties, and it fires in the system browser on the Cloud web app's anonymous person while Desktop's events fire in main on the Desktop anon id or Firebase UID — there's no `$anon_distinct_id` bridge across devices. Neither `sign_in_started` nor `browser_handoff` carries an attempt identifier. So "of the handoffs the OS accepted, how many reached the login page" has no query behind it; you get two unrelated counts on two populations. (Checked the project-token question: Desktop's hardcoded `phc_iKfK86id…` in `src/shared/posthogConfig.ts:6` *is* project 204330, the same project the Cloud app reports into — so at least the events do land in one place.)

Suggestion: mint an `attempt_id` per attempt, attach it to `sign_in_started` / `browser_handoff` / `sign_in_failed` / the success leg, and carry it (or a non-reversible hash of the code — never the code) as a second query param so the arrival can emit it. Alternative that sidesteps the cross-person problem entirely: emit arrival from the redeem endpoint server-side, since the backend already receives `installation_id` at code creation (`desktopLoginCode/index.ts:138`).

**2. `shell.openExternal` is now awaited behind an `executeJavaScript` on a page we don't control** (:187-202 here, `firebaseBridge/index.ts:151-158`). `showCopyLinkBanner` awaits `insertCSS` + `executeJavaScript` against the embedded Cloud view; `.catch()` covers rejection but not a *pending* promise, so a busy or mid-navigation renderer means the browser never opens — the exact "Sign in does nothing" case the comment at `firebaseBridge/index.ts:65-68` exists to prevent. `await openExternalSafely` has no timeout either, and in the login-code path the poll loop only starts after it, so a hung `xdg-open` means the code is never exchanged even if the user finishes via Copy link. Suggest firing `openExternalSafely` first, racing it against a timeout (emitting `result: 'timeout'`), and leaving the panel injection `void`-ed with status updated when the result lands.

**3. Cancel and Start over emit nothing.** This PR adds both controls; the cancel path returns early on `signal.aborted` so no terminal event fires at all, leaving deliberate cancellation indistinguishable from silent abandonment inside the `sign_in_started` denominator. Start over re-enters the flow and emits a *second* `sign_in_started` with `trigger: 'automatic'`, indistinguishable from a fresh attempt. Net: the PR meant to fix the denominator adds a new inflation source. A `sign_in_cancelled {provider, flow, attempt_id, at_stage}` plus `trigger: 'start_over'` with a `parent_attempt_id` would make both subtractable.

**4. `result: 'accepted'` will be ~99.x% a constant.** `openExternal` resolves whenever any handler exists; it doesn't report that a page loaded or a tab was focused. The PR body says as much — worth stating in the code comment too, so nobody builds a dashboard on it.

Consent gating checked as designed: `browser_handoff` via `capture()` and `sign_in_failed` via `emit()` ride the same gate, and keeping `browser_handoff` out of `DATADOG_MIRRORED_EVENT_NAMES` is the right call. But the asymmetry causing the success-side undercount is untouched — `app:user_logged_in` only escapes from inside `applyFirebaseUserBinding` (`telemetry.ts:773-860`), which returns or defers on four conditions beyond consent, while `sign_in_failed` fires unconditionally. Until a flow-owned success event exists (e.g. `sign_in_succeeded {provider, flow, attempt_id}` via plain `capture()` where the custom token is exchanged), no correlation id will produce a usable rate.

No PII/token findings — neither event carries a URL, code, email or user string, and all properties are bounded enums.

CI: `CI / Test` fails on `localeCoverage.test.ts` — `en.json` gained nine `cloud.signInBanner.*` keys and `zh.json` wasn't updated. Note #1357 already updates both locales and touches nearly this exact file set while adding its own Start over affordance; sequencing #1357 first and rebasing this on top resolves the conflict and this failure together.

For context on why this work matters: production telemetry now puts the login-code flow at ~6% event-level failures (down from ~36% on the loopback bridge), but only 64.7% of people who click Sign in complete it within 7 days, at a median of 2 attempts each, and ~92% of residual failures are expired-code 404s. In other words the remaining loss is almost entirely "user never finished in the browser" — which is precisely the step this pair is trying to make visible, and precisely why it needs a join key to be worth landing.

PR #1368, opened 2026-08-03
feat(telemetry): remove the comfy.desktop.comfyui.model_usage event

benceruleanlu:
## Why

`comfy.desktop.comfyui.model_usage` ships **~46M events/month from ~347k installs** into PostHog — and onward into Snowflake via the `Events to Snowflake (Prod)` batch export — with effectively no readers:

- **PostHog**: zero insights (searched the query JSON of all saved + deleted insights, not just names), zero dashboards, alerts, cohorts, actions, experiments, notebooks, CDP destinations, or saved SQL views reference the event.
- **Snowflake / data-platform (dbt)**: nothing selects the event by name or touches its `model_class` / `load_trigger` properties. It only passes through the generic `stg_posthog__events` → `int_events_unified` layers as anonymous `comfy_desktop` rows. (`fct_cloud_model_usage` is the cloud backend's unrelated `model_usage` event.) Desktop run metrics key on `comfy.desktop.execution.session_summary`.
- The only consumer ever was a one-off HogQL analysis in `comfy-desktop-metrics` (`queries/model_class_by_custom_node.sql`, findings 2026-07-08). That query stops working, by design; everything else is unaffected.

## What

- Remove the model-load half of the hardware tap: the three log-line parsers (`Requested to load` / dynamic-VRAM prepare / multi-GPU deepclone), the per-(class, trigger, device) delta counter, the cardinality cap, and the 5-minute flush interval.
- `accelerator_detected` is untouched, and the tap keeps its `ingest` / `beginBoot` / `flushSummary` interface — `flushSummary` still drains unterminated trailing lines and emits a pending accelerator event.
- The two stream-buffering tests that used model-load lines (stdout/stderr splice guard, trailing unterminated line) are rewritten against `Device:` lines so those guarantees stay covered.
- Update the three `launch.ts` comments that referenced the model-usage flush interval.

## Testing

- Full unit suite: 226 files / 3,678 tests pass.
- `typecheck`, `eslint`, `prettier --check` clean.

### DIFF
--- a/src/main/lib/hardwareTap.test.ts
+++ b/src/main/lib/hardwareTap.test.ts
@@ -11,15 +11,7 @@ vi.mock('electron', () => ({
   BrowserWindow: { getAllWindows: () => [] }
 }))
 
-const {
-  createHardwareTap,
-  parseDeviceLine,
-  parseVramLine,
-  parseRequestedModelLoad,
-  parseDynamicVramPrepare,
-  parseModelDeepclone,
-  parseModelLoad
-} = await import('./hardwareTap')
+const { createHardwareTap, parseDeviceLine, parseVramLine } = await import('./hardwareTap')
 const telemetry = await import('./telemetry')
 
 describe('parseDeviceLine', () => {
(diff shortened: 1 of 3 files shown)


deepme987:
I'm inclined to moving it to logs [if the assumptions on logs being cheaper are true]

benceruleanlu:
> I'm inclined to moving it to logs [if the assumptions on logs being cheaper are true]

it's already in logs, this is being parsed from our logs. Yes logs are cheaper. We'd also lose our 7 year event retention to keep it as 14 days, when the one time we used this event was for longitudinal model adoption. Also it makes it a headache to query when we move it to logs.

deepme987 (approved):

PR #1384, opened 2026-08-08
feat(titlebar): move Log in into the file menu

shrimbly:
Moves logging in out of the dashboard and into the title-bar file menu, so it sits with the other app-level actions instead of showing a CTA to every user on the chooser.

Closes DES-693.

> [!NOTE]
> Based on **#1382** (draft), not `comfy-builder` — that PR merges 41 commits of `main` in, and rewrites both files this touches. Until #1382 merges, GitHub will show its ~42k lines here too. **Review the 6 commits on this branch, or the file list below.**

## What changed

**The file menu grew a `Log in` item**, at the head of the settings group:

```
Open Dashboard
─────────────
New Instance / Add Existing Instance / Load Snapshot
─────────────
Log in            ← new, while signed out only
Desktop Settings / Send Beta Feedback / (Reset Zoom)
─────────────
Close Window / Quit Desktop
```

**The chooser's log-in button is gone.** `DevPlatformAccountChip` now renders nothing while signed out; the chip appears only once there is an account to name. The signed-in face, workspace switcher and sign-out are untouched.

**No feature flag.** An earlier revision gated this on a PostHog rollout flag. That was wrong and has been removed: the Comfy Builder rollout is decided *per account*, so the decision cannot be made until the user has logged in — gating the login button on it puts the flag's own precondition behind the flag. Gating what a signed-in account may *do* is a separate piece of work. No flag needs creating for this PR.

## Two things worth review attention

**1. The sign-out race guard is now shared.** The menu activates in main, with no renderer in the loop, so it cannot reuse the chip's `store.signIn()` path. The login call previously sat inside `ipcMain.handle(signIn)` with a **closure-local** `signOutGeneration`; a second call site would have silently forked it, and a sign-out could no longer kill a browser flow the menu had started. The counter is now module-scope behind one `signInToCloud()` primitive, and the IPC handler is a thin delegate over it.

**2. `authChanged` never reached the dashboard renderer.** Signing in from the menu completed but changed nothing on screen. A host window loads **no page of its own** — the dashboard lives in a child `panelView` `WebContentsView` — so broadcasting only to `BrowserWindow.getAllWindows()` delivered to an empty `webContents`.

This bug predates the branch but was unobservable: the only sign-in trigger used to be the chip, which set its store from `signIn()`'s *return value* and never needed the push. Moving the trigger into main made the broadcast the only path back. `broadcastAuthChanged` now also fans out to each live `panelView`; the original window loop is kept in case other surfaces rely on it.

## Verification

- `pnpm test` — **3865 passed**, 2 pre-existing skips
- `pnpm lint`, `pnpm typecheck` (node/web/e2e/integration), `pnpm format:check` — all clean, nothing new
- CodeRabbit — clean (it caught 2 exposure-hygiene issues on the earlier flagged revision; both fixed, then made moot by the flag's removal)
- **Manually tested in the running dev app**: menu item present while signed out, absent once signed in; dashboard shows no log-in button; sign-in from the menu now repaints the chip.
- **Not verified with e2e** — authored in a linked worktree, where Playwright's config would drive a different checkout.

New tests: 4 in `registerDevPlatformHandlers.test.ts` covering the panel-view broadcast (verified failing before the fix, by reverting the loop), the menu-driven `signInToCloud()` path, sign-out, and skipping torn-down/absent panel views. `titlePopup.test.ts` order assertions are updated rather than weakened — still exact `toEqual`, now with signed-out and signed-in variants. `DevPlatformAccountChip.test.ts` is new; the component had no test file.

`fileMenu.signIn` is added to both `en.json` and `zh.json` (`Log in` / `登录`); `localeCoverage.test.ts` is green.

## Files

| File | |
|---|---|
| `src/main/popups/titlePopup.ts` | the menu item + its activation branch |
| `src/main/lib/ipc/registerDevPlatformHandlers.ts` | shared login primitive; broadcast fix |
| `src/renderer/src/views/devplatform/DevPlatformAccountChip.vue` | signed-out branch removed |
| `locales/{en,zh}.json` | `fileMenu.signIn` |
| 3 test files | +308 lines, one of them new |


shrimbly:
Closing — we're pushing directly to `comfy-builder` instead of stacking on #1382 (now closed unmerged). Reopening the same work against `comfy-builder` shortly.

PR #1385, opened 2026-08-08
feat(titlebar): move Log in into the file menu

shrimbly:
Moves logging in out of the dashboard and into the title-bar file menu, so it sits with the other app-level actions instead of showing a CTA to every user on the chooser.

Closes DES-693.

## What changed

**The file menu grew a `Log in` item**, at the head of the settings group:

```
Open Dashboard
─────────────
New Instance / Add Existing Instance / Load Snapshot
─────────────
Log in            ← new, while signed out only
Desktop Settings / Send Beta Feedback / (Reset Zoom)
─────────────
Close Window / Quit Desktop
```

**The chooser's log-in button is gone.** `DevPlatformAccountChip` now renders nothing while signed out; the chip appears only once there is an account to name. The signed-in face, workspace switcher and sign-out are untouched.

**No feature flag.** An earlier revision gated this on a PostHog rollout flag. That was wrong and has been removed: the Comfy Builder rollout is decided *per account*, so the decision cannot be made until the user has logged in — gating the login button on it puts the flag's own precondition behind the flag. Gating what a signed-in account may *do* is a separate piece of work. No flag needs creating for this PR.

## Two things worth review attention

**1. The sign-out race guard is now shared.** The menu activates in main, with no renderer in the loop, so it cannot reuse the chip's `store.signIn()` path. The login call previously sat inside `ipcMain.handle(signIn)` with a **closure-local** `signOutGeneration`; a second call site would have silently forked it, and a sign-out could no longer kill a browser flow the menu had started. The counter is now module-scope behind one `signInToCloud()` primitive, and the IPC handler is a thin delegate over it.

**2. `authChanged` never reached the dashboard renderer.** Signing in from the menu completed but changed nothing on screen. A host window loads **no page of its own** — the dashboard lives in a child `panelView` `WebContentsView` — so broadcasting only to `BrowserWindow.getAllWindows()` delivered to an empty `webContents`.

This bug predates the branch but was unobservable: the only sign-in trigger used to be the chip, which set its store from `signIn()`'s *return value* and never needed the push. Moving the trigger into main made the broadcast the only path back. `broadcastAuthChanged` now also fans out to each live `panelView`; the original window loop is kept in case other surfaces rely on it.

## Verification

- `pnpm test` — **3037 passed**, 2 pre-existing skips
- `pnpm lint`, `pnpm typecheck` (node/web/e2e/integration) — clean
- `pnpm format:check` — **red on this branch, and on `comfy-builder` itself**: 481 files already fail `prettier --check` there, including all 5 of the files this PR edits, before any of my changes. I have deliberately NOT run `pnpm format` — it would bury this diff under a 481-file reformat. My added lines are prettier-conformant. Worth fixing separately.
- CodeRabbit — clean
- **Manually tested in the dev app** — menu item present while signed out, absent once signed in; dashboard shows no log-in button; sign-in from the menu repaints the chip. Note this was exercised while the branch still sat on the `des-691` base; the source changes here are byte-identical, but the surrounding tree is not, so it has not been re-run by hand since rebuilding on `comfy-builder`.
- **Not verified with e2e** — authored in a linked worktree, where Playwright's config would drive a different checkout.

Originally opened as #1384 against #1382; rebuilt directly on `comfy-builder` after that stack was abandoned. The change applied to this base with a clean 3-way merge — no conflicts.

New tests: 4 in `registerDevPlatformHandlers.test.ts` covering the panel-view broadcast (verified failing before the fix, by reverting the loop), the menu-driven `signInToCloud()` path, sign-out, and skipping torn-down/absent panel views. `titlePopup.test.ts` order assertions are updated rather than weakened — still exact `toEqual`, now with signed-out and signed-in variants. `DevPlatformAccountChip.test.ts` is new; the component had no test file.

`fileMenu.signIn` is added to both `en.json` and `zh.json` (`Log in` / `登录`); `localeCoverage.test.ts` is green.

## Files

| File | |
|---|---|
| `src/main/popups/titlePopup.ts` | the menu item + its activation branch |
| `src/main/lib/ipc/registerDevPlatformHandlers.ts` | shared login primitive; broadcast fix |
| `src/renderer/src/views/devplatform/DevPlatformAccountChip.vue` | signed-out branch removed |
| `locales/{en,zh}.json` | `fileMenu.signIn` |
| 3 test files | +308 lines, one of them new |

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -1291,6 +1291,7 @@
     "newInstall": "New Instance",
     "addExistingInstall": "Add Existing Instance",
     "loadSnapshot": "Load Snapshot",
+    "signIn": "Log in",
     "globalSettings": "Desktop Settings",
     "sendFeedback": "Send Feedback",
     "returnToDashboard": "Return to Dashboard",
--- a/locales/zh.json
+++ b/locales/zh.json
@@ -1291,6 +1291,7 @@
     "newInstall": "新建实例",
     "addExistingInstall": "添加现有实例",
     "loadSnapshot": "加载快照",
+    "signIn": "登录",
     "globalSettings": "桌面端设置",
     "sendFeedback": "发送反馈",
     "returnToDashboard": "返回仪表板",
--- a/src/main/cloud/session.test.ts
+++ b/src/main/cloud/session.test.ts
@@ -59,6 +59,80 @@ describe('CloudSession.getAccessToken', () => {
   })
 })
 
+describe('CloudSession browser auth', () => {
+  afterEach(() => vi.clearAllMocks())
+
+  it('single-flights repeated login requests', async () => {
+    let resolveLogin!: (value: Awaited<ReturnType<typeof signIn>>) => void
+    mocked(signIn).mockReturnValue(
+      new Promise((resolve) => {
+        resolveLogin = resolve
+      })
+    )
+    const session = new CloudSession()
+
+    const first = session.login()
+    const second = session.login()
+    resolveLogin({
+      tokens: { accessToken: 'current', expiresAt: future },
+      status: { signedIn: true, email: 'current@comfy.org' }
+    })
+
+    await expect(Promise.all([first, second])).resolves.toHaveLength(2)
+    expect(signIn).toHaveBeenCalledOnce()
+    expect(saveTokens).toHaveBeenCalledOnce()
+  })
+
+  it('does not persist a login that finishes after logout', async () => {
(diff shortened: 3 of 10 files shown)


Kosinkadink (approved):
Kosinkadink reviewed final head 022d3e32. The signed-out-only File menu placement, chooser CTA removal, panelView auth broadcast, i18n, and tests align with DES-693. Review found an auth-ordering race; 022d3e32 moves latest-intent handling into CloudSession and adds coverage. Local typecheck, lint, build, 3,042 unit tests, and 27 integration tests pass.

PR #1419, opened 2026-08-17
feat(templates): dynamic starter templates, live install progress, and a Cloud showcase

MaanilVerma:
## Summary

Starter templates can now be changed from PostHog without shipping a release, the install screen reports what it is actually doing instead of a stalling bar, and the wait itself now carries a rotating showcase of what Comfy Cloud offers.

## Changes

**What**

- Starter templates are served from a remote flag, resolved against the template index that the installed ComfyUI version pins, and fall back to the disk cache and then the built-in list when the network is unavailable. The picker also honours each template's availability window and release channel, so a template can be scheduled or limited to nightly without a code change.
- The installer emits structured per-item progress across IPC, so the takeover can show which file is downloading and how far along it is rather than a single opaque percentage.
- The dependency sync used to look frozen for minutes. `uv pip install` publishes no machine-readable progress, so we parse its human output into coarse stages (resolving, preparing, installing) and report those. There are no byte counts to be had here, so this is deliberately stage-level, not a percentage.
- The progress bar no longer starves its longest phase. Phase weighting was letting the slowest stretch of the install occupy a sliver of the bar while quick phases ate the rest.
- The install screen now runs a carousel above the wordmark, cycling nine reasons to run on Comfy Cloud (hardware, the 900+ pre-installed models, partner models, commercial licensing, custom nodes, LoRA import, queueing, and the runtime-only billing model). Copy comes from the marketing site, re-cut as plain text.
- Opening Comfy Cloud actually opens Cloud. It previously called a helper that only focuses an already-open window and otherwise falls through to the chooser, so the user landed on the dashboard.
- Every request through the shared fetch helper is now bounded by a timeout, and ops flags accept JSON payloads.

**Breaking**

None. IPC message shapes are additive, existing telemetry events are unchanged, and the built-in curated template list still ships as the final fallback so an offline first run behaves exactly as before.

## Review Focus

The interesting decision is in the template resolution chain: remote flag, then disk cache, then the compiled-in list, with the index resolved at the version ComfyUI pins rather than latest. `templateCatalogResolution.test.ts` is where that contract is pinned down and is the file worth reading first.

Worth a second look: `uvProgress.ts` is parsing human-readable output, which is inherently brittle. It is scoped to three coarse stages precisely so that a wording change upstream degrades to "no progress detail" rather than to wrong numbers.

The Cloud carousel is deliberately unconditional. There are no timing gates, no error branching, and no tiering, since it is a feature showcase for the whole wait rather than a targeted offer. That was a direct change in direction from an earlier build, and the earlier install-receipt card was removed rather than kept alongside it, so there is one Cloud surface rather than two.

Out of scope: the model download pipeline, launch sequencing, and the finished and error states of the install takeover are untouched.

## Testing

The risky parts here are resolution order and time-based behaviour, so both are covered by unit tests with fake timers rather than by driving the UI. Template resolution has the largest surface and gets the deepest coverage, since a wrong fallback silently ships the wrong templates to every new install.

- `templateCatalogResolution.test.ts` (806 lines): remote wins over cache, cache wins over built-in, malformed payloads fall through cleanly, and availability windows plus release channels filter as expected.
- `starterTemplateManifest.test.ts`: flag parsing, id validation, modality ordering, and cache round-tripping.
- `templatePin.test.ts`: index resolution against the pinned ComfyUI version.
- `uvProgress.test.ts`: stage detection across real uv output, including the sub-1 MiB case that announces nothing.
- `useShowcaseCarousel.test.ts`: rotation wraps past the last card and keeps looping (this one is a regression guard, an earlier build stopped dead on the final card), pause on hover, focus, and the explicit control, and manual moves restart the dwell.
- `useCloudGate.test.ts`: the gate fails closed on every signal, and opening Cloud runs the install's launch action instead of the window-focus helper. Both suites were mutation-tested by reverting the fix, which fails them.
- `InstallShowcase.test.ts`: the Cloud action stays present on all nine cards, and lives outside the rotating element so it cannot animate away.

Gate: `pnpm run typecheck`, `pnpm run lint`, and `npx vitest run` all clean, 3929 passed across 231 files.

Manual: ran a real install end to end and confirmed the dependency-sync stages appear instead of a frozen bar, the carousel rotates and loops, and the Cloud button opens a Comfy Cloud window rather than the dashboard.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -977,6 +977,10 @@
     "updateDepsChecking": "Checking requirements…",
     "updateDepsDryRun": "Checking for conflicts…",
     "updateDepsInstalling": "Installing requirements…",
+    "uvResolved": "Worked out {count} dependencies…",
+    "uvDownloading": "Fetching {name}…",
+    "uvDownloadingSized": "Fetching {name} — {size}",
+    "uvInstalling": "Installing dependencies…",
     "updateDepsUpToDate": "Requirements unchanged.",
     "updateFailed": "Update process failed with exit code {code}.",
     "updateNoGit": "No .git directory found in ComfyUI/. Commit-based updating requires a git repository.",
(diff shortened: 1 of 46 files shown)


MaanilVerma:
Closing: wrong scope. This bundled the dynamic-starter-templates and install-progress work, which belong in their own PRs. Reopening limited to the Cloud upsell.

PR #1422, opened 2026-08-17
fix(downloads): aggregate concurrent model download progress in the macOS Dock

### LINKED ISSUE #1421: [macOS] Dock progress jumps between concurrent model downloads instead of showing aggregate progress
## Description

When multiple models are downloaded concurrently, the macOS Dock progress bar alternates between the progress values of individual downloads instead of displaying the aggregate progress of the complete download batch.

Because macOS exposes a single Dock progress bar for the application, per-download progress updates overwrite each other. As a result, the displayed value can jump forwards and backwards as different downloads report progress.

Screenshot showing concurrent model downloads:

Concurrent model downloads in Comfy Desktop

Video showing the Dock progress jumping between downloads:

https://github.com/user-attachments/assets/791c58b5-a882-4807-acf8-c29df2cc2ac3

## Environment

- Comfy Desktop version: 1.0.39, locally packaged from source
- Source base: `Comfy-Org/Comfy-Desktop@1ff443aa6f5e71ac54d7b8b978d206f80c027ac4`
- Operating system: macOS 15.7.7 (24G720)
- Architecture: Apple Silicon (`arm64`)
- Installation method: Desktop app built from source
- Reproducibility: Consistent
- Custom nodes: Not involved

## Steps to Reproduce

1. Launch a packaged Comfy Desktop build on macOS.
2. Start downloading two or more models concurrently.
3. Use models with different file sizes or download speeds so their percentages differ.
4. Observe the progress indicator on the Comfy Desktop Dock icon.

## Expected Behavior

The Dock should display one byte-weighted aggregate progress value for the complete active download batch:

`sum(received bytes) / sum(total bytes)`

The aggregate should remain accurate as downloads start, complete, fail, or are cancelled, and the Dock progress bar should be cleared only after no downloads remain active.

## Actual Behavior

- The Dock progress bar alternates between the percentages of individual downloads.
- The displayed progress can jump forwards and backwards.
- Concurrent downloads are not represented as one aggregate workload.

## Logs

No application error or exception is emitted. This is a native Dock progress-state issue.

## Additional Context

This is separate from #1063, which concerns identifying concurrent downloads that share a URL but have different destination paths.

The issue occurs in the Electron main-process download progress handling and does not depend on a ComfyUI workflow, model execution, GPU, or custom nodes.

I am willing to work on a focused fix for this issue and submit a pull request. I have already investigated the relevant download progress path and would include regression tests for concurrent aggregation and download lifecycle edge cases.

xwsjjctz:
## Summary

- Aggregate concurrent model downloads into one byte-weighted Dock progress value on macOS.
- Preserve native per-window taskbar progress behavior on other platforms.
- Handle downloads with unknown sizes, completed jobs, cancellations, failures, and detached windows.
- Add regression coverage for concurrent progress aggregation and cancellation.

## Problem

Each model download currently reports its own progress directly to Electron's `setProgressBar()` API.

On macOS, the application has one shared Dock progress indicator. Concurrent downloads therefore overwrite each other's values, causing the Dock progress to jump forwards and backwards instead of representing the complete download workload.

Fixes #1421.

## Root Cause

Taskbar progress was tracked as a property of an individual download update. There was no persistent batch keyed by download ID and no aggregation across windows on macOS.

As a result, whichever download reported most recently controlled the Dock indicator.

## Changes

- Track taskbar download state by stable download ID.
- Aggregate macOS Dock progress across all live application windows.
- Calculate aggregate progress using total received bytes divided by total expected bytes.
- Keep completed downloads in the batch until the remaining downloads finish, preventing the aggregate from jumping backwards.
- Remove cancelled and failed downloads immediately and recalculate progress from the remaining workload.
- Keep the Dock progress empty while an active download has not reported its total size, avoiding a brief full-progress flash on macOS.
- Clear the native progress indicator only after the active batch finishes.
- Remove taskbar state when its owner window is detached or destroyed.
- Retain per-window progress behavior on non-macOS platforms.

### Automated

- `pnpm exec vitest run src/main/lib/comfyDownloadManagerModelJobs.test.ts`
  - 61 tests passed
- `pnpm test`
  - 255 test files passed
  - 4,343 tests passed
  - 2 existing tests skipped
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run format:check`
- `pnpm run build`

### Manual macOS Testing

Tested with a locally packaged Comfy Desktop 1.0.39 build on macOS 15.7.7, Apple Silicon (`arm64`).

1. Started multiple large model downloads concurrently with different file sizes and download speeds.
2. Compared the individual percentages in the download list with the progress displayed on the Dock icon.
3. Verified that concurrent downloads are represented by one aggregate Dock progress value instead of alternating between individual percentages.
4. During the first validation pass, identified two aggregation edge cases:
   - Starting a download could briefly display a full Dock progress bar before its total size was known.
   - Cancelling one concurrent download did not immediately update the aggregate.
5. After addressing both edge cases, repeated the same concurrent-download scenario.
6. Verified that the Dock no longer flashes as complete when a download starts.
7. Cancelled one download while another remained active and verified that the Dock progress was recalculated from the remaining workload.
8. Continued downloading multiple models and observed normal aggregate progress behavior.

## Compatibility

This change does not modify public APIs, download identifiers, file layout, or model download behavior. Non-macOS platforms retain per-window native taskbar progress.

### DIFF
--- a/src/main/lib/comfyDownloadManager.ts
+++ b/src/main/lib/comfyDownloadManager.ts
@@ -716,9 +716,10 @@ interface TaskbarDownload extends DownloadProgress {
   id: string
 }
 
-/** A taskbar batch retains terminal jobs until the last job finishes. Without
+/** A taskbar batch retains completed jobs until the last job finishes. Without
  *  that retention, completing one of two downloads would remove its bytes
- *  from the denominator and make the aggregate bar jump backwards. */
+ *  from the denominator and make the aggregate bar jump backwards. Cancelled
+ *  and failed jobs are removed because they no longer belong to the workload. */
 const taskbarBatches = new Map<BrowserWindow, Map<string, TaskbarDownload>>()
 
 function mergeTaskbarDownload(
@@ -739,17 +740,17 @@ function mergeTaskbarDownload(
   return merged
 }
 
-/** Aggregate known byte totals. An active job without a known size makes the
- *  native bar indeterminate (> 1 per Electron) rather than showing a false
- *  percentage. Terminal unknown-size jobs do not hide progress for work that
- *  is still measurable. */
+/** Aggregate known byte totals. Keep the bar empty until every active job has
+ *  reported its size: Electron's indeterminate value (> 1) briefly renders as
+ *  a full Dock bar on macOS. Completed unknown-size jobs do not hide progress
+ *  for work that is still measurable. */
 function aggregateTaskbarProgress(downloads: Iterable<TaskbarDownload>): number {
   const jobs = [...downloads]
   if (!jobs.some((job) => !isTerminalStatus(job.status))) return -1
   if (
     jobs.some((job) => !isTerminalStatus(job.status) && !(job.totalBytes && job.totalBytes > 0))
   ) {
-    return 2
+    return 0
   }
 
   let receivedBytes = 0
@@ -763,7 +764,7 @@ function aggregateTaskbarProgress(downloads: Iterable<TaskbarDownload>): number 
         : (job.receivedBytes ?? job.progress * job.totalBytes)
     receivedBytes += Math.max(0, Math.min(job.totalBytes, received))
   }
-  return totalBytes > 0 ? Math.max(0, Math.min(1, receivedBytes / totalBytes)) : 2
+  return totalBytes > 0 ? Math.max(0, Math.min(1, receivedBytes / totalBytes)) : 0
 }
 
 function liveTaskbarBatches(): Array<[BrowserWindow, Map<string, TaskbarDownload>]> {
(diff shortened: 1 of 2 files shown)


xwsjjctz:
I have read and agree to the Contributor License Agreement

xwsjjctz:
Hi @Kosinkadink, just a friendly follow-up on this PR. Thanks for your response in #1421. When you have a chance, could you please take a look? The fix is ready for review and includes focused regression tests and manual verification on a packaged macOS build. I would be happy to address any feedback. Thank you!

xwsjjctz:
The progress bar in the dock was disabled in a later version, so this draft is now closed.

PR #1426, opened 2026-08-19
feat(install): live output showcase during the install wait

MaanilVerma:
## Summary
The install/first-launch wait now plays a looping showcase of real Comfy outputs behind the progress bar, with a rotating one-line pitch and a Try Cloud shortcut across the top. It rides the whole wait, both the environment setup and the instance launch that follows, instead of leaving the user staring at a bare progress bar.

## Changes
**What**
- While an install runs, a set of real output clips animate in travelling rounded frames behind the centered progress column, so the wait shows what Comfy can make rather than dead space.
- A short pitch line rotates in the top-right (7s a card, pausing when the pointer or focus is on it), with a fixed Try Cloud button beside it that launches an existing Cloud instance. The C mark sits opposite it on the left, so the two read as one top bar.
- The showcase now stays up across both legs of the install. It used to vanish at the ~70% mark when setup handed off to the instance launch, since that second leg is a separate operation; it now covers that leg too.
- The output clips loop, so a frame whose clip is shorter than its time on screen no longer freezes on its last frame.
- The animation renders natively in the panel (one shared document and compositor with the carousel and progress stepper) and pauses itself when the window is hidden, so it does not compete with the install for the main thread.

**Breaking**: None. The progress bar, stepper, footer controls (Return to Dashboard, port-conflict actions, Reboot, View logs) and their state transitions are unchanged; telemetry and IPC are untouched.

## Review Focus
- The crux is where the showcase is gated: it shows for an install operation or the launch leg chained off it, for the full duration of each. Worth confirming it neither disappears at the setup to launch handoff nor lingers onto unrelated operations.
- The scene is a native rebuild of an After Effects comp (keyframed bezier paths driving masked `<video>` layers on one rAF loop). The clips buffer up front on purpose, since reducing that leaves masks black in the real renderer under install load, so that is deliberate, not an oversight.
- Try Cloud shows on every card and is not gated behind a tier or feature flag, by product decision. The launch itself is still gated in the main process (no cloud install, no launch).
- Scope note: this is only the install-wait screen. It deliberately does not touch the chooser, the finished/error surfaces, or any other takeover.
- The nine output clips (~11MB webm) are first-party Comfy brand assets committed to the repo; Git LFS is a reasonable follow-up but out of scope here.

## Testing
The animation runtime is split so the deterministic math is unit-tested while the rAF/video/lifecycle surface, which jsdom cannot run without fragile fakes, is left to manual verification (the repo's zero-tolerance flaky-test policy). Everything timer-driven uses fake timers with teardown before the real clock returns.

- `useBrandScene.test.ts`: the bezier easing and keyframe interpolation (endpoints, clamping, linear and eased segments, zero-span guard).
- `useShowcaseCarousel.test.ts`: rotation, dwell floor, wrap-forever, pause/resume with a full dwell on resume, and that every card's copy resolves in en and zh within the single-line length budget.
- `useCloudGate.test.ts`: opening Cloud (launch vs focus-an-open-window vs no install), fail-closed on a thrown lookup, and double-click de-duping to a single launch.
- `InstallShowcase.test.ts`: one card at a time, the CTA emits and stays put while text rotates, hover/focus pause and resume.
- `ProgressModal.test.ts`: the showcase shows during install, rides the chained launch leg, and is absent for unrelated operations.

Gate: `typecheck`, `lint`, `format:check` clean; full renderer suite 1365 passing. Manual: run `pnpm dev` to an install and confirm the clips fill their frames, the top bar aligns, and the showcase stays up past the 70% handoff.

https://github.com/user-attachments/assets/c221e639-aafd-4110-b494-828882b73b3a


MaanilVerma:
Thanks @coderabbitai. Addressed the two comments:

**1. `useBrandScene` — visibility/reduced-motion on mount** ✅ Fixed. `onMounted` now always registers the `visibilitychange` listener and only calls `play()` when `!document.hidden`, so mounting on a hidden tab no longer starts the rAF loop or video decode; `onVisibility` begins playback when the tab shows. The `prefers-reduced-motion` early return is preserved. Added deterministic mount tests (stubbed `requestAnimationFrame`/`matchMedia`/`document.hidden`) covering visible-start, hidden-mount-no-start, reduced-motion-no-start, and unmount teardown.

**2. `InstallShowcase` carousel — pause rotation under reduced-motion** — Skipping in this PR, with a reason. The rotating carousel (`useShowcaseCarousel`) shipped in #1420 and this PR does not change its rotation behavior; it only re-homes the component into the top bar. Pausing rotation under reduced-motion is a valid a11y improvement, but it belongs to the shipped carousel, not this feature, and would pull #1420's composable into scope. Better as a focused follow-up so it lands with its own test rather than riding a layout/animation PR.

Rebased onto `main` (the earlier conflicts came from #1420 landing the showcase/cloud-gate that this branch had also carried); the branch now adds only the scene animation and the top-bar wiring on top of main.

src/renderer/src/components/BrandSceneAnimation.vue
```diff
@@ -0,0 +1,84 @@
...
+<template>
+  <div class="brand-scene-wrap" aria-hidden="true">
+    <div ref="stageRef" class="brand-scene-stage">
+      <div
+        v-for="(scene, si) in data.scenes"
+        :key="scene.id"
+        class="brand-scene"
+        :style="{ zIndex: data.scenes.length - si, borderRadius: `${scene.maskRadius}px` }"
+      >
+        <!-- loop + preload="auto" are load-bearing: some layers outlast their clip's
```
MaanilVerma: strip comments from here.

src/renderer/src/composables/useBrandScene.test.ts
```diff
@@ -0,0 +1,139 @@
...
+  it('pins the endpoints and clamps outside [0,1]', () => {
+    const ease = makeBezier(0.42, 0, 0.58, 1)
+    expect(ease(0)).toBe(0)
+    expect(ease(1)).toBe(1)
+    expect(ease(-0.5)).toBe(0)
+    expect(ease(2)).toBe(1)
+  })
+
+  it('is the identity for a linear control curve', () => {
+    // Precision matches the solver's own 1e-5 x-tolerance, not exact equality.
```
MaanilVerma: either remove the comment or use it as message in expect

src/renderer/src/lib/brandScene/installShowcaseScene.json
```diff
@@ -0,0 +1,337 @@
+{
```
MaanilVerma, line 1: can we improve the naming here + file name as well. scene3 seems to be a ported name. Add some useful name here

src/renderer/src/views/ProgressModal.test.ts
```diff
@@ -247,6 +247,17 @@ describe('ProgressModal — brand branch state transitions', () => {
     expect(body.exists('[data-testid="install-showcase"]')).toBe(false)
   })
 
+  it('keeps the Cloud showcase through the chained launch leg (past the 70% handoff)', async () => {
+    // The install→launch chain hands off at 70%: the launch leg is a separate op
```
MaanilVerma: again, comments either in expect message or remove

MaanilVerma:
Addressed the review comments (pushed):

- **BrandSceneAnimation.vue** — stripped the template comment.
- **useBrandScene.test.ts** / **ProgressModal.test.ts** — moved the rationale into the `expect(...)` message instead of a preceding comment.
- **scene3 naming** — renamed the ported `scene3` throughout: `scene3.json` → `installShowcaseScene.json`, `public/scene3/` → `public/install-showcase-scene/`, and the video paths + import updated to match.

Full renderer suite green (1389). Rebased on `main` so it's mergeable.

MaanilVerma:
@coderabbitai addressed the latest round:

- **`scene3` rename incomplete (Critical)** — resolved. The import and the in-JSON asset paths now point at `installShowcaseScene.json` / `/install-showcase-scene/Assets/`; that finding was against a mid-rename commit. E2E should pass now.
- **hidden-at-mount playback** — already fixed (`if (!document.hidden) play()`), with mount tests.
- **`useBrandScene.test.ts` unmount in afterEach** — fixed. Every mounted wrapper is tracked and unmounted before mocks are restored, so no listener leaks across cases.
- **ProgressModal standalone-launch negative case** — added. A plain launch (`opKind: 'launch'`, `chainSpan: null`) now asserts the showcase stays hidden, so a regress-to-always-true would fail.

- **Carousel rotation under prefers-reduced-motion (Major)** — deferring. `useShowcaseCarousel` shipped in #1420 and this PR does not change its rotation; the a11y gap is pre-existing there. Better as a focused follow-up with its own test than widening this layout/animation PR into shipped code.

Full renderer suite green (1390).

deepme987 (approved):

PR #1427, opened 2026-08-19
feat(templates): serve the starter-template list from R2

MaanilVerma:
## Summary
The install picker's starter-template list is now served from R2 instead of being hardcoded, so the curated set can change without shipping a desktop release. The bundled `CURATED_TEMPLATES` stays as the floor: if the remote document is missing, malformed, or partially invalid, the picker still renders its four cards per modality from the built-in set.

## Changes
- The starter list is fetched from `starter-templates.json` on R2 once per process, in parallel with the existing template index, and neither read can reject the picker path (both fall back on failure).
- The remote document is treated as fully untrusted input and validated field by field: bad entries drop individually, an unusable document is ignored whole, and per-modality slots are filled from the curated floor so the grid is never short.
- Deleting the remote document is the rollback, so there is no disk cache to outlive it; the list is memoized only for the life of the process.
- Guardrails on the parsed values: id pattern and reserved-name checks, text length caps, a non-negative integer size with an upper bound past the largest real template (beyond it the disk-space gate is unsatisfiable), an entry cap, and a fetch timeout since the picker blocks on this read.
- `apiNode` and `recommended` must be real booleans; a truthy non-boolean (e.g. a string) is rejected rather than coerced.

**Breaking**: None. With no remote document, or any fetch/parse failure, behavior is identical to today's hardcoded list. The template index read, its URL, and the card contract are unchanged.

## Review Focus
- `remoteStarterTemplates.ts` is the whole trust boundary. The riskiest hunks are `parseSnapshot` and the per-entry validation loop: a value that slips through here reaches the picker and the downstream disk-space gate, so the size bound and the integer/boolean rejections are load-bearing, not cosmetic.
- The `templateCatalog.ts` seam: `loadStarterTemplates()` must never reject (the `Promise.all` has no catch around it by design, the function owns its own fallback). Worth confirming the fallback-to-curated path can't throw.
- Scope boundary: this only swaps the source of the list. It does not touch the index lookup, the card component, or how templates install.

## Testing
Unit tests drive the parser and the catalog seam directly, since this is pure main-process logic with no UI.

- `remoteStarterTemplates.test.ts`: the validation matrix, one case per rejection (bad id, oversized text, negative/huge/non-integer size, non-boolean flags, reserved ids, over-cap entries, fetch failure, timeout) plus the happy path and the partial-drop-then-fill-from-curated behavior.
- `templateCatalog.test.ts` / `index.test.ts`: the seam, that the remote list feeds the grid when valid and falls back to curated on failure without changing the index read.

Gate: `typecheck:node` clean; the three affected suites pass (149 tests). No mutation-QA matrix on this branch (it did not go through the qs pipeline), so this section is from the real test files, not recorded survivor data.

Rebased on the latest `main` (`1.0.40-rc.1`); no conflicts with our files.

### DIFF
--- a/src/main/sources/standalone/index.test.ts
+++ b/src/main/sources/standalone/index.test.ts
@@ -261,7 +261,11 @@ describe('standalone.getFieldOptions bundledTemplate', () => {
   })
 
   it('falls back to snapshot metadata when the index fetch fails (offline)', async () => {
-    mockedFetchJSON.mockImplementationOnce(() => Promise.reject(new Error('offline')))
+    mockedFetchJSON.mockImplementation((url: string) =>
+      String(url).includes('index.json')
+        ? Promise.reject(new Error('offline'))
+        : Promise.resolve(null)
+    )
     const options = await standalone.getFieldOptions!('bundledTemplate', {}, {})
     const first = CURATED_TEMPLATES[0]!
     const card = options.find((o) => o.value === first.id)!
--- a/src/main/sources/standalone/remoteStarterTemplates.test.ts
+++ b/src/main/sources/standalone/remoteStarterTemplates.test.ts
@@ -390,7 +390,7 @@ describe('G. an oversized or hostile document cannot cost more than it should', 
   it('stops reading long past the point where every slot could be filled', () => {
     const huge = Array.from({ length: 5000 }, (_, i) => entry({ id: `r${i}`, modality: 'image' }))
     const parsed = parseRemoteStarterTemplates(doc(huge))
-    expect(parsed!.length, 'only 16 slots can ever be filled').toBeLessThanOrEqual(256)
+    expect(parsed!.length, 'parsing stops at the MAX_ENTRIES cap').toBe(256)
   })
 
   it.each([
@@ -437,8 +437,8 @@ describe('G. an oversized or hostile document cannot cost more than it should', 
     const parsed = parseRemoteStarterTemplates(doc([entry({ id: 'none' }), entry({ id: 'ok' })]))
     expect(
       parsed?.map((t) => t.id),
-      'the sentinel is not a template'
-    ).not.toContain('none')
+      'the sentinel drops while its valid neighbour survives'
+    ).toEqual(['ok'])
   })
 })
 
--- a/src/main/sources/standalone/remoteStarterTemplates.test.ts
+++ b/src/main/sources/standalone/remoteStarterTemplates.test.ts
@@ -3,6 +3,7 @@ import { describe, it, expect, vi, beforeEach } from 'vitest'
 vi.mock('../../lib/fetch', () => ({ fetchJSON: vi.fn() }))
 
 import {
+  STARTER_TEMPLATES_URL,
   parseRemoteStarterTemplates,
   resolveStarterTemplates,
   loadStarterTemplates,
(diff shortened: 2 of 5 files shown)


MaanilVerma:
## `starter-templates.json` — the document this PR reads

Recording the shape here so it is reviewable alongside the code and findable later. Everything below is taken from the parser in `remoteStarterTemplates.ts`, not written by hand.

**Location:** `https://desktop-assets.comfy.org/standalone-environments/starter-templates.json`

Not published yet. Until it exists the picker renders the built-in list, so this merges as a no-op.

### Shape

```json
{
  "schemaVersion": 1,
  "templates": [
    {
      "id": "image_z_image_turbo",
      "modality": "image",
      "recommended": true,
      "snapshot": {
        "title": "Z-Image-Turbo Text to Image",
        "description": "An Efficient Image Generation Foundation Model with Single-Stream Diffusion Transformer, supports English & Chinese.",
        "sizeBytes": 20830591386,
        "mediaSubtype": "webp"
      }
    },
    {
      "id": "api_bytedance_seedream_5_0_pro_image_edit",
      "modality": "image",
      "apiNode": true,
      "snapshot": {
        "title": "Seedream 5.0 Pro: Image Edit",
        "description": "Edit images with precision using Seedream 5.0 Pro…",
        "sizeBytes": 0,
        "mediaSubtype": "webp"
      }
    }
  ]
}
```

### Fields

| Field | Required | Rules |
|---|---|---|
| `schemaVersion` | yes | Must be exactly `1` (number). Anything else and the whole document is ignored, so an older client cannot half-apply a future schema. |
| `templates` | yes | Array. Only the first 256 entries are read; only 16 slots can ever be filled. |
| `id` | yes | `^[a-zA-Z0-9_.-]+$`, ≤128 chars. `.`, `..` and the `none` sentinel are rejected. The id reaches a deeplink URL and a package-relative path. |
| `modality` | yes | One of `video`, `image`, `3d`, `audio`. |
| `snapshot.title` | yes | Non-empty after trimming, ≤4096 chars. |
| `snapshot.description` | yes | Non-empty after trimming, ≤4096 chars. |
| `snapshot.sizeBytes` | yes | Integer, `0` … `2199023255552` (2 TiB). Drives the install-time disk-space gate. Use `0` for API-node cards. |
| `snapshot.mediaSubtype` | yes | Non-empty. `webp` for image previews, `mp3` for audio (renders the glyph instead). |
| `recommended` | no | `true` only. The auto-selected pick. At most one per tab, never on a paid card. |
| `apiNode` | no | `true` only. Runs on API nodes: downloads nothing, spends credits per run. |

`recommended` and `apiNode` must be real booleans if present. A quoted `"true"` or a `1` drops the entry rather than degrading it, since a truthy-but-not-`true` `apiNode` would otherwise offer a paid template as free.

### What happens when something is wrong

The built-in list is the floor, so the picker always shows 4 cards per tab across all 4 tabs.

- **One bad entry** → that entry drops, its slot backfills from the built-in list. Three good and one broken keeps the three.
- **Whole document unusable** (bad JSON, unknown `schemaVersion`, `templates` missing or not an array, nothing valid, 404, offline, >5s) → ignored entirely, all 16 built-in cards.
- **Document fine** → only its cards render, in document order, honouring its badges.

Two rules are enforced after parsing rather than trusted from the file: at most one `recommended` per tab, and never on an `apiNode` card. A tab with no free card gets no recommendation at all and the wizard offers skip. When a tab is partly backfilled the badge stays on a remote card, so a built-in top-up cannot claim the auto-pick content did not choose.

A remote entry reusing a built-in id must keep that id's modality, otherwise filing the four image ids under `video` would consume them before the image tab is reached and leave it empty.

### Notes for whoever publishes it

- Upload with `Content-Type: application/json`. Served as `application/octet-stream` it may not parse, and desktop silently falls back to the built-in list, which reads as "the feature doesn't work" rather than an error.
- Fetched once per process, so a change lands on the next launch. There is deliberately no disk cache: deleting the document is the rollback and it must take effect rather than be outlived by a cache.
- Tab order (`video, image, 3d, audio`) is UI structure and stays in the app.

deepme987 (approved):

PR #1440, opened 2026-08-21
feat(comfybuilder): the client calls the build API and the copy says Build (BE-8391)

james00012:
**TL;DR:** Desktop's comfy-builder client now speaks the renamed build API (`/v1/builds`, `/v1/build-versions`, `builds` envelope) and every user-facing "distribution" now says Build.

Linear: [BE-8391](https://linear.app/comfyorg/issue/BE-8391) · part of [BE-8386 Rename distribution to build across the public API and its callers](https://linear.app/comfyorg/issue/BE-8386)

**Why:** The comfy-builder API renamed its public vocabulary from distribution to build; the old paths keep serving as legacy aliases, so this change moves Desktop onto the new vocabulary before the aliases eventually go away. Copy follows so the product says the same word the API does.

**What changed:**
- Client paths and fields: `src/main/comfybuilder/client.ts` calls `/v1/builds`, `/v1/builds/{id}/versions`, `/v1/build-versions/{id}` and `/manifest` (`/v1/build-artifacts/{id}/download` unchanged) and reads the `builds` list envelope; the URL-asserting unit tests moved with it. `types.ts` needed no change: the renamed version-detail field `distributionId` was never mirrored or read by Desktop.
- en and zh copy values, keys untouched: 11 en values and 11 zh values now say Build / 构建版, including `distVersion` "Dist v{version}" -> "Build v{version}" (kept labelled so it cannot be read as the ComfyUI version beside it) and `blockedReason.buildFailed` reworded to "This build has no successful version yet."
- Hardcoded strings: the source description and the broken-record message in `src/main/sources/comfybuilder/index.ts`, and the "already installs this build" handler message in `src/main/lib/ipc/registerDevPlatformHandlers.ts`.

**Contradicts:** nothing.

**Release order:** merge is safe any time (main is not a release), but the next Desktop RELEASE must not ship before the builder rename is live on production, because Desktop talks to production only. **Chinese term 构建版 is provisional pending confirmation on the parent ticket; one-line swap if it changes.**

**Validation:**
- Unit: full vitest suite green (261 files, 4519 tests), with differential proof: the two URL-asserting client tests fail against the origin/main client (`/v1/builds` vs `/v1/distributions`, `/v1/build-versions` vs `/v1/distribution-versions`).
- Typecheck (node, web, e2e, integration) and eslint: green.
- Integration suite: green (40 tests). E2E: the comfybuilder specs (`comfybuilder-launch`, `comfybuilder-models`, macos project) green locally; the windows/linux/lifecycle projects cannot run on this macOS machine.
- Not run: a dev build against staging; staging serves /v1/builds only once the builder rename deploys, so the observed chooser proof is queued behind that.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -21,7 +21,7 @@
     "legacyDesktop": "Legacy Desktop",
     "remote": "Remote",
     "unknown": "Unknown",
-    "distribution": "Distribution"
+    "distribution": "Build"
   },
   "chooser": {
     "title": "Choose an instance",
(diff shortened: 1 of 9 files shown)


Kosinkadink:
My agent apparently forgot to make a PR for my comfy builder changes yesterday from my machine, I'll get my PR up and get the two PRs in one after the other + compatible. https://github.com/Comfy-Org/Comfy-Desktop/pull/1441

wei-hai (approved):

PR #1460, opened 2026-08-26
fix(mcp): await the flag fetch before the sidebar gate decides

MaanilVerma:
## Summary

The Local MCP sidebar icon never appears in packaged/ToDesktop builds, even though the `mcp_sidebar_enabled` flag is on in PostHog. It works in `pnpm dev` but not for real users. This makes the whole feature invisible in production.

## Root cause

The injection gate read the flag synchronously (`getFlag`) at `dom-ready`, before the async PostHog fetch had populated the cache. On a cold start (fresh or wiped install, i.e. every real first launch) `getFlag` returns `undefined`, so the icon is never injected.

Worse, the boot fetch is capped at a 1500ms timeout, and on timeout it returns `{}` — which skips `writeCache`, so the flag is never seeded on disk either. That means it fails on every subsequent launch too, not just the first. Dev only worked because the fetch happened to win the race on a warm machine.

## Changes

- `attach.ts`: the gate now `await`s `getFlagAsync` (the accessor every other flag consumer already uses) so injection waits for the fetch to resolve instead of reading an empty cache. Guards `comfyContents.isDestroyed()` since we now await across a tick.
- `experiments.ts`: `initExperiments` is fire-and-forget and its only awaiter is `getFlagAsync`, so the 1500ms cap wasn't protecting boot — it just discarded slow-but-successful fetches and left the cache unseeded. Raised to a 10s hang-guard so a slow network still resolves and seeds the cache.

## Review focus

- `experiments.ts` `DEFAULT_TIMEOUT_MS`: `loadFeatureFlagsImmediate` has exactly one caller (`initExperiments`), which is `void`-called at boot, so widening this timeout doesn't block startup. Worth a sanity check that no other path depends on the 1500ms bound.

## Testing

- Reproduced the failure locally by forcing the timeout to 1ms and clearing the flag cache: the icon disappeared, and logs confirmed the chain (empty cache -> fetch returns {} -> getFlag undefined -> willInject=false).
- With the fix, verified the icon appears on a cold-boot **packaged** build (empty cache) built via `electron-builder --dir` — the same packaging path as the failing ToDesktop build.
- Also built and smoke-tested an integration build (this fix merged with upstream/main) to confirm no regression in other features.

### DIFF
--- a/src/main/lib/experiments.ts
+++ b/src/main/lib/experiments.ts
@@ -32,7 +32,13 @@ import { configDir } from './paths'
 import * as mainTelemetry from './telemetry'
 import type { FeatureFlagValue } from './telemetry'
 
-const DEFAULT_TIMEOUT_MS = 1500
+// The boot fetch is fire-and-forget (`void initExperiments`) and its only
+// awaiter is `getFlagAsync`, which wants the real value. A short timeout here
+// doesn't protect boot — it just discards a slow-but-successful fetch, leaving
+// the cache unseeded and every `getFlagAsync` reader stuck on `undefined`. So
+// the cap is a hang-guard, not a UI budget: long enough that a slow network
+// still resolves and seeds the cache.
+const DEFAULT_TIMEOUT_MS = 10_000
 
 export type ExperimentExposureSource = 'cache' | 'remote' | 'fallback'
 
--- a/src/main/host/attach.test.ts
+++ b/src/main/host/attach.test.ts
@@ -24,12 +24,6 @@ describe('shouldInjectMcpSidebar', () => {
       true
     )
   })
-
-  // The finding: `getFlagAsync` can resolve up to ~10s later. If a detach /
-  // hot-swap ran `_installCleanup` in the meantime it clears `attachActive`,
-  // and because detach leaves `comfyContents` alive the destroyed check passes.
-  // Without the attachActive gate this would inject into a detached / re-attached
-  // view.
   it('does not inject when the attach was retired before the flag resolved', () => {
     expect(shouldInjectMcpSidebar({ attachActive: false, enabled: true, destroyed: false })).toBe(
       false
--- a/src/main/host/attach.test.ts
+++ b/src/main/host/attach.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('electron', () => ({
+  app: {
+    isPackaged: false,
+    getPath: () => '/tmp',
+    getVersion: () => '0.0.0-test',
+    getLocale: () => 'en',
+    on: () => {}
+  },
+  ipcMain: { handle: vi.fn(), on: vi.fn(), off: vi.fn() },
+  dialog: { showMessageBox: vi.fn() },
+  shell: {},
+  WebContentsView: class {},
+  BrowserWindow: { getAllWindows: () => [] },
+  nativeTheme: { on: vi.fn(), shouldUseDarkColors: false }
+}))
+
+import { shouldInjectMcpSidebar } from './attach'
+
+describe('shouldInjectMcpSidebar', () => {
+  it('injects when the attach is live, the flag is enabled, and the view is alive', () => {
+    expect(shouldInjectMcpSidebar({ attachActive: true, enabled: true, destroyed: false })).toBe(
+      true
+    )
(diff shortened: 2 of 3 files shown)


src/main/lib/experiments.ts
```diff
@@ -32,7 +32,13 @@ import { configDir } from './paths'
...
 import type { FeatureFlagValue } from './telemetry'
 
-const DEFAULT_TIMEOUT_MS = 1500
+// The boot fetch is fire-and-forget (`void initExperiments`) and its only
+// awaiter is `getFlagAsync`, which wants the real value. A short timeout here
+// doesn't protect boot — it just discards a slow-but-successful fetch, leaving
+// the cache unseeded and every `getFlagAsync` reader stuck on `undefined`. So
+// the cap is a hang-guard, not a UI budget: long enough that a slow network
+// still resolves and seeds the cache.
+const DEFAULT_TIMEOUT_MS = 10_000
```
deepme987, line 41: 10s sounds a lot for this.

deepme987 (commented):
it might be easier to either make the feature flag on by default; OR remove it entirely.
The flag was meant to be a failsafe in case things go wrong, not something we expect to be leveraging anyway

src/main/lib/experiments.ts
MaanilVerma: Fair flag. The thing is nothing actually *waits* on this for 10s — it's not a UI budget, it's a hang-guard on a fire-and-forget background fetch. `initExperiments` is `void`-ed at boot, and the only awaiter (`getFlagAsync`, used by the MCP-sidebar gate) sits inside a fire-and-forget IIFE, so nothing blocks dom-ready or the canvas. The timeout is a `Promise.race` that resolves `{}` and clears the timer the instant flags come back, so a fast fetch returns immediately — the 10s only elapses on a genuinely slow/dead network.

Why I bumped it up from 1500ms: a short cap here doesn't speed anything up, it just discards a slow-but-successful fetch and leaves the flag cache unseeded, so `getFlagAsync` returns `undefined` and the feature silently defaults off. Worst case at 10s is just the sidebar icon showing a few seconds late on a cold first launch.

That said, 10s is admittedly arbitrary — happy to drop it to ~5s if that feels less alarming. Still comfortably generous for a slow fetch, and the ceiling only matters on a dead network anyway. Wdyt?

MaanilVerma: To be clear I think the timer itself is the right call here — a timer bounding an otherwise-unbounded wait is exactly the correct primitive, not something to work around. The 10s isn't a UI wait (nothing blocks on it, as above), it's a hard ceiling so `getFlagAsync` can never hang forever.

posthog-node does have its own `requestTimeout` (default 10s, real AbortController), so in the happy path they line up — but I'd rather keep our explicit timer than lean on the SDK's internal default, since that default could quietly change on a version bump and strand every `getFlagAsync` reader. Our own cap keeps that guarantee independent of the SDK.

Net: I'll keep the timer. Only open question is the number — 10s is deliberately generous for a slow-but-successful fetch (a short cap just discards it and leaves the flag cache unseeded), but I'm happy to bring it to ~5s if you'd prefer a tighter ceiling. Your call.

deepme987: so will this render the canvas, just wait up to 10s for a response?
and when that response comes in, it'll render the late-render the icon?

MaanilVerma: Yep, exactly that on both counts.

The canvas doesn't wait at all. This whole block is inside the `dom-ready` handler, which fires *after* the page has already painted, and the flag check is a fire-and-forget `void (async () => …)()` — so `onDomReady` kicks it off and returns immediately. The canvas is fully interactive while that `await getFlagAsync` sits off to the side.

When the flag resolves (instant on a warm cache, up to the 10s ceiling on a cold + slow network), the task wakes, re-checks the gate, and `executeJavaScript`-injects the sidebar button into the live page — so the icon just late-renders on top of the already-rendered canvas.

So worst case is purely: canvas up instantly, MCP icon pops in a few seconds later. Never a delayed/blank canvas.

deepme987 (approved):

src/main/lib/experiments.ts
deepme987: okay sgtm

PR #1465, opened 2026-08-26
feat: expose matched payloads for operational flags

benceruleanlu:
## Summary

Operational flags currently expose only a boolean or variant, so a caller cannot read the JSON configuration matched to that same evaluation. Return the value and payload together through `getOpsFlagResult`, and pass both to each `makeOpsFlag` parser.

This is the shared foundation for the Desktop Core beta stack. It introduces no canary enrollment, launch-argument changes, UI, or flag initialization. Existing `cloudFreeRuns` behavior is preserved: one boot-time fetch, an awaiting accessor, and the caller's fallback on unavailable or unrecognized configuration. Operational reads retain their existing consent-independent behavior and disable implicit PostHog feature-flag capture.

## Stack

1. **This PR:** shared operational-flag payload plumbing.
2. #1487: versioned Core beta grants, the beta setting/UI, launch-time injection, persistence/retraction, and launch attribution.
3. #1488: assets event log parsing and telemetry delivery; paired with Comfy-Org/ComfyUI#16082.

The canary is introduced by #1487 as one behavior change, rather than enrolling installs in this base and replacing that behavior in the next PR. Rollout decisions remain tracked in [DESK2-152](https://linear.app/comfyorg/issue/DESK2-152) and the [canary RFC](https://app.notion.com/p/3d06d73d36508177a895c6ff630111fa).

## Validation

- Focused Vitest: `telemetry.test.ts`, `opsFlag.test.ts`, and `cloudFreeRuns.test.ts` — 141 passed.
- Repository pre-commit checks: typecheck (node, web, E2E, integration), lint, and formatting passed.
- `git diff --check` passed.

Coverage verifies matched payload forwarding, disabled results overriding stale variants, disabled implicit capture, existing fallback behavior, and sharing/awaiting the boot fetch. No live PostHog rollout was performed.

## Change breakdown

| Category | Files | Paths | Added | Deleted | Share of changed lines |
| --- | ---: | --- | ---: | ---: | ---: |
| Product code | 2 | `src/main/lib/opsFlag.ts`<br>`src/main/lib/telemetry.ts` | 36 | 29 | 31.7% |
| Test code and fixtures | 3 | `src/main/lib/cloudFreeRuns.test.ts`<br>`src/main/lib/opsFlag.test.ts`<br>`src/main/lib/telemetry.test.ts` | 97 | 43 | 68.3% |

[...]

PR #1487, opened 2026-09-04
feat: launch-time version-gated core beta flags

synap5e:
## Summary

Desktop needs a retractable Core beta cohort that can include existing installations. On each Core launch, select allowlisted grants from `desktop_core_beta_features` against the installed Core version and the beta setting, filter them through Core's supported-argument schema, and inject the resulting flags in memory. Installation records are not rewritten.

Based on #1465, which now contains only shared operational-flag payload plumbing. This PR introduces the canary policy, initialization, settings/UI, launch integration, and attribution together. #1488 adds the assets-specific telemetry consumer.

## Grant and launch contract

```json
{"flags":[{"arg":"--enable-assets","min_core_version":"0.3.80"}]}
```

The version above is an example, not a rollout target. Arguments include leading dashes and must be in `CORE_CANARY_ALLOWED_FLAGS` (`--enable-assets`, `--enable-asset-hashing`, `--disable-assets`). `min_core_version` is required and inclusive; optional `max_core_version` is exclusive and requires an exact release-tag match. Strict version parsing prevents numeric-leading commit hashes being treated as releases. Malformed or unsupported grants apply nothing; the earlier bare-string payload format grants nothing.

`--disable-assets` does not exist in Core yet. It is the planned remote force-off for when assets go default-on, so the canary can disable rather than only stop granting enable; until Core accepts it, a grant is filtered by the supported-argument schema and reported as `dropped_unsupported`.

**The canary may only add flags.** It has no authority over the user's own launch arguments: an argument the user set is never removed, rewritten, or overridden, and the running Core's argument schema stays the one thing that drops a user token. A grant is withheld when the user already supplied that argument, and also when they supplied its opposite — paired mechanically across the `--enable-`/`--disable-` prefixes, in both directions — so the user's own argument always wins. The same pairing applies between grants: a payload naming both a flag and its opposite is refused outright rather than half-applied, since that is an operator mistake and applying either would pick a silent winner from payload order. Contradictory flags therefore never reach one command line, from either source. Consequently assets can run without any grant, in which case the launch correctly reports no beta flags and the install is not counted in the cohort. Discovery failure injects nothing and launches the user's arguments unchanged.

## Beta participation

- A dedicated **Opt in to beta features** setting controls participation, separately from telemetry reporting.
- Existing installations seed the setting once from their current telemetry choice, then persist it independently. This is inherited enrollment, not a fresh consent prompt.
- First use exposes a separate beta checkbox that follows the telemetry choice until edited.
- Entering beta requires telemetry consent, enforced in the settings handler as well as the UI. Leaving is always allowed. Disabling telemetry after joining stops reporting but does not automatically remove beta participation.

These are the implemented defaults; the rollout policy discussion remains open in the [current assets thread](https://comfy-organization.slack.com/archives/C0BSA9Z8H8V/p1789333445064529).

## Retraction and offline behavior

PostHog is fetched once per Desktop process; local eligibility is evaluated for each Core launch. A remote change is observed on Desktop restart, not by merely relaunching Core in the same process. An explicit OFF/control result retracts grants and persists the revoked state. A timeout, outage, or missing/deleted flag retains the last successful cached result. Operators must disable the flag and let clients receive that result before deleting it; this is not a bounded offline kill switch. The local beta toggle remains an exit without changing telemetry consent.

## Attribution

One post-schema result drives the final arguments, visible launch-log lines, `core_beta.applied` (including `dropped_unsupported`), `core_beta.opt_state`, and the `core_beta_flags` context on execution/hardware telemetry. Events report what survived filtering, rather than the requested payload. Emission remains consent-gated centrally.

## Validation

- Focused Vitest across canary/ops flags, telemetry, free-runs, versions, settings, settings IPC, launch, execution tap, and hardware tap: 275 passed on the five files touched by the latest changes (canary, launch, ops flag, telemetry, free runs); the wider focused suite passes with 1 pre-existing skip. The scanner-error assertion introduced on main was updated to include the beta context.
- Repository typecheck, lint, and formatting passed; `git diff --check` passed.
- Existing E2E coverage is included, but Electron/Windows/Linux E2E and live canary/telemetry delivery were not rerun for the latest changes. Those rollout checks remain outstanding.
- Two behaviours were added after the live canary run below, so that run does not cover them: the late-result persistence path, and the removal of user-argument stripping. Both are covered by focused tests, including a case that reproduces the field failure the persistence fix addresses.

## Live verification

A packaged build of this stack was exercised on a fresh Windows 11 install against a controlled feature-flag response, covering every cell of the grant state machine: applied, persisted across an unreachable fetch, revoked on an explicit `false`, staying revoked, and re-granted. Version gating, opt-state gating and cohort attribution behaved correctly, and the assets database survived a revoke/re-grant cycle intact. Details are in the comments below.

That run also measured the cold flag fetch at ~2572 ms against a 2000 ms deadline, which is what the late-result persistence change addresses: an explicit revocation that loses the boot race now lands on the following launch rather than being discarded.

## Change breakdown

| Category | Files | Paths | Added | Deleted | Share of changed lines |
| --- | ---: | --- | ---: | ---: | ---: |
| Product code | 18 | `src/main/index.ts`<br>`src/main/lib/coreCanary.ts`<br>`src/main/lib/executionTap.ts`<br>`src/main/lib/hardwareTap.ts`<br>`src/main/lib/ipc/registerSettingsHandlers.ts`<br>`src/main/lib/ipc/sessionActions/launch.ts`<br>`src/main/lib/opsFlag.ts`<br>`src/main/lib/telemetry.ts`<br>`src/main/lib/version.ts`<br>`src/main/popups/titlePopup.ts`<br>`src/main/settings.ts`<br>`src/preload/comfyTitlePopupPreload.ts`<br>`src/renderer/src/comfyTitlePopup/GlobalSettingsView.vue`<br>`src/renderer/src/comfyTitlePopup/TitlePopupApp.vue`<br>`src/renderer/src/views/FirstUseTakeover.vue`<br>`src/renderer/src/views/comfyUISettings/BooleanToggle.vue`<br>`src/renderer/src/views/comfyUISettings/SettingsSectionList.vue`<br>`src/types/ipc.ts` | 898 | 48 | 24.0% |
| Test code and fixtures | 17 | `e2e/core-beta-optin.test.ts`<br>`src/main/lib/cloudFreeRuns.test.ts`<br>`src/main/lib/coreCanary.test.ts`<br>`src/main/lib/executionTap.test.ts`<br>`src/main/lib/hardwareTap.test.ts`<br>`src/main/lib/ipc/registerSettingsHandlers.test.ts`<br>`src/main/lib/ipc/sessionActions/launch.test.ts`<br>`src/main/lib/opsFlag.test.ts`<br>`src/main/lib/telemetry.test.ts`<br>`src/main/lib/version.test.ts`<br>`src/main/popups/titlePopup.test.ts`<br>`src/main/settings.test.ts`<br>`src/preload/comfyTitlePopupPreload.test.ts`<br>`src/renderer/src/comfyTitlePopup/GlobalSettingsView.test.ts`<br>`src/renderer/src/comfyTitlePopup/TitlePopupApp.test.ts`<br>`src/renderer/src/views/FirstUseTakeover.test.ts`<br>`src/renderer/src/views/comfyUISettings/SettingsSectionList.test.ts` | 2959 | 19 | 75.7% |
| Localization | 2 | `locales/en.json`<br>`locales/zh.json` | 8 | 2 | 0.3% |

Total changed lines: **3934** (added + deleted), measured against this PR's immediate base. Changes brought in only by merging the base are excluded. No generated files or vendored code. No lockfile changes.


src/main/lib/ipc/sessionActions/launch.ts
```diff
@@ -694,7 +852,18 @@ async function runLaunch(
           }
         }
 
-        launchCmd.args = [...prefixArgs, ...desktopFlagArgs, ...filtered]
+        const built = buildLaunchArgs({
```
huang47, line 855: Strip the canary-managed tokens before the schema-discovery try/catch. Currently that only happens inside `buildLaunchArgs`, so when `getComfyArgsSchema` fails or times out, the fallback launches the original saved `--enable-assets` even after beta opt-out or remote revocation.

Reproduced at `0ac73f3` using the existing `handleLaunch` test harness: set `schemaThrows = true`, `betaEnabled = false`, `grants = []`, and include `--enable-assets` in the saved launch command. Launch succeeds, but `expect(spawnArgs).not.toContain('--enable-assets')` fails. Please cover this fallback path alongside the successful-discovery stripping tests.

src/main/lib/version.ts
```diff
@@ -40,6 +43,39 @@ export function formatComfyVersion(v: ComfyVersion | undefined, style: 'short' |
...
+ * Validation is deliberately whole-token (`semver.valid`, never
+ * `semver.coerce`) and fail-closed. Coercion would break the gate two ways:
+ * a git install stores only the first 8 commit chars in `version`
+ * (`src/main/sources/git.ts`), so a numeric-leading SHA like `61e5e3b5` would
+ * coerce to `61.0.0` and satisfy any minimum; and `0.3.80-rc.1` would coerce
+ * to stable `0.3.80`, enrolling prereleases that a `>=0.3.80` range must
+ * exclude. Keeping prereleases intact lets semver order them correctly.
+ */
+export function coreSemver(inst: InstallationRecord): string | null {
+  const raw = inst.comfyVersion?.baseTag ?? inst.version
```
benceruleanlu, line 59: Potential version-eligibility edge case; leaving severity to the author.

A build that really contains 0.17.2 plus additional patches correctly satisfies a 0.17.2 minimum. The concern is that `baseTag` does not always establish that relationship. In `resolveLocalVersion()`, when the newer release is not an ancestor of the installed code and the comparison cannot establish an equivalent release, the fallback can still assign `latestTagName` to `baseTag` and count changes from the shared earlier code.

For example, a release branch could add fix A after 0.17.0 to produce 0.17.2, while a development build adds change B without fix A. That fallback can label the development build as 0.17.2 + changes, and this new check then accepts a minimum of 0.17.2. Checking whether Core recognizes the startup flag would not detect a missing later fix.

A focused test using mocked Git results reproduced that fallback selecting the grant. I have not demonstrated a real release/installation affected by it or an actual feature failure. If this matters for the rollout, consider using a verified release base for eligibility rather than the display label. Approved overall; severity and follow-up are the author's call.

benceruleanlu (approved):
Approved. I left one concern about the version label used for beta eligibility, including the limits of the reproduction. We will leave severity and whether to address it in this PR to the author.

src/main/lib/telemetry.ts
```diff
@@ -1499,13 +1502,14 @@ export async function getOpsFlagResult(
       timer = setTimeout(() => resolve(undefined), timeoutMs)
     })
     const result = await Promise.race([flagPromise, timeoutPromise])
-    if (!result) return undefined
+    if (!result) return { kind: 'unreachable' }
```
christian-byrne: issue: Persist an explicit result that arrives after this deadline. `Promise.race` returns `unreachable` here, and no continuation observes `flagPromise`, so a late `false` never reaches `writePersistedResult`. With cached ON and a slow fetch on each boot, the grant can survive every restart. Keep the 2,000 ms deadline for the current launch, but separately persist an eventual explicit value for the next launch. Timeout, missing, and error outcomes must not overwrite the cache. A regression case should resolve OFF after the deadline, then verify the next unreachable launch reads cached OFF; late missing/error controls must preserve the last good cache.

christian-byrne (commented):
- Surface late-result persistence inline.
- [Version-eligibility review already tracked](https://github.com/Comfy-Org/Comfy-Desktop/pull/1487#discussion_r4011386707).

This forwards the existing finding into a resolvable review thread, not a new whole-PR approval. Verified the timeout and cache-write paths against the current head and its stacked base. Source establishes discarded late results. [Simon's Windows timing report](https://github.com/Comfy-Org/Comfy-Desktop/pull/1487#issuecomment-5673627565) remains reported runtime evidence, not independently reproduced here.

Glossary: version eligibility means whether the installed Core release qualifies for a flag; explicit result means a definite enabled/disabled response rather than an unknown or failed request.
</details>

christian-byrne (commented):
Late-result fix verified.
Manual-argument policy needs acceptance.
Version provenance remains open.

Re-reviewed the changed flag selection, launch assembly, telemetry deadline and persistence code at this exact head. Independently ran 278 tests across telemetry, opsFlag, coreCanary, cloudFreeRuns and launch; all passed. The late explicit false reaches disk for the next launch without changing this launch. Late missing/error results do not overwrite the cache. The existing late-result thread is addressed in code and tests; this is not current-package Windows or remote-response evidence.

Manual same/opposite arguments now take precedence and contradictory remote payloads grant nothing. That matches the author's new additive-only behavior, but reverses our former stripping expectation. I am recording it as an acceptance change, not silently waiving the old test. Beta OFF no longer implies Assets OFF when the user supplied --enable-assets. The existing version-provenance finding remains unaddressed; no duplicate comment.

The main merge conflict is repaired. This is a targeted COMMENT review, not whole-stack approval or authorization to release. The current child package still needs late-response/manual-argument runtime checks.

Glossary: provenance means verified release ancestry; additive-only means remote grants may add arguments but do not remove a user's explicit arguments.

</details>

src/main/lib/version.ts
christian-byrne: Keep the ancestry-only rule in [the provenance fix](https://github.com/Comfy-Org/Comfy-Desktop/pull/1537). Descendant builds remain eligible. The 149 focused tests pass locally.

I checked `findBestBackportTag`, the resolution paths, launch gating and provenance-only persistence at [the reviewed head](https://github.com/Comfy-Org/Comfy-Desktop/commit/f883bc3adf858e809684086d374e78012a4e92ce). Its `unique <= branchDist` comparison can tolerate unmatched release commits. Treating that result as ancestry-verified would retain the original over-reporting risk.

Even `unique === 0` establishes patch equivalence, not ancestry. I would keep it unverified for this field. Ordinary descendants remain eligible when their base tag is reachable; my original suggestion did not require `commitsAhead === 0` for the lower bound.

The explicit comparison of `baseTagVerified` also handles re-persisting an old record whose displayed version has not changed. I ran the version-resolve, version, coreCanary and launch suites at the exact head: 149 tests passed. This answers the source-policy question; it is not packaged Windows verification or approval of the whole stack.

Glossary: ancestry means the release commit is in the installed build's history; patch equivalence means matching changes without necessarily sharing that history.

</details>

christian-byrne:
Manual launch flags should survive opt-out.

Settling the remaining acceptance question: beta opt-out should remove remotely granted flags, not manually supplied launch arguments. Preserve explicit user flags, including `--enable-assets`, even when beta enrollment is off. This is the intended policy, not a request to change that behavior.

This updates the earlier [manual-flag decision request](https://github.com/christian-byrne/blocked-on-christian/issues/485#issuecomment-5690479615). It does not waive consent checks or approve a production rollout.

Glossary: opt-out means disabling enrollment in remotely managed beta features.

</details>

christian-byrne:
Enabled beta clicks survive restart. Both directions verified.

Packaged Linux check: open Desktop Settings through the title menu, assert the beta switch is enabled, click OFF to ON, verify its accessible state and settings file, fully close Desktop, then restart without settings or installation seeds. ON remains. Click ON to OFF and repeat; OFF remains. Parent rerun passed in 4.1 seconds. No direct settings IPC call substituted for the clicks.

Executed on the current [stacked provenance head](https://github.com/Comfy-Org/Comfy-Desktop/commit/74e4456ebb51845bb41eb327ba1766157f12a9c3), staging app.asar SHA-256 `16b5303f9f2f8d1bac74acbf8eb6cd829f7f5e4302c4572639b94ae9fe36a805`. A direct Git-object comparison against [this PR head](https://github.com/Comfy-Org/Comfy-Desktop/commit/16471d675c378fe63c1b3c09058e75f334f408ed) found GlobalSettingsView.vue, BooleanToggle.vue, main/settings.ts, main/index.ts and preload/index.ts identical. That is bounded source equivalence, not an exact lower-head build or Windows claim.

Retained harness: `fleet/qa/toggle-2026-09-17-1237/toggle-persistence.test.ts`; assertions and readouts in `evidence/result.json`; parent log `research/qa/toggle-parent-2026-09-17-1237.log`. Every spawn used the authenticated staging guard. No product source change or formal approval.

Terms: IPC means renderer-to-main communication; unseeded restart means persisted files, not injected test settings, supply state.

</details>

christian-byrne (approved):
No blocking findings at this head. Manual launch flags survive beta opt-out by policy; ancestry enforcement follows in [the provenance fix](https://github.com/Comfy-Org/Comfy-Desktop/pull/1537). The existing [development-only cache race](https://github.com/Comfy-Org/Comfy-Desktop/pull/1487#discussion_r4021988658) remains nonblocking, not fixed.

PR #1517, opened 2026-09-10
fix: Preserve workspace context when returning to dashboard

alexisrolland:


MaanilVerma (approved):

src/renderer/src/views/ChooserView.vue
```diff
@@ -100,6 +100,24 @@ defineExpose({ activeFilter })
...
 let dashboardScopeInitialized = false
 
+async function initializeDashboardScope(): Promise<void> {
+  const [persistedWorkspaceId] = await Promise.all([
+    window.api.getSetting(DASHBOARD_WORKSPACE_SETTING).catch(() => undefined),
+    authStore.fetchStatus().catch(() => authStore.status)
+  ])
+  selectedWorkspaceId.value =
+    authStore.isSignedIn && typeof persistedWorkspaceId === 'string' && persistedWorkspaceId.trim()
+      ? persistedWorkspaceId
```
benceruleanlu, line 110: **Stale saved workspace:** A saved ID can belong to a team the user has left, leaving an empty/mislabelled dashboard and passing a stale ID to New Instance. #1555 validates membership in the shared scope store and persists a valid fallback; failed membership requests preserve the selection.

src/renderer/src/views/ChooserView.vue
```diff
@@ -100,6 +100,24 @@ defineExpose({ activeFilter })
...
+      : workspaceContextId(authStore.status)
```
benceruleanlu, line 111: **Dashboard/New Instance mismatch:** Without a saved preference, the dashboard selects the authenticated team without persisting it, while menu-driven New Instance defaults to Personal. #1555 persists the fallback and makes both entry points read the same initialized scope store, keeping them consistent even if saving fails.

benceruleanlu (commented):
Two workspace restoration issues addressed by the stacked follow-up #1555.

PR #1524, opened 2026-09-13
feat(telemetry): compare Assets boot performance by cohort

christian-byrne:
## Summary

- tag `boot_started`, `boot_completed`, and `boot_failed` with applied Core beta flags and an explicit `assets_enabled` cohort
- derive the cohort from post-version/post-schema grants, so it reflects actual runtime state rather than opt-in intent
- document privacy constraints and reproducible PostHog queries for duration and success rate by Desktop version and Assets state

## Integration boundary

This PR intentionally measures only the Desktop-owned ComfyUI boot lifecycle. It does not add Core-side instrumentation or collect asset metadata.

`app_version` remains centrally attached by `src/main/lib/telemetry.ts`. The new properties contain only a boolean and Desktop-controlled CLI flag names.

## Dependency

This branch is based on and targets [Desktop PR 1487](https://github.com/Comfy-Org/Comfy-Desktop/pull/1487) (`synap5e/feat/core-beta-launch-gate`) because that PR owns the version-gated, applied Core-beta launch state. It should be rebased/retargeted to `main` after that dependency lands.

Related integration work: [Desktop PR 1485](https://github.com/Comfy-Org/Comfy-Desktop/pull/1485).

## Query evidence

`docs/assets-performance-telemetry.md` contains the metric contract and ready-to-run HogQL for:

1. mean/p50/p95 boot duration grouped by `app_version` and `assets_enabled`
2. boot starts/completions/failures and success rate grouped by the same dimensions

## Change breakdown

| Category | Files | Paths | Added | Deleted | Share of changed lines |
| --- | ---: | --- | ---: | ---: | ---: |
| Product | 1 | `src/main/lib/ipc/sessionActions/launch.ts` | 15 | 0 | 9.9% |
| Tests | 1 | `src/main/lib/ipc/sessionActions/launch.test.ts` | 70 | 0 | 46.1% |
| Docs | 1 | `docs/assets-performance-telemetry.md` | 67 | 0 | 44.1% |
| Total | 3 | | 152 | 0 | 100% |

## Verification

- `pnpm exec vitest run src/main/lib/ipc/sessionActions/launch.test.ts` (55 passed)
- pre-commit full typecheck (node, web, e2e, integration)
- pre-commit ESLint
- pre-commit Prettier check

Merge requires Christian's explicit approval.


src/main/lib/ipc/sessionActions/launch.ts
```diff
@@ -668,6 +668,18 @@ async function runLaunch(
...
 
+  /** Applied launch state shared by every boot-lifecycle event. This is deliberately
+   *  derived from the post-schema grants rather than the opt-in toggle: an opted-in
+   *  launch on an older Core may still have Assets disabled. `app_version` is added
+   *  centrally by telemetry.ts. */
+  function bootCohort(): { core_beta_flags: string[]; assets_enabled: boolean } {
+    const coreBetaFlags = coreBeta.applied.map((grant) => grant.arg)
+    return {
+      core_beta_flags: coreBetaFlags,
+      assets_enabled: coreBetaFlags.includes('--enable-assets')
```
christian-byrne, line 679: issue: On the updated launch-gate base, manual --enable-assets survives beta opt-out and yields coreBeta.applied = []. This expression then reports assets_enabled=false although Core runs Assets, putting enabled boots in the disabled performance cohort. Keep core_beta_flags as grant attribution, but derive the effective Assets measurement independently from the final launch state, including manual arguments and Core defaults. Add a case with manual --enable-assets and no grant when integrating the current base.

christian-byrne (commented):
Assets measurement needs current-base integration.

Targeted self-review of the three-file boot-cohort change against the updated launch-gate behavior. This draft still uses the old base. The inline finding covers semantic integration needed before using these events for performance comparisons; no replacement PR is needed. No packaged execution of this draft is claimed.

Glossary: cohort means the group used to compare boot measurements; effective arguments are what Core actually receives.

</details>

src/main/lib/ipc/sessionActions/launch.ts
```diff
@@ -509,6 +650,36 @@ async function runLaunch(
...
+
+  /** Applied launch state shared by every boot-lifecycle event. This is deliberately
+   *  derived from the post-schema grants rather than the opt-in toggle: an opted-in
+   *  launch on an older Core may still have Assets disabled. `app_version` is added
+   *  centrally by telemetry.ts. */
+  function bootCohort(): { core_beta_flags: string[]; assets_enabled: boolean } {
+    const coreBetaFlags = coreBeta.applied.map((grant) => grant.arg)
+    return {
+      core_beta_flags: coreBetaFlags,
+      assets_enabled: coreBetaFlags.includes('--enable-assets')
```
synap5e: **This derivation stops being correct once the branch is rebased onto `main`.**

`main` reversed the design this branch was written against. The `stripCanaryArgs` function that made the canary the sole authority over these tokens has been deleted, and `coreCanary.ts` now states the opposite rule explicitly: *"This system may only ADD args. It has no authority over the user's own launch arguments and never removes or overrides one — several of these flags are first-class, user-settable options in Desktop's launch-args UI."* `selectCoreCanaryArgs` enforces it with `if (presentArgs.has(arg)) continue`.

So on `main`, a user who types `--enable-assets` into the launch-args UI has the grant **withheld** — `coreBeta.applied` won't contain it, and this line reports `assets_enabled: false` while Assets is genuinely on. No failure condition required; that's the success path. Every hand-configured install lands in the wrong cohort.

Suggested fix: derive the boolean from the final spawned arguments, constrained to `CORE_CANARY_ALLOWED_FLAGS` so no user-typed token can ever reach telemetry:

```ts
assets_enabled: (launchCmd.args ?? []).includes('--enable-assets')
```

This is safe to read here — every mutation of `launchCmd.args` downstream of arg assembly is additive (the five `push` calls in `applyStorageLaunchArgs`, plus `setPortArg`); nothing splices or filters, and `bootCohort()` is first called well after the last of them. So reading final args can only turn a false negative into a true positive, never manufacture a false positive.

Please keep `core_beta_flags` sourced from `coreBeta.applied` — it correctly means "grants applied", and `executionTap.ts` / `hardwareTap.ts` emit it from the same source, so changing it here would break that comparability. The divergence between the two fields is itself useful signal.

Related: `main` also added `--disable-assets` to the allowlist as the planned remote force-off for when Assets goes default-on. Once that happens, a launch with no flags has Assets **on** but reports `false`, and this property silently inverts. Worth handling in the same change.

src/main/lib/ipc/sessionActions/launch.ts
```diff
@@ -694,7 +864,18 @@ async function runLaunch(
...
+          userArgs,
+          desktopFlagArgs,
+          schema,
+          betaFlags: await getCoreCanaryFlagsAsync(),
+          coreVersion: coreSemver(inst),
+          coreVersionExact: coreSemverExact(inst),
+          betaEnabled
+        })
+        launchCmd.args = built.args
+        coreBeta = built.beta
```
synap5e: This is the only place `coreBeta` is ever assigned, and it's inside the `try`. The `catch` immediately below is bare — its whole body is the comment `// Schema not available — pass args as-is.`

When `getComfyArgsSchema` rejects, `coreBeta` stays `noCoreBeta(betaEnabled)` **and** `stripCanaryArgs` never runs either, since it lives inside `buildLaunchArgs`. A `--enable-assets` sitting in the install record's user-editable `launchArgs` therefore survives onto the command line: Core boots with Assets on, and the boot reports `assets_enabled: false`, `core_beta_flags: []`.

Both conditions are realistic. The stored-flag half is documented rather than hypothetical — the `stripCanaryArgs` docstring exists specifically for *"a dogfood install whose stored `launchArgs` had a flag baked in at install time"*. The schema-rejection half covers a broken venv, a custom node that raises during `main.py --help`, or simply the 15 s `execFile` timeout in `runHelp`.

Deriving from the final spawned args (see the comment on `bootCohort()`) fixes this path too. Worth a regression test: `launchHarness.schemaThrows = true` plus an `--enable-assets` in the harness `launchCommand.args`.

docs/assets-performance-telemetry.md
```diff
@@ -0,0 +1,67 @@
...
+comparison uses the existing ComfyUI boot lifecycle, which is emitted for both Assets-enabled and
+Assets-disabled launches. It does not add instrumentation inside Core or collect asset metadata.
+
+## Cohort contract
+
+Each `comfy.desktop.comfyui.boot_started`, `boot_completed`, and `boot_failed` event carries:
+
+| Property          | Meaning                                                                                                      |
+| ----------------- | ------------------------------------------------------------------------------------------------------------ |
+| `assets_enabled`  | Whether `--enable-assets` was actually applied to this launch after Desktop's version and Core-schema gates. |
```
synap5e: "actually applied" is stated as an absolute, and there are two paths where it isn't true — a schema-discovery failure leaving a baked flag on the command line, and (after rebase onto `main`) any user who sets the flag by hand. Both noted on the code.

Either fix the derivation and keep this wording, or soften it here to describe what the field really is: whether the *canary granted* the flag.

docs/assets-performance-telemetry.md
```diff
@@ -0,0 +1,67 @@
...
+| `core_beta_flags` | All Core beta arguments actually applied to this launch.                                                     |
+| `app_version`     | Desktop version, attached centrally by `src/main/lib/telemetry.ts`.                                          |
+| `boot_id`         | Per-launch join key shared by the lifecycle events. Retries reuse the same key.                              |
+
+`assets_enabled = false` means the running launch did not receive the Assets argument, including an
```
synap5e: **This is the part I'd most want to resolve before merging.** The paragraph is honest about what the field is, but the queries below then use this bucket as the comparison baseline — and it isn't one.

`assets_enabled = true` requires four independent conditions to hold at once: the beta toggle resolves true, PostHog actually serves a grant to this client, the install's Core semver sits inside the grant window, and the running Core's `--help` schema knows the flag. `false` is *everything else* — five distinct causes collapsed into one value.

The Core-version skew isn't incidental, it's structural. `parseCoreCanaryFlags` drops any grant without a parseable `min_core_version`, so every grant that can ever apply carries a version floor. The Assets-on arm is bounded below by that floor; the Assets-off arm contains, by construction, every install beneath it. The two distributions can only match if no install is below the floor — which would make the gate pointless. Boot time and boot-failure rate both move with Core version, so that's a confound sitting directly on the metric.

And nothing on these events lets an analyst narrow the control arm:

- `variant` is the standalone build flavour, a frozen install-time fingerprint — not a Core version.
- `core_beta.opt_state` (`opted_in`) and `core_beta.applied` (`core_version`) do carry the missing facts, but neither carries `boot_id` or `installation_id`, so they can't be joined to a boot.
- `core_beta.applied` only fires when something applied or was dropped, so a fully opted-out boot emits no Core version at all.

Minimum fix: add `core_beta_opted_in` and `core_version` to `bootCohort()`. Both are already on the `CoreBetaLaunch` it reads and both already ship on other events, so there's no new privacy surface and no new plumbing.

That still doesn't fully establish eligibility, though — selection also checks an upper bound, exact/verified/current tag status, and schema support. If we want a genuinely comparable control, the cleaner shape is an explicit ineligibility *reason* (below-floor / above-ceiling / unverified / schema-refused / not-granted) rather than one boolean absorbing all five.

docs/assets-performance-telemetry.md
```diff
@@ -0,0 +1,67 @@
...
+opted-out launch or one whose Core version/schema did not support the grant. This is intentional:
+the comparison is actual runtime state, not user intent.
+
+The normal telemetry consent gate still applies. No paths, filenames, asset names, prompts, model
+metadata, or other user content are added. This follows the telemetry privacy rules documented in
+`src/main/lib/telemetry.ts` and ADR-029.
```
synap5e: `ADR-029` doesn't exist in this repo — it appears exactly once, on this line. There's no `docs/adr/` directory and no other ADR reference in any `.md`, `.ts` or `.json`. Anyone following the citation to audit the privacy claim has nothing to follow.

The `telemetry.ts` half is real and does carry the rule being honoured here. Either link the ADR if it lives in an external wiki, or drop it and cite the `telemetry.ts` privacy block directly.

docs/assets-performance-telemetry.md
```diff
@@ -0,0 +1,67 @@
...
+
+## PostHog queries
+
+Boot duration by Desktop version and applied Assets state:
+
+```sql
+SELECT
```
synap5e: `app_version` is the **Desktop** version. It controls for none of the four conditions that decide `assets_enabled`, and in particular not for Core version — which is the one that's structurally skewed between the arms (see the comment on the cohort contract above).

As written, this query answers "newer-Core, opted-in installs versus everyone else", not "Assets on versus Assets off". That's worth stating in the doc even after the extra fields are added, since the comparison stays observational either way.

docs/assets-performance-telemetry.md
```diff
@@ -0,0 +1,67 @@
...
+WHERE event = 'comfy.desktop.comfyui.boot_completed'
+  AND timestamp >= now() - INTERVAL 14 DAY
+GROUP BY desktop_version, assets_enabled
+ORDER BY desktop_version DESC, assets_enabled DESC
+```
+
+Boot outcome by Desktop version and applied Assets state:
+
+```sql
+SELECT
```
synap5e: Same grouping problem as the duration query — and it bites harder here, because boot *failure* correlates with exactly the install conditions that also push a boot into the `assets_enabled = false` bucket.

The schema-discovery failure path is the concrete case: installs where `main.py --help` times out or raises are disproportionately the slow or damaged ones, and those boots are both more likely to fail and guaranteed to be tagged `false`. That's a plausible mechanism for the control arm's failure rate to be inflated by construction. I wouldn't claim a magnitude, but the direction is the one that flatters Assets, which is the uncomfortable direction for a decision metric.

The `uniqIf(properties.boot_id, ...)` dedup is correct, though — `boot_id` is generated once per logical boot and reused across port/reboot retries, so retry-inflated `boot_started` counts don't skew the rate. Worth keeping.

docs/assets-performance-telemetry.md
```diff
@@ -0,0 +1,67 @@
...
+  'comfy.desktop.comfyui.boot_started',
+  'comfy.desktop.comfyui.boot_completed',
+  'comfy.desktop.comfyui.boot_failed'
+)
+  AND timestamp >= now() - INTERVAL 14 DAY
+GROUP BY desktop_version, assets_enabled
+ORDER BY desktop_version DESC, assets_enabled DESC
+```
+
+Use the first query as a trend grouped by `assets_enabled`; use the second as a table. Do not compare
```
synap5e: This caveat covers sample size but not the bigger problem: the two cohorts aren't comparable populations in the first place.

Suggest replacing/extending it with an explicit statement that `assets_enabled = false` is a population rather than a control, that the arms differ systematically in Core version, and that differences between them should not be read as the causal effect of Assets without further segmentation.

src/main/lib/ipc/sessionActions/launch.test.ts
```diff
@@ -256,3 +365,658 @@ describe('handleLaunch model-download startup await (#1322)', () => {
...
+          if (typeof payload?.text === 'string') sent.push(payload.text)
+        }
+      }
+    } as unknown as Electron.IpcMainInvokeEvent,
+    installationId,
+    inst: harnessInstall(),
+    actionData: {}
+  })
+
+  beforeEach(() => {
```
synap5e: `launchHarness.waitForPort` isn't reset here, though eight sibling fields are (`schemaThrows`, `betaEnabled`, `betaEnabledThrows`, `schemaNames`, `grants`, `duringResourceAcquire`, `spawn`, `launchCommand`).

The new terminal-failure test sets it to `async () => { throw new Error('boot timed out') }` and never restores it, so it persists for the remainder of the file. Nothing breaks today — both tests that follow run with `skipPortWait: true`, so the probe is never reached. But the next test added to this block on the normal port path would silently inherit a rigged boot failure with no cause visible at its own call site, which is the kind of thing AGENTS.md's zero-tolerance flaky-test line is aimed at.

One line: `launchHarness.waitForPort = null` alongside the other resets. That also makes the pre-existing leak from the port-conflict test safe.

src/main/lib/ipc/sessionActions/launch.ts
christian-byrne: Fixed in [the rebased revision](https://github.com/Comfy-Org/Comfy-Desktop/commit/859cb3f148397aa3185d8664de1c9bbf88080f5d): `assets_enabled` reads the final launch arguments independently of grants. Added opt-out/manual-flag and no-grant cases. The contract now explicitly measures an argument, not Core defaults or service health. Remote CI pending; local runtime deferred.

src/main/lib/ipc/sessionActions/launch.ts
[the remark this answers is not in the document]
christian-byrne: Fixed in [the rebased revision](https://github.com/Comfy-Org/Comfy-Desktop/commit/859cb3f148397aa3185d8664de1c9bbf88080f5d): the cohort reads final arguments, so preserved manual flags count even when schema discovery fails. Added both failure arms, with and without a manual flag. Remote CI pending; no new local test result claimed.

docs/assets-performance-telemetry.md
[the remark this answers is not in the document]
christian-byrne: Updated the derivation and contract in [the rebased revision](https://github.com/Comfy-Org/Comfy-Desktop/commit/859cb3f148397aa3185d8664de1c9bbf88080f5d). The field measures the final explicit argument, including manual flags, not grant presence or successful service initialization.

docs/assets-performance-telemetry.md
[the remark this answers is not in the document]
christian-byrne: Added opt-in and recorded Core version to all three boot events in [the rebased revision](https://github.com/Comfy-Org/Comfy-Desktop/commit/859cb3f148397aa3185d8664de1c9bbf88080f5d). Both queries segment by those fields. The docs explicitly retain the remaining eligibility/selection limits and make no causal claim; this does not add an eligibility-reason taxonomy.

docs/assets-performance-telemetry.md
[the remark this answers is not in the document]
christian-byrne: Removed the unavailable citation and linked the existing `telemetry.ts` privacy rules directly in [the rebased revision](https://github.com/Comfy-Org/Comfy-Desktop/commit/859cb3f148397aa3185d8664de1c9bbf88080f5d).

docs/assets-performance-telemetry.md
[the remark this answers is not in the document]
christian-byrne: The duration query now groups by Desktop version, recorded Core version, beta opt-in and the Assets argument in [the rebased revision](https://github.com/Comfy-Org/Comfy-Desktop/commit/859cb3f148397aa3185d8664de1c9bbf88080f5d). The surrounding text calls it observational and notes that version labels can lag a modified checkout.

docs/assets-performance-telemetry.md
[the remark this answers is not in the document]
christian-byrne: Applied the same segmentation to outcomes and retained distinct `boot_id` counts in [the rebased revision](https://github.com/Comfy-Org/Comfy-Desktop/commit/859cb3f148397aa3185d8664de1c9bbf88080f5d). Also documented recent incomplete boots, cancellations and time-window boundary effects; no claim that segmentation eliminates selection bias.

docs/assets-performance-telemetry.md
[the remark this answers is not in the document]
christian-byrne: The contract now says this is observational, identifies the mixed off population, and warns against attributing differences to Assets in [the rebased revision](https://github.com/Comfy-Org/Comfy-Desktop/commit/859cb3f148397aa3185d8664de1c9bbf88080f5d). Counts alone are not treated as evidence of comparability.

src/main/lib/ipc/sessionActions/launch.test.ts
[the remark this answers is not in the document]
christian-byrne: Added `launchHarness.waitForPort = null` to the per-case reset in [the rebased revision](https://github.com/Comfy-Org/Comfy-Desktop/commit/859cb3f148397aa3185d8664de1c9bbf88080f5d). Remote CI is running; local checks remain deferred under the host-resource restriction.

synap5e:
Happy to approve once tests resolved

synap5e (approved):

docs/assets-performance-telemetry.md
```diff
@@ -0,0 +1,97 @@
...
+  round(100 * boots_completed / nullIf(boots_started, 0), 2) AS success_rate_pct
+FROM events
+WHERE event IN (
+  'comfy.desktop.comfyui.boot_started',
+  'comfy.desktop.comfyui.boot_completed',
+  'comfy.desktop.comfyui.boot_failed'
+)
+  AND timestamp >= now() - INTERVAL 14 DAY
+  AND properties.assets_enabled IS NOT NULL
+GROUP BY desktop_version, core_version, beta_opted_in, assets_enabled
```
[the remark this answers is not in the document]
christian-byrne: The cited partial schema comes from this PR's earlier branch revisions, not an earlier `main` implementation. Both fields first reach `main` together in [the telemetry merge](https://github.com/Comfy-Org/Comfy-Desktop/commit/659896cd41083a3838dbc81c347267abe1991363); its parent has neither field.

<details><summary>Evidence and disposition</summary>

Checked the merge parent and the `main` history of both properties in `launch.ts`. The shared event-property helper now emits `assets_enabled` and `core_beta_opted_in` together. A row with an unknown opt-in value would form its own NULL group rather than join either known opt-in group.

This does not establish that old branch/test events are absent from PostHog; no production query was run. The extra predicate can explicitly exclude those rows, but the claim that an earlier published implementation emitted the partial schema is not supported by the inspected history. No product correction or rollout blocker established here.

</details>

PR #1537, opened 2026-09-16
fix(core-canary): refuse grants when the version record cannot support the check

synap5e:
## Summary

The Core beta version gate can accept an install that does not satisfy the window it is checked against. Two independent causes, both fixed here by refusing the grant:

1. **The label may name a release the install does not contain.** `resolveLocalVersion` can assign a newer tag on a path only reached because that tag is *not* an ancestor, and the gate reads that label. Found in review on #1487 by @benceruleanlu, reproduced there with mocked Git results.
2. **The record may no longer describe the checkout being launched.** Schema discovery keys on live git HEAD while the gate reads the persisted record; nothing reconciles them, so a `git pull` leaves the gate deciding on a superseded commit. Found by a whole-stack review.

Split out of #1487 so neither blocks it.

## The bug

In the backport branch of `resolveLocalVersion` (`src/main/lib/version-resolve.ts`), when cherry-pick detection fails the code falls back to the merge base:

```ts
const mergeBase = await findMergeBase(comfyuiDir, latestTagRef!, commit)
if (dist !== undefined) {
  baseTag = latestTagName   // the newer tag
  commitsAhead = dist       // counted from the merge base
}
```

That branch is only reached because `latestTagRef` is **not** an ancestor of `commit`. So the install is labelled `<newer release> + N`, `coreSemver()` returns that release, and `semver.gte(version, minCoreVersion)` passes a minimum the install does not meet.

Core's supported-argument schema absorbs the common case — an install without the feature does not know the flag, so the grant is dropped. What is left exposed is a `min_core_version` raised to require a later **fix** on an install that already has the flag, which is exactly how the minimum would be used to exclude a known-bad build mid-rollout.

## The change

`ComfyVersion` gains `baseTagVerified`, set on every return path in `resolveLocalVersion`:

| Path | Verified |
| --- | --- |
| Direct-ancestor upgrade | yes |
| `findNearestTag` / no upgrade | yes when a tag was found (`git describe` is ancestry-based) |
| `findBestBackportTag` success | **no** — see below |
| Merge-base fallback | no |
| Caller-supplied `fallbackTag` | no |

`coreSemverVerified()` reads it as `=== true`, so a record persisted before the field existed is unverified rather than trusted. `CoreVersionState.verified` is required rather than optional, so a call site cannot omit it and silently inherit the permissive behaviour, and `selectCoreCanaryArgs` withholds every grant when it is false.

**`findBestBackportTag` is treated as unverified.** It is cherry-pick aware, which makes "yes" tempting, but its acceptance rule tolerates up to `branchDist` of the tag's commits having no counterpart in the install — that is the same failure mode, bounded rather than unbounded. It also matches on patch-id, which is content equivalence rather than reachability. Fail closed instead.

**Display behaviour is unchanged.** The merge-base fallback still produces the same version string; this teaches the canary to distrust it, it does not remove it.

## The record may be stale (second cause)

`launch.ts` derives "which Core version is this" twice within one block, from two sources, and used the weaker one:

- `getComfyArgsSchema` keys on `readGitHead(...) ?? fallbackRevision` (`comfy-args.ts:374`) — **live HEAD wins**, the record is only a fallback.
- `coreSemver` / `coreSemverExact` / `coreSemverVerified` read **only** `inst.comfyVersion` (`version.ts:69, 86, 101`).

Nothing on the launch path reconciles them, and the resolver that refreshes the record is fire-and-forget and TTL-throttled. So on a git-managed install the record can name a commit that is no longer checked out, and `coreSemverExact` still returns true — `commitsAhead: 0` was true *of the superseded commit*.

The existing guards cannot catch this: both answer questions about the recorded commit, not about the checkout still being at it. And the argument-schema filter is asymmetric — an older Core does not know `--enable-assets` so the **minimum** bound is covered, but a newer Core still parses it, so the **maximum** bound had nothing behind it.

`coreRecordCurrent(inst, liveHead)` compares the two and `selectCoreCanaryArgs` refuses every grant when they disagree, with its own log line so an operator can tell the two refusals apart. A null HEAD (standalone archive installs, no `.git`) is not a contradiction and changes nothing; an unreadable record while HEAD is readable fails closed. Comparison is case-insensitive but whole-token, so an abbreviation does not satisfy it.

This is the cheap fail-closed fix. The structural one — have `getComfyArgsSchema` return the revision it keyed on so one notion of version per launch holds by construction — is deliberately not taken here.

## Re-persisting existing records

`_resolveAndBroadcastVersions` decided whether to persist by comparing formatted version strings, and `formatComfyVersion` does not read the new field. An install whose displayed version never changes would therefore keep the field absent forever — permanently refused by a gate that is working exactly as designed, with no error anywhere. The change test now also compares provenance.

This was the one part capable of silently shrinking the cohort to zero, so it is worth a second pair of eyes.

## Validation

- Written test-first: the unverified install received a grant before the fix, which is the reported bug reproduced as a unit test.
- 297 tests across `version-resolve`, `version`, `coreCanary`, `launch`, `cloudFreeRuns`, `assetsTap` and `opsFlag`; typecheck clean on all four projects; eslint, prettier and `git diff --check` clean.
- Two test defects found by the same review are fixed here: `coreCanary.test.ts` was writing a live `--enable-assets` grant into the real user config directory (it mocked `./telemetry` but not `./paths`), and `cloudFreeRuns.test.ts`'s fixtures predated the classified result type, so its "unreachable" case passed through an exception path rather than the classification it is named for.
- Full unit suite: 4929 passed, 2 skipped, 1 failure — `process.test.ts > skips a port that is actually in use` (`EADDRINUSE 127.0.0.1:49500`), proven pre-existing by stashing these changes and reproducing it identically.
- Disabling either refusal turns three tests red across both the unit and integration layers.
- No real installation affected by the original fallback has been demonstrated, here or in the original report.

## Note for operators

The gate now refuses an install whose base could not be verified, and logs `[core-canary] refused: base <version> not verified` when it had grants to withhold. Without that line the change would shrink the cohort with no operator-visible signal.


christian-byrne:
Changed-stack unit tests: 928 passed, one skipped.

Fresh Linux verification at [this exact tip](https://github.com/Comfy-Org/Comfy-Desktop/commit/74e4456ebb51845bb41eb327ba1766157f12a9c3): all 21 changed unit-test files across the launch, assets-tap and provenance stack passed in one serial invocation. Vitest reported 928 passed, one skipped, exit 0, 17.31 seconds. The tested commit was unchanged before and after.

Reproduction: collect `src/**/*.test.ts` paths changed between the merged foundation commit and this tip, then run `node_modules/.bin/vitest run --maxWorkers=1` with that explicit file list. Existing installed dependencies were reused; no dependency or product changes.

A preceding, broader 18-suite run at four workers had five 5-second timeouts in title-popup and settings-content tests. Both affected suites passed in an isolated one-worker run (58/58). Those failures remain recorded; the serial result does not establish their root cause or waive intermittent failures.

The focused six-suite launch/provenance/tap run also passed 313 tests. Counts overlap and must not be added. These are unit results at the stacked tip, not independent-base test runs for each lower PR, a formal review approval, packaged Windows verification, or rollout acceptance.

Glossary: tip means the latest commit at the top of the stack; serial means one test worker at a time.

</details>

christian-byrne:
- Current Linux package verified.
- Manual flags survive beta opt-out.
- Windows and remaining lifecycle tests pending.

Packaged Playwright verification passed at [the current provenance tip](https://github.com/Comfy-Org/Comfy-Desktop/commit/74e4456ebb51845bb41eb327ba1766157f12a9c3), covering the [approved manual-argument behavior](https://github.com/Comfy-Org/Comfy-Desktop/pull/1487#issuecomment-5707938915).

Seed an installation with manually supplied `--enable-assets` and beta enabled. Disable beta through product IPC, verify the persisted setting, quit, and restart without either settings or installation test seeds. Verify beta remains disabled, stored launch arguments are unchanged, and launching that installation produces a real Core process with `--enable-assets` in its arguments and a healthy local server. One test passed; owned processes closed.

Four typechecks, build and Linux packaging passed. The launched staging package has SHA256 `16b5303f9f2f8d1bac74acbf8eb6cd829f7f5e4302c4572639b94ae9fe36a805`. Its compiled Datadog environment is `qa-assets`; each launch passed the authenticated approved-PostHog-project guard. Telemetry was seeded false. No vendor receipt or transport-silence claim.

This is current Desktop with an explicitly older isolated Core 0.35.0 fixture and frontend package 1.52.7. It does not establish current integrated Core, Windows, delayed flag responses, restore/adoption or general provenance acceptance. Initial harness process matching assumed an absolute main.py argument; correcting it to require the fixture working directory and relative main.py produced the passing run. No product source edits, branch mutation, approval or merge.

Glossary: IPC means Electron inter-process communication; unseeded restart means reading persisted profile files rather than injected test values; provenance means verified release ancestry.

</details>

christian-byrne:
Packaged adoption check: green. Verified checkout enabled; divergent checkout excluded. Restore still untested.

Linux staging package at [current head](https://github.com/Comfy-Org/Comfy-Desktop/commit/74e4456ebb51845bb41eb327ba1766157f12a9c3), unchanged archive SHA-256 `16b5303f9f2f8d1bac74acbf8eb6cd829f7f5e4302c4572639b94ae9fe36a805`.

Used natural product probe, track and launch operations on two synthetic standalone Git checkouts. An exact release-tag checkout persisted verified ancestry and its real launched Python argv contained `--enable-assets`. An orphan checkout carrying the same manifest tag label persisted unverified ancestry and omitted that flag. Both used persisted beta ON and the same cached grant. The packaged Playwright pair passed.

Earlier missing-positive-control failures were fixture defects: the synthetic server did not advertise the flag in `--help`, and the environment setting seed left beta OFF. Corrected help and persisted settings; no product change or version-metadata override. Failed attempts remain retained, not product findings.

Boundary: the server is synthetic, not real Core. This proves adopted-checkout provenance gating, not snapshot restore, Windows, delayed flag response/restart or vendor receipt. Those acceptance cases remain open. Every app spawn used the authenticated approved staging guard. No formal approval or merge claim.

Glossary: provenance means verified Git ancestry; argv means actual process launch arguments; orphan means a checkout whose history does not reach the release tag.
</details>

christian-byrne:
Snapshot restore: green. Late OFF survives offline restart. Windows remains unverified.

This supersedes the restore/late-response gaps in my previous test comment. Both checks use the Linux staging package at [this unchanged head](https://github.com/Comfy-Org/Comfy-Desktop/commit/74e4456ebb51845bb41eb327ba1766157f12a9c3), app.asar SHA-256 `16b5303f9f2f8d1bac74acbf8eb6cd829f7f5e4302c4572639b94ae9fe36a805`.

- Snapshot restore: the actual packaged action moves a natural synthetic Git descendant back to a tagged commit. It persists the exact restored commit, verified base tag and zero commits ahead before launching with `--enable-assets`. Parent rerun passed in 3.1 seconds. Dependency synchronization was skipped and the launched server is synthetic; this does not prove package/PyTorch restoration or real Core behavior.
- Late OFF: start with cached ON. Intercept only the PostHog SDK HTTP fetch boundary and return explicit OFF after 3501 ms, beyond the two-second launch deadline and 963 ms after observing the current Core process with `--enable-assets`. Verify false on disk, quit, confirm the old Core exited, then restart with controlled network errors. Actual restarted Core arguments omit the injected flag. Strengthened parent rerun passed in 31.7 seconds. The SDK decoding, deadline, persistence and launch code are unchanged; no SDK-result or cache-setter mock supplies the late result.

The late test uses an older real [Core fixture](https://github.com/Comfy-Org/ComfyUI/commit/f42b24efbeee194513fff465d84ad6913a2698d5), not the current integrated Core. A debugger installs the fetch interceptor at the packaged request callsite because initialization overwrites the early wrapper. Earlier failed interception attempts remain recorded as harness failures, not product defects. Every spawn uses the authenticated staging guard; owned processes and ports are clear. No vendor-service response, Windows acceptance, formal approval or merge is claimed.

Retained workspace evidence: `fleet/qa/restore-2026-09-17-0937/evidence/acceptance.json`; `fleet/qa/late-flags-2026-09-17-0937/evidence/{network.jsonl,restart-network.jsonl,result.json}`; parent logs `research/qa/{restore-parent,late-flags-strengthened}-2026-09-17-0937.log`.

Glossary: provenance means verified release ancestry; SDK means the PostHog client library; injected flag means an argument supplied by a remote grant rather than the user's manual arguments.

</details>

christian-byrne:
Real Linux permission denial: grant withheld. Dangling-link assertion passes.

At [this unchanged head](https://github.com/Comfy-Org/Comfy-Desktop/commit/74e4456ebb51845bb41eb327ba1766157f12a9c3), a non-root test creates a readable `.git` directory, removes traversal permissions from its parent, and confirms the actual kernel returns EACCES from lstat. The production handleLaunch and checkout resolver withhold `--enable-assets`. Restoring permissions makes the directory readable again. No syscall error was injected.

A second case creates a real dangling `.git` symlink and asserts the link exists even though stat-based existence is false. The launch resolver withholds the grant. Linux symlink creation failure throws rather than silently returning before assertions. Parent focused rerun: two passed in 1.14 seconds; the other 79 tests were excluded by the name filter, not newly verified.

Scope: real Linux filesystem and production launch/resolver functions, but the existing suite stubs Electron, grants, argument schema and process spawning. Assertions inspect intercepted spawn arguments, not a running Core process. This is not packaged Electron refusal or Windows permission/symlink proof. The app need not refuse all launching; it must withhold the unverified remote grant. No new product defect is asserted.

Retained test-only patch and runner: `fleet/qa/permissions-2026-09-17-1237/`; parent log `research/qa/permissions-parent-2026-09-17-1237.log`. Temporary source patch, permissions and dependency link were cleaned up; no product source changes remain.

Terms: EACCES means operating-system permission denied; dangling symlink means a link to a missing target; grant means a remotely supplied flag rather than a manual user argument.

</details>

christian-byrne (approved):
No blocking findings at the pinned head.

christian-byrne (approved):
Restack verified: this PR layer is byte-identical to the previously approved layer. The separate main-branch draft-handoff fix is preserved unchanged. Existing nonblocking observations and packaged-acceptance limits still apply.

christian-byrne (approved):
Post-merge restack verified: both base and head trees match the previous approval exactly. No source changes; the reviewed provenance checks and existing acceptance limits remain unchanged.

PR #1541, opened 2026-09-20
feat(telemetry): make Datadog the usable exception sink

christian-byrne:
Exceptions reached Datadog already, but as a fixed string, so no monitor could tell one failure from another. This makes that payload usable and moves the PostHog copy behind an opt-in.

## What was wrong

`reportRendererError` builds a scrubbed, length-capped `Error` from the forwarded payload, then throws it away:

```ts
const datadogError = new Error('Desktop application exception')
datadogError.name = 'DesktopTelemetryError'
datadogError.stack = undefined
datadogRum.addError(datadogError, { origin, source, forwarded_source, level, reason, exitCode, exit_code, type })
```

No message, no stack, and no `error_type`. Every distinct failure mode collapsed into one indistinguishable Datadog error. You could alert on "desktop exceptions rose" and never on *which* one. That matters now that the hosted frontend routes every `reportError()` through this path (#1485).

## Changes

- `forwardExceptionToRenderer` takes the scrubbed error and forwards its message and stack. Context keys stay allow-listed; `error_type` is added to that list. Free-form diagnostics still never reach Datadog.
- The renderer reports the scrubbed error it already built instead of the fixed string, and passes `error_type` through.
- The PostHog exception copy is opt-in behind `POSTHOG_EXCEPTIONS=1`. Both sinks carried the same scrubbed error and PostHog bills per event.

## Deliberate reversal, flagged for review

#1271 intentionally sent no message or stack to Datadog ("Forward scrubbed exception context (reason/exitCode/type) to the renderer Datadog mirror; keys stay allow-listed"). This reverses that half, on purpose, because Datadog is now the primary exception sink rather than a secondary mirror. Sending nothing identifiable was defensible when PostHog held the detail; it is not once PostHog is opt-out.

No new unscrubbed surface: `telemetry.ts` is the only `dd-error` producer, the payload is `deliverException`'s `safeError` (already `scrubAll`-ed and capped), and `reportRendererError` scrubs again on receipt. This is the same text PostHog already received.

## Change breakdown

Total changed lines 143 (118 added, 25 deleted).

| Category | Files | Paths | Added | Deleted | Changed | Share |
| --- | --- | --- | --- | --- | --- | --- |
| Product code | 4 | `src/main/lib/telemetry.ts`, `src/renderer/src/lib/rendererBootstrap.ts`, `src/shared/posthogConfig.ts`, `src/types/ipc.ts` | 64 | 24 | 88 | 61.5% |
| Test code | 1 | `src/main/lib/telemetry.test.ts` | 54 | 1 | 55 | 38.5% |

No documentation, configuration, generated files, lockfiles, or vendored code. No merge-only changes.

The product-code count is inflated by comment and docstring updates inside those four files: the module header's provider-split section, the `DatadogForwardedError.skipPostHog` doc, and the `forwardExceptionToRenderer` docstring all described the old behaviour and would have been actively misleading if left. The executable change is roughly 20 lines.

## Verification

- Full unit suite: 276 files, 5100 passed, 2 skipped.
- `pnpm typecheck` (node, web, e2e, integration) clean; `pnpm lint` clean; `prettier --check` clean.
- Both new tests were control-checked: reverting each half of the fix turns the matching test red, restoring it turns it green.

New tests:
- forwards the scrubbed message, stack and `error_type` so a monitor can tell failures apart
- suppresses the PostHog exception copy unless `POSTHOG_EXCEPTIONS` opts in, while the Datadog forward still happens

The existing "forwards exceptions to Datadog without message, stack, or arbitrary context" test is renamed to "forwards a bare notice to Datadog when no error is supplied" and kept: it still pins the no-error call path and the context allow-list.

Justification: Merged without pre-merge approval under release time pressure, authorized offline by the author, to complete the Desktop error-reporting path landed in #1485 and Comfy-Org/ComfyUI_frontend#16917 before the release window closes. No human has reviewed this PR. Risk is limited: no new unscrubbed data reaches any sink, the only behaviour change to an existing path is richer detail on an error already being sent, full unit suite and all gates green. Post-merge peer review is owed.


christian-byrne:
Admin-merged without waiting for in-band approval, same release window as #1485 and Comfy-Org/ComfyUI_frontend#16917. Authorized offline by me as the author.

Stating it plainly: no human has reviewed this PR. A post-merge peer review is owed under CC8.1. CI was 15/15 green at merge, and the full unit suite (276 files, 5100 tests) passed locally.

The one thing worth a reviewer's eye is the deliberate reversal of #1271's "no message or stack to Datadog" decision, explained in the PR body. Short version: that was right when PostHog held the error detail, and wrong now that Datadog is the primary exception sink and PostHog is opt-out. No new unscrubbed data reaches Datadog, it is the same scrubbed text PostHog was already receiving.

@huang47 @benjcooley if either of you has ten minutes, the diff is ~20 executable lines across `telemetry.ts` and `rendererBootstrap.ts`.

